PR #3, opened 2019-03-20
pod annotations added

maximmi:
## what
* add support for pod annotations added

## why
* pod annotations are used for specifying extra metadata needed by other controllers, like `kiam`

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -15,6 +15,11 @@ spec:
     metadata:
       labels:
         app: cost-analyzer
+{{- if .Values.podAnnotations}}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Hi maximmi, I can look at this a bit more, but right now I'm getting 
> helm template .
Error: parse error in "cost-analyzer/templates/cost-analyzer-deployment-template.yaml": template: cost-analyzer/templates/cost-analyzer-deployment-template.yaml:71: unexpected EOF

when applying this change with helm client 
Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}

Mind digging into that when you have a moment?

AjayTripathy:
Thanks for getting this fixed!

PR #10, opened 2019-04-27
Create cost-analyzer-ingress.yaml

dwbrown2:
This PR is incomplete because 1) kubecost-cost-analyzer needs to become a NodePort and 2) we need to add Values.ingress.enabled

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-ingress.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingress.yaml
@@ -1,0 +1,29 @@
+{{- if .Values.ingress.enabled -}}
+
+apiVersion: v1
+data:
+  auth: YWRtaW46JGFwcjEkZ2tJenJxU2ckMWx3RUpFN1lFcTlzR0FNN1VtR1djMAo= # default is admin : admin
+kind: Secret
+metadata:
+  creationTimestamp: "2019-04-26T23:36:47Z"
+  name: kubecost-auth
+  namespace: kubecost
+type: Opaque
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: kubecost-ingress
+  namespace: kubecost
+  labels:
+    app: kubecost
+  annotations:
+     nginx.ingress.kubernetes.io/auth-type: basic
+     nginx.ingress.kubernetes.io/auth-secret: kubecost-auth
+     nginx.ingress.kubernetes.io/auth-realm: "Authentication Required - ok"
+spec:
+  backend:
+    serviceName: kubecost-cost-analyzer
+    servicePort: 9090
+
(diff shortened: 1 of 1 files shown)


mdaniel:
> This PR is incomplete because 1) kubecost-cost-analyzer needs to become a NodePort

Please don't do that, or at bare minimum make it a configuration knob. We have an Ingress controller expressly to avoid having to carve off NodePorts for every Service that wishes to expose itself. Those ports are a limited resource, and I treat them that way

cost-analyzer/templates/cost-analyzer-ingress.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.ingress.enabled -}}
+
+apiVersion: v1
+data:
+  auth: YWRtaW46JGFwcjEkZ2tJenJxU2ckMWx3RUpFN1lFcTlzR0FNN1VtR1djMAo= # default is admin : admin
+kind: Secret
+metadata:
+  creationTimestamp: "2019-04-26T23:36:47Z"
+  name: kubecost-auth
+  namespace: kubecost
```
mdaniel: `{{ .Release.Namespace }}` or [just omit it like the other templates](https://github.com/kubecost/cost-analyzer-helm-chart/blob/820a5031a13cd6391d3b859fade03048d28e263a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L3-L7), right? Same with its `Ingress` friend below

dwbrown2:
Thanks for feedback on this... just to clarify, my proposal is to only make kubecost-cost-analyzer service a NodePort if and only if Values.ingress.enabled = true. Let me know if you still feel like that would be the wrong approach.

mdaniel:
> my proposal is to only make kubecost-cost-analyzer service a NodePort if and only if

I do, yes -- the whole point of an Ingress is to avoid having to make new `type: LoadBalancer` or `type: NodePort` `Service`s -- it trades the almost unbounded number of virtual hosts for the very bounded number of ports and $$$ that comes with actual load balancers

I mean, I can appreciate not everyone is jammed up by N.P.s so that's why I mentioned the configuration knob, but as a default, I think `type: ClusterIP` is a very sane default

dwbrown2:
I was under the impression that I couldn't use a `ClusterIP` (issue/docs below) with the GKE ingress controller. Regardless, I agree that default should not be a `NodePort`. Let me investigate further.

https://cloud.google.com/kubernetes-engine/docs/concepts/ingress
https://github.com/kubernetes/kubernetes/issues/26508

dwbrown2:
I'm going to close this PR. It was originally meant to be an example for teams to have a quick ingress solution but I believe it introduces more problems than it solves.

Issue #13, opened 2019-04-30
cost-analyzer-checks resource limit is too low


mdaniel:
Ironic, given #12 but one can see in the `Reason:` code that node is being OOMKilled

```
  cost-analyzer-checks:
    Container ID:  docker://d9d86367a7c7b54e22331f2bc6db2178a94667239105d491f3bde9877e4b4171
    Image:         ajaytripathy/kubecost-checks:prod-1.18.2
    Image ID:      docker-pullable://ajaytripathy/kubecost-checks@sha256:f9c45b42c8facd366a0515544f2a9fbfc8b75af8dea5e54f29bcbd4ecbdfeff8
    Port:          <none>
    Host Port:     <none>
    Args:
      node
      ./node/cron.js
    State:          Terminated
      Reason:       OOMKilled
      Exit Code:    0
      Started:      Tue, 30 Apr 2019 15:00:21 -0700
      Finished:     Tue, 30 Apr 2019 15:01:48 -0700
    Ready:          False
    Restart Count:  0
    Limits:
      cpu:     10m
      memory:  55M
    Requests:
      cpu:     10m
      memory:  55M
```

dwbrown2:
Hopefully fixed by #17

Issue #12, opened 2019-04-30
ideally containers would specify resource limits


mdaniel:
As of this issue, none of the [`resources:` blocks](https://github.com/kubecost/cost-analyzer-helm-chart/blob/249ce8bd6bc795e5e17e583ad5af9f333f800f0e/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L86-L88) contain `limits:` clauses, meaning those containers can grow without bound

That's not an idea situation, and if the memory requests are only 55MB then setting a 256M upper bound seems like a perfectly reasonable starting point -- although the correct answer is influenced by observing the actual memory pressure of a running instance

dwbrown2:
I agree we should add limits back to all of these containers.

For context, we removed because certain systems were being CPU throttled into oblivion during critical events, e.g. AWS billing data download. Like Matthew, I propose a set of conservative limits that can be lowered over time, maybe 256MB and 500M for cost-model?

dwbrown2:
Fixed by #17

PR #17, opened 2019-05-01
Ajay tripathy add limits

### LINKED ISSUE #11: ideally cost-analyzer-server would declare its resource needs
As of this issue, the [`cost-analyzer-server` container](https://github.com/kubecost/cost-analyzer-helm-chart/blob/249ce8bd6bc795e5e17e583ad5af9f333f800f0e/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L47-L64) does not declare any `resources:`, which makes the kubernetes scheduler's job harder (and, ironically, AIUI makes assigning the cost associated with that container harder, too)

AjayTripathy:
Fix #11 #12 #13 with better resourcing.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -43,12 +43,18 @@ spec:
           resources:
             requests:
               cpu: "10m"
-              memory: "55M"
+              memory: "55Mi"
+            limits:
+              cpu: "100m"
+              memory: "256Mi"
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -84,8 +107,8 @@ spec:
                   key: kubecost-token
           resources:
             requests:
-              cpu: 10m
-              memory: 55M
+              cpu: "10m"
```
dwbrown2: am i wrong or is this duplicated from line 77-79?

dwbrown2 (approved):

PR #24, opened 2019-05-08
Update cost-analyzer-deployment-template.yaml

dwbrown2:


dwbrown2:
Yep, I'll apply to all the other templates. I just wanted to see if this set looked good to you before I did. I didn't feel like `instance` and several of the other labels applied and wanted to see if you disagreed.

Issue #21, opened 2019-05-03
Make container image registries customisable 


luisdavim:
Could you make the docker images configurable through the values file? Where I work I can’t pull public images directly it would be nice to be able to change it in the values file instead of having to go through the templates every time.

AjayTripathy:
Fixed by #28 , let us know if you need anything else here!

PR #28, opened 2019-05-10
Add configurable repository for Docker images

Pinpin31:
Hi,
we work with a private proxy docker registry (Artifactory) so we need to be able to configure image repository.
So I made some modifications in helm templates to be able to change these parameters.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-checks-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-checks-template.yaml
@@ -10,7 +10,7 @@ spec:
         spec:
           containers:
           - name: cost-analyzer-checks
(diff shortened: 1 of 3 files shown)


dwbrown2:
@Pinpin31 thanks for contribution! We're testing now. This should also fix #21. Anything else we can get you guys as you start using the tool?

AjayTripathy:
This lgtm for now, merging.

 Will be supported in the next release, but before that I will have to add a check similar to [this](https://github.com/kubecost/cost-analyzer-helm-chart/commit/de6b40c935252893e637f04cd976c9163aa55273) to support users installing via --reuse-values .

PR #32, opened 2019-05-13
use configmaps to load grafana charts

### LINKED ISSUE #25: Move Grafana Dashboards to Files (JSON)
## what
* Can we move these dashboards to files instead then modify the chart to include them?

image

## why
this will make it easier for 
- (a) you to maintain them since you can just drop a file in without packing it all on one line
- (b) for others use them if not using your grafana, since grafana supports pulling dashboards from a URL, so we could pull them using the github raw url

## references
* https://github.com/helm/helm/blob/master/docs/chart_template_guide/accessing_files.md

AjayTripathy:
Fixes #25 , as well as makes configmap reload for additions to the file automatic.


osterman:
This is awesome!! Thank you for the prompt turnaround!.

PR #33, opened 2019-05-17
Ajay tripathy update grafana

AjayTripathy:


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -27,6 +27,8 @@ data:
         server_name _;
         root /var/www;
         index index.html;
+        add_header Cache-Control "no-cache";
```
dwbrown2: why no-cache? don't we want these cached until a release?

AjayTripathy: The no-cache directive means that a browser may cache a response, but must first submit a validation request to an origin server ie check etag/last modified

dwbrown2: Got it, thanks

PR #37, opened 2019-05-18
Add init container to eliminate restarts

AjayTripathy:


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -52,8 +65,8 @@ spec:
               cpu: "10m"
               memory: "55Mi"
             limits:
-              cpu: "100m"
-              memory: "256Mi"
+              cpu: "1000m"
```
dwbrown2: i believe this were included by accident.

PR #40, opened 2019-05-20
Add an ingress resource

sebastien-prudhomme:
Add an ingress resource to the chart.

I had to rename the frontend service name because ingress ```serviceName`` is limited to 15 characters. For consistency, i've done the same thing for other service names

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
@@ -1,0 +1,40 @@
+{{- if .Values.ingress.enabled -}}
+{{- $fullName := include "cost-analyzer.fullname" . -}}
+{{- $ingressPaths := .Values.ingress.paths -}}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ $fullName }}
+  labels:
+    app.kubernetes.io/name: {{ include "cost-analyzer.name" . }}
+    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+{{- if .Values.ingress.tls }}
+  tls:
+  {{- range .Values.ingress.tls }}
+    - hosts:
+      {{- range .hosts }}
+        - {{ . | quote }}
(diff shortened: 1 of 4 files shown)


dwbrown2:
Thanks for this contribution, Sébastien! Assigning to @AjayTripathy for review.

cost-analyzer/templates/cost-analyzer-ingress-template.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if .Values.ingress.enabled -}}
```
AjayTripathy: We also need to check if .Values.ingress here for safety, as some users install with the same values.yaml. See https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/templates/cost-analyzer-network-policy.yaml#L1 for example.

cost-analyzer/templates/cost-analyzer-service-template.yaml
```diff
@@ -9,12 +9,12 @@ spec:
     app: cost-analyzer
   type: ClusterIP
   ports:
-    - name: cost-analyzer-server
+    - name: server
```
dwbrown2: Can these be `kc-server`, `kc-model`, and `kc-frontend` to minimize risk of collisions? +@AjayTripathy

cost-analyzer/templates/cost-analyzer-ingress-template.yaml
sebastien-prudhomme: This is the standard ingress.yaml file generated by ```helm create```.

I see not other chart that are doing a check on the existence off .Values.ingress. It always exists because it is in the values.yaml

But I can add it if you really want to

cost-analyzer/templates/cost-analyzer-service-template.yaml
sebastien-prudhomme: I don't figure out where is the risk of collision as all this ids come from the same chart. Can you give more explanations?

dwbrown2: You can disregard my comment. I was referring to our internal development process but this should be fine.

cost-analyzer/templates/cost-analyzer-ingress-template.yaml
AjayTripathy: This is to protect against legacy values.yaml installed via helm --reuse-values.  We'll look to remove support for old values.yaml files in a future release, but for now we're supporting them, so it would be preferable to have this.

sebastien-prudhomme: Done

PR #41, opened 2019-05-21
don't install configmaps on no grafana configuration

AjayTripathy:


cost-analyzer/templates/grafana-dashboard-cluster-metrics-template.yaml
```diff
@@ -1,3 +1,7 @@
+{{- if .Values.grafana -}}
+{{- if .Values.grafana.sidecar -}}
+{{- if .Values.grafana.dashboards -}}
+{{- if .Values.grafana.dashboards.enabled -}}
```
Nuru: This is undesirable overkill
```
{{- if .Values.grafana -}}
{{- if .Values.grafana.sidecar -}}
{{- if .Values.grafana.dashboards -}}
{{- if .Values.grafana.dashboards.enabled -}}
```
All we need or want for the dashboards is `{{- if .Values.grafana.dashboards.enabled -}}`

If you want to get fancy, you can also add a value specific to each dashboard, such as `{{- if .Values.grafana.dashboards.clusterMetrics.enabled -}}`

cost-analyzer/templates/grafana-datasource-template.yaml
```diff
@@ -1,3 +1,7 @@
+{{- if .Values.grafana -}}
+{{- if .Values.grafana.sidecar -}}
+{{- if .Values.grafana.datasources -}}
+{{- if .Values.grafana.datasources.enabled -}}
 apiVersion: v1
```
Nuru: Like with the dashboards, all we need for the datasource is `{{- if .Values.grafana.datasources.enabled -}}`; the other conditionals are not helpful.

Nuru (changes requested):

cost-analyzer/templates/grafana-dashboard-cluster-metrics-template.yaml
AjayTripathy: These conditionals are defensive programming to protect people using --reuse-values from earlier values.yaml, which we in general support in version 1.X.X. Supplying a default isn't enough We can look into simplifying in a later version.

cost-analyzer/templates/grafana-datasource-template.yaml
AjayTripathy: Ditto re: unnecessary conditionals. However, this does raise a point about proxying to grafana and assumptions we've made about grafana being reachable-- I've addressed those in another commit.

Nuru (approved):

cost-analyzer/templates/grafana-dashboard-cluster-metrics-template.yaml
Nuru: @AjayTripathy okay

PR #43, opened 2019-05-23
Template changes to support grafana url being supplied from configmaps.

AjayTripathy:


cost-analyzer/templates/external-grafana-config-map-template.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if eq .Values.global.grafana.proxy false -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: external-grafana-config-map
+data:
+    grafanaURL: {{ .Values.global.grafana.domainName }}
```
Nuru: Please be consistent and accurate about `domainName` (which is like `grafana.example.com`) and `URL` (which is like `https://grafana.example.com`). For backwards compatibility and adequate flexibility, I suggest
```
data:
  grafanaURL: {{ .Values.global.grafana.scheme | default "http" }}://{{- .Values.global.grafana.domainName }}
```

AjayTripathy: Good suggestion, thanks.

PR #50, opened 2019-05-29
Add configurable PVC size

botzill:
Added configurable PVC size. For Digitalocean k8s, the minimal size of the volume is 1Gi.

Thx.


botzill:
@dwbrown2 Any chance you can take a look?

dwbrown2:
Looks good @botzill! Thanks for this contribution. Merging now. 

@AjayTripathy will add a safeguard here for teams using old values.yaml with --reuse-values

PR #55, opened 2019-06-03
ingress fixes

### LINKED ISSUE #53: Ingress spec is hard to use
Several users have reported our Ingress as hard to use. Here are my initial ideas on improving:

1) Add suggested paths or make these not required. 
2) Is default `servicePort` correct?
3) Make it easy to connect this to a NodePort service
4) In comments, let users know they don't need to route to 9001 and 9003 directly, because kubecost ngnix will handle.

AjayTripathy:
Fixes #53

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-service-template.yaml
@@ -7,9 +7,9 @@ metadata:
 spec:
   selector:
     app: cost-analyzer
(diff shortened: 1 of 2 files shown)


sebastien-prudhomme:
@AjayTripathy Using `/*` for the path breaks my installation. Any reason you use that value instead of `/`?

dwbrown2:
Also, @AjayTripathy do you mind updating the README to reflect the default path change?

Issue #5, opened 2019-04-18
Add recommended labels


dwbrown2:
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

dwbrown2:
@luisdavim I have #24 right now. Anything else you would want to see there?

sebastien-prudhomme:
@dwbrown2 the recommended labels fro Helm charts are there: https://helm.sh/docs/chart_best_practices/#labels-and-annotations

In practice for an application with only one component (without taken into account subcharts)
```
  labels:
    app.kubernetes.io/name: {{ include "cost-analyzer.name" . }}
    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
    app.kubernetes.io/instance: {{ .Release.Name }}
    app.kubernetes.io/managed-by: {{ .Release.Service }
```

This must be added in every resource managed by the chart (cronjob, service, service account, ...)

The easiest way to do that is to add a ```define``` in the helper YAML file with all this labels and use that ```define``` in every resource

dwbrown2:
Thanks for sharing @sebastien-prudhomme! There seems to be a small amount of conflict between helm and kubernetes recommendations (e.g. helm vs tiller for managed-by). I don't have a strong opinion here but just an observation. 

The only label that I worry about is `app.kubernetes.io/instance`. We've seen teams install multiple deployments of Kubecost and it's not clear to me they would know these need to be unique. Let me give this some thought. Also, I would welcome your opinion on this!

sebastien-prudhomme:
@dwbrown2 with the current version of Helm (v2), the release name must be unique in a Kubernete cluster.

With the next version (v3), the release name will be namespace scoped.

luisdavim:
There is no conflict.  Helm recommends the same labels as kubernetes. If you run helm create with a recent version of helm those are the labels that helm will set by default in all templates.

luisdavim:
Also the label that you mentioned doesn't have to be unique, the combination of labels should be.
I'd suggest getting the latest stable release of helm and using helm create to check what labels and selectors are used.

dwbrown2:
Fixed by PR #58 from Sébastien!

PR #58, opened 2019-06-04
Fix labels #5

sebastien-prudhomme:
Trying to use same labels in each resource.

I added an helper so that changing labels is easy.


dwbrown2:
Nice job with this helper addition!! Thanks for contributing this.

AjayTripathy:
Fixes #5

PR #61, opened 2019-06-05
set ingress default path to /

AjayTripathy:


AjayTripathy:
@sebastien-prudhomme there wasn't a particular reason to set this to '/*', just wanted to provide some sensible default suggestions to people using the ingress. '/' works as well for this.

PR #63, opened 2019-06-06
Support for Prometheus operator

### LINKED ISSUE #48: [prometheus] Create ServiceMonitor and PrometheusRule resources
Because multiple applications installed in a single cluster all want to have visibility via Prometheus, Grafana, and Alertmanager, CoreOS has developed the [Prometheus-Operator chart](https://github.com/helm/charts/tree/master/stable/prometheus-operator) which installs a single instance of these tools that can be shared by all the applications in the cluster. Following the principle of Kubernetes that each application should specify its own resource needs, Prometheus Operator includes [new Custom Resources](https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md) that allow applications to add configuration items to Prometheus, Grafana, and Alertmanager. 

I am grateful that [you did this work](https://github.com/kubecost/cost-analyzer-helm-chart/pull/32) for the Grafana dashboards and ask that you now do it for [ServiceMonitors]() (which replace [`scrape_configs`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/6ea6d201da050627249af23999501209f2df8430/cost-analyzer/charts/prometheus/values.yaml#L1107-L1118)) and [PrometheusRules](https://github.com/kubecost/cost-analyzer-helm-chart/blob/6ea6d201da050627249af23999501209f2df8430/cost-analyzer/charts/prometheus/values.yaml#L838-L845).

One tricky thing is that `ServiceMonoitors` are meant to monitor `Services`, although technically they monitor `Endpoints` that are typically automatically created by Kubernenets for `Services`. The important point is that `ServiceMonitors` target endpoints to scrape via [`LabelSelectors`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#labelselector-v1-meta), which means your `Services` need to have stable and unique labels. Currently, your `kubecost-cost-analyzer` `Service`, which is what you target with your custom `scrape_config` and I now want you to target with a `ServiceMonitor`, only has 1 label: `chart: cost-analyzer-1.26.0`. This is not robust enough for this purpose. I suggest you add the `app: cost-analyzer` label to it as well. With that, the following `ServiceMonitor` should work:
```
kind: ServiceMonitor
metadata:
  labels:
    app: prometheus-operator
    release: prometheus-operator
  name: cost-analyzer-model
spec:
  endpoints:
  - honorLabels: true
    interval: 1m
    metrics_path: /metrics
    port: cost-analyzer-model
    scheme: http
    scrapeTimeout: 10s
  selector:
    matchLabels:
      app: cost-analyzer
```

I don't know if you have `PrometheusRules` meant for publishing. If the ones you have are just examples so your Prometheus installation is not empty, then they do not need to be converted to `PrometheusRule` resources. However, if you have some you want everyone to use, then please install them as `PrometheusRule` resources rather than as configuration files.

sebastien-prudhomme:
Fix #48 

I add a way to disable the creation of the Prometheus datasource in Grafana because it's yet done by the Prometheus Operator helm chart.

I also add an helper for the labels used in selectors in different resources.

I tried to reuse the things defined in values.yaml. It works fine for rules as i can reuse  `.Values.prometheus.serverFiles.rules` but it's not possible to do the same thing for the scraping part as `.Values.prometheus.extraScrapeConfigs` is a string and not structured data and i can not see a way to convert that string to a Go structure

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -13,7 +13,7 @@ spec:
   template:
     metadata:
       labels:
-        app: cost-analyzer
+        {{ include "cost-analyzer.selectorLabels" . | nindent 8 }}
 {{- if .Values.global.podAnnotations}}
       annotations:
 {{- with .Values.global.podAnnotations }}
--- a/README.md
+++ b/README.md
@@ -19,3 +19,8 @@ Parameter | Description | Default
 `ingress.hosts` | Ingress hostnames | `[cost-analyzer.local]`
 `ingress.tls` | Ingress TLS configuration (YAML) | `[]`
 `networkPolicy.enabled` | If true, create a NetworkPolicy to deny egress  | `false`
+`serviceMonitor.enabled` | Set this to `true` to create ServiceMonitor for Prometheus operator | `false`
+`serviceMonitor.additionalLabels` | Additional labels that can be used so ServiceMonitor will be discovered by Prometheus | `{}`
+`prometheusRule.enabled` | Set this to `true` to create PrometheusRule for Prometheus operator | `false`
+`prometheusRule.additionalLabels` | Additional labels that can be used so PrometheusRule will be discovered by Prometheus | `{}`
+`grafana.sidecar.datasources.defaultDatasourceEnabled` | Set this to `false` to disable creation of Prometheus datasource in Grafana | `true`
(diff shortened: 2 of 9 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   template:
     metadata:
       labels:
-        app: cost-analyzer
```
AjayTripathy: Can we leave this on as one of the selectorlabels to support things outside of the helm chart that may have used this label? Let's move this to _helpers.tpl

sebastien-prudhomme: You mean in common labels? I see no reason to keep it in selector labels as all resources referenced by the selectors are in the helm chart.

AjayTripathy: I believe common labels is fine, we just don't want to remove this entirely from labels as people may be relying on it still.

sebastien-prudhomme: I add the `app` label again.

PR #70, opened 2019-06-20
add app field to commonlabels

AjayTripathy:


AjayTripathy:
@dcherman were you thinking about something like this to fix your issue?

PR #73, opened 2019-07-20
up requests to avoid throttling

AjayTripathy:


dwbrown2 (commented):
How do you feel about bumping cpu limits too? Proposal is 1 cpu for `cost-model` and 500m cpu for `cost-analyzer-server`?

PR #76, opened 2019-07-26
Ajay tripathy fix customname

AjayTripathy:
changing the --name parameter from 'kubecost' to something else was previously broken because of the way we do service discovery of our prometheus. This should fix that.


AjayTripathy:
Fixes #77

PR #84, opened 2019-08-14
fix: the service name should be based on the release name

### LINKED ISSUE #83: Generated Service name is incorrect
In https://github.com/kubecost/cost-analyzer-helm-chart/pull/76/files, it looks like the name of the service was changed to be hard-coded to `kubecost-cost-analyzer`, however that goes against the Helm norms where the release name is part of the generated service name.

It also broke the ingress definition since the ingress spec does use the generated / full name for the target service which no longer matches:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/1c4c624b0e33c620c9f4ed5bffbe5c6065c008c0/cost-analyzer/templates/cost-analyzer-ingress-template.yaml#L34

dcherman:
In order to generate the scrape config for prometheus, we need to have a
well-known service name, however there's a limitation on Helm where you can't
access variables from the parent context, so we cannot know what the actual
chart name is.

Prefixing a hard-coded service name with the release name is a middle-ground
compromise which should still allow for multiple kubecost installations if
desired.

Fixes #83

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -24,6 +24,10 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 {{- end -}}
 
+{{- define "cost-analyzer.serviceName" -}}
+{{- printf "%s-%s" .Release.Name "cost-analyzer" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
 {{/*
 Create chart name and version as used by the chart label.
 */}}
--- a/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
@@ -1,6 +1,7 @@
 {{- if .Values.ingress -}}
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "cost-analyzer.fullname" . -}}
+{{- $serviceName := include "cost-analyzer.serviceName" . -}}
 {{- $ingressPaths := .Values.ingress.paths -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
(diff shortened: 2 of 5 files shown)


cost-analyzer/templates/_helpers.tpl
```diff
@@ -24,6 +24,10 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 {{- end -}}
 
+{{- define "cost-analyzer.serviceName" -}}
+{{- printf "%s-%s" .Release.Name "kubecost-cost-analyzer" | trunc 63 | trimSuffix "-" -}}
```
AjayTripathy: This makes our servicename for the recommended installation path (--name kubecost --namespace kubecost) kubecost-kubecost-cost-analyzer -- it is kind of unintuitive. Can we just do {{ .Release.Name }}-cost-analyzer?

cost-analyzer/templates/cost-analyzer-ingress-template.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.ingress -}}
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "cost-analyzer.fullname" . -}}
+{{- $serviceName := include "cost-analyzer.serviceName" . -}}
```
AjayTripathy: Mind updating the nginx config as well to support users who don't use an Ingress? 

https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L1

Servicenames are used here:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L12

https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L16

dcherman:
@AjayTripathy Should be all set

AjayTripathy:
LGTM, thanks! We'll get this into the next release.

PR #87, opened 2019-08-28
parametrise resources

rccrdpccl:
In our case, if not limited, kubecost has CPU usage very high. We would like to reserve such CPU usage for the application, but at the moment it's impossible as it's hardcoded

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-checks-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-checks-template.yaml
@@ -21,7 +21,7 @@ spec:
             - node
             - ./node/cron.js
             resources:
-{{ toYaml .Values.kubecostFrontend.resources | indent 14 }}
+{{ toYaml .Values.kubecostChecks.resources | indent 14 }}
             env:
             - name: COST_ANALYZER_SERVER_ENDPOINT
               value: {{ template "cost-analyzer.fullname" . }}.{{ .Release.Namespace  }}.svc.cluster.local:9001
--- a/cost-analyzer/templates/cost-analyzer-checks-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-checks-template.yaml
@@ -21,12 +21,7 @@ spec:
             - node
             - ./node/cron.js
             resources:
-              requests:
-                cpu: "20m"
-                memory: "100Mi"
-              limits:
-                cpu: "100m"
-                memory: "200Mi"
+{{ toYaml .Values.kubecostFrontend.resources | indent 14 }}
             env:
             - name: COST_ANALYZER_SERVER_ENDPOINT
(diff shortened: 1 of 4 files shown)


rccrdpccl:
Thanks for the quick response.

In this PR we are maintaining previous default values, however there is a non BC change: if people would have already defined limits in their value file, after an upgrade to this version it would not work, as it is defined in another place now (i.e. `.Values.kubecostFrontend.limits` moved to `.Values.kubecostFrontend.resources.limits`).

By very high, I mean relative to its declared consumption: in our clusters, `cost-analyzer` has a constant usage of around 500m, compared to the 100m declared. This causes it to be an exception in our alerting system. I don't  see this as a problem, but I would like to declare it this way.

Another way to tackle this, it would be to implemente VPA, that would automatically set requests value based on usage. https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler
If you'd prefer that solution, I'd be glad to implement it.

rccrdpccl:
Below CPU usage of container cost-model, so that you can have more data.

If you need more details please just ask.

kubecost_cost_model

dwbrown2:
Reviewing this PR now, just need to decide about non-BC change.

I'm unable to reproduce your high CPU usage though. How many pods/nodes do you have running in this cluster? Also, how many people are using the Kubecost application? Thanks again for reporting.

image

cost-analyzer/templates/cost-analyzer-checks-template.yaml
```diff
@@ -21,12 +21,7 @@ spec:
             - node
             - ./node/cron.js
             resources:
-              requests:
-                cpu: "20m"
-                memory: "100Mi"
-              limits:
-                cpu: "100m"
-                memory: "200Mi"
+{{ toYaml .Values.kubecostFrontend.resources | indent 14 }}
```
AjayTripathy: This should be .Values.kubecostChecks.resources , right?

rccrdpccl: Oops, my bad

rccrdpccl:
> Reviewing this PR now, just need to decide about non-BC change.
> 
> I'm unable to reproduce your high CPU usage though. How many pods/nodes do you have running in this cluster? Also, how many people are using the Kubecost application? Thanks again for reporting.
> 
> image

We have over 50 nodes and over 1500 pods. If by "Kubecost application" are you referring to the UI, I think barely one person is using it (at the time the chart was taken, the UI wasn't being used).

If you prefer to continue this conversation in another channel (as this PR will probably be closed) feel free to reach me at my email (available in commits message).

dwbrown2:
Thanks, let us try to recreate your environment. A lot of this CPU usage is caused by our jobs for alerts/updates. One potential solution to reduce would be to provide a helm flag to disable these jobs if you're not using alerting.

PR #101, opened 2019-09-23
add network scrape configs

AjayTripathy:


cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
```diff
@@ -0,0 +1,48 @@
+{{- if .Values.networkCosts -}}
+{{- if .Values.networkCosts.enabled -}}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ template "cost-analyzer.networkCostsName" . }}
+  labels:
```
AjayTripathy: This block is probably unnecessary given the block on L17-18

cost-analyzer/values.yaml
```diff
@@ -165,6 +165,15 @@ prometheus:
             - expr: sum(container_memory_working_set_bytes{container_name!="POD",container_name!=""})
               record: kubecost_cluster_memory_working_set_bytes
 
+networkCosts:
+  enabled: true
+  image: gcr.io/guestbook-227502/kubecost-network-costs:v26
+  resources:
+    #requests:
```
AjayTripathy: Pretty big CPU request for a daemonset. Any chance we could dogfood our own app to come up with something good here?

mbolt35: Excellent

cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
mbolt35: Ok, so which is recommended? Use template to provide name, labels, or metadata at the root of the yaml?

cost-analyzer/values.yaml
```diff
@@ -166,13 +166,12 @@ prometheus:
...
 networkCosts:
-  enabled: true
+  enabled: false
   image: gcr.io/guestbook-227502/kubecost-network-costs:v26
   resources:
     #requests:
-    #  cpu: "200m"
-    #  memory: "200Mi"
-
+    #  cpu: "50m"
```
dwbrown2: I think we should actually ship with some default requests/limits enabled given it's a daemonset. Otherwise, a memory leak or something could cause widespread problems. This can come later this week after we run scale tests.

PR #107, opened 2019-09-30
add values, bump to 1.42.2

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -166,7 +166,7 @@ prometheus:
               record: kubecost_cluster_memory_working_set_bytes
 
 networkCosts:
-  enabled: false
+  enabled: true
```
dwbrown2: naww, not yet =P

PR #111, opened 2019-10-02
Network Costs Updates...

mbolt35:
* Updates to network costs daemonset to include the host node name as an environment variable. 
* Update image to latest.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
@@ -18,7 +18,6 @@ spec:
         app: {{ template "cost-analyzer.networkCostsName" . }}
     spec:
       hostNetwork: true
-      dnsPolicy: ClusterFirstWithHostNet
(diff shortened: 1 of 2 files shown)


cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
```diff
@@ -18,6 +18,7 @@ spec:
         app: {{ template "cost-analyzer.networkCostsName" . }}
     spec:
       hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
```
dwbrown2: @mbolt35 @AjayTripathy any reason to believe this would have impact outside of network costs pods?

mbolt35: @dwbrown2 Good catch here. I don't think this will have an impact on anything currently, but it wasn't really supposed to be included. I think the default dnsPolicy is fine in this case.

PR #113, opened 2019-10-02
Image Pull Policy Configuration

mbolt35:
Add configuration values for `imagePullPolicy`  for kubecost images which defaulted to `Always`


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -85,7 +89,11 @@ spec:
               mountPath: /etc/nginx/conf.d/
           resources:
 {{ toYaml .Values.kubecostFrontend.resources | indent 12 }}
+        {{- if .Values.kubecostFrontend.imagePullPolicy }}
```
dwbrown2: Looks like this may be missing corresponding parameter in values.yaml?

cost-analyzer/values.yaml
```diff
@@ -168,6 +173,7 @@ prometheus:
 networkCosts:
   enabled: false
   image: gcr.io/kubecost1/kubecost-network-costs:v1
```
AjayTripathy: This is now on v3, right?

mbolt35: Yes

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
mbolt35: Thanks!

mbolt35: It would still be ok since it would default to Always, but yeah, we can add it to values as well

PR #125, opened 2019-10-14
add missing cache header

AjayTripathy:
You need to send this header back to get the browser to deal with caching properly.

https://www.keycdn.com/support/cache-control

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -32,6 +32,7 @@ data:
         server_name _;
         root /var/www;
         index index.html;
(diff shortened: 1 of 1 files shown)


dwbrown2 (commented):
Should you add "must revalidate" to this? https://stackoverflow.com/questions/1046966/whats-the-difference-between-cache-control-max-age-0-and-no-cache

AjayTripathy:
'must revalidate' and max-age-0 should be approximately the same but we can add must-revalidate also.

dwbrown2:
Just going off that Stackoverflow suggestion, so your call. This feels
really hard to test.

Webb
Kubecost, Co-founder

ᐧ

dwbrown2 (approved):

PR #133, opened 2019-10-21
Remove no-cache

dwbrown2:


dwbrown2:
@mbolt35 @nikovacevic are either of you intimately familiar with these headers? We've had issues with our files being cached too long in the past, and we aren't certain how all browsers handle this exact header. We think it's safe to remove but just wanted to see if you had concerns/input.

nikovacevic:
I would consider that a safe header to remove.

PR #143, opened 2019-11-04
Fix for breaking changes in kubernetes API v1.16

james-choncholas:
see https://kubernetes.io/blog/2019/09/18/kubernetes-1-16-release-announcement/

"in 1.16 if you submit a Deployment to the API server and specify extensions/v1beta1 as the API group it will be rejected with:
error: unable to recognize "deployment": no matches for kind "Deployment" in version "extensions/v1beta1"

### DIFF
--- a/kubecost.yaml
+++ b/kubecost.yaml
@@ -12766,7 +12766,7 @@ spec:
             path: /sys
 ---
 # Source: cost-analyzer/charts/grafana/templates/deployment.yaml
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: kubecost-grafana


dwbrown2:
Thanks for reporting, James! This looks like an oversight from our change last week, so a good catch. We'll get this tested and merged. 

We may also need to add selector labels for this new API version, but we can confirm with testing.

james-choncholas:
This may be a more elegant solution
https://github.com/helm/charts/pull/17268/files

AjayTripathy:
Thanks for this! This file is also generated from our helm template, I have a fix there in https://github.com/kubecost/cost-analyzer-helm-chart/pull/144

PR #145, opened 2019-11-07
Allow users to create the service accounts

gopisaba:
Allowing the users to create the service account on their own helps to use EKS feature to map the [IAM roles with service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)


AjayTripathy:
LGTM! Thanks for this PR!

PR #152, opened 2019-11-14
Drop kubernetes-apiservers target and two large cadvisor metrics

dwbrown2:


mbolt35 (approved):
Looks good to me!

AjayTripathy (approved):

dwbrown2:
@mbolt35 @nikovacevic looks like this PR may actually break efficiency metrics on aggregated cost API. any idea why? was a surprise to me and no other obvious impact on product.

PR #153, opened 2019-11-18
Fix docs about Prometheus server

adri:
The default in the [values.yaml](https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml#L4)  `true`, not `false`. 

Also `false` means that a custom prometheus server should be used.


dwbrown2:
Thanks for catching and reporting this!

PR #163, opened 2019-12-04
Remove ksm default access to secrets

dwbrown2:


AjayTripathy:
This has no impact but we will have to remove the corresponding ksm module or logs will barf: 
```
E1207 00:14:11.276177       1 reflector.go:126] k8s.io/kube-state-metrics/internal/collector/builder.go:482: Failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:kubecost:kubecost-prometheus-kube-state-metrics" cannot list secrets at the cluster scope
```

dwbrown2:
Let's create another issue to track this log... stay tuned

dwbrown2:
Closing until we figure out why this is needed.

PR #177, opened 2019-12-17
Thanos Optimizations

mbolt35:
Based on observations of memory spiking occurring on `thanos-store`, we decided to test reducing concurrency to `1` in order to force each query to be processed serially in the hopes that would honor memory configuration values. Based on multiple concurrent requests through the cost-model product, we were able to show that using higher concurrency values (ie: 20) yielded much higher spikes in memory (well beyond configured values), while a single threaded approach sat well below the configured limits. 

Additionally, given the nature of cost-model queries containing many time series, we reduced the number of concurrent requests allowed to be made from `thanos-query`. This can be tuned depending on datasets, but prevents memory bloat from aggregating in `thanos-query`.

* Expose thanos-query maxConcurrent and timeout configuration options
* Increase memory requests to be inline with default thanos-store cache-pool-size + index-cache-size 
* Reduce block sync concurrency and grpc series concurrency to prevent bloating memory usage from well exceeding the configured limitations.

### DIFF
--- a/cost-analyzer/values-thanos.yaml
+++ b/cost-analyzer/values-thanos.yaml
@@ -51,11 +51,15 @@ prometheus:
 thanos:
   store:
     enabled: true
+    grpcSeriesMaxConcurrency: 1
+    blockSyncConcurrency: 1
     resources: 
       requests:
-        memory: "2Gi"
+        memory: "2.5Gi"
   query: 
     enabled: true
+    timeout: 3m
+    maxConcurrent: 10
   # Thanos Sidecar Service Discovery
   sidecar:
     enabled: true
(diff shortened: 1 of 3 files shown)


cost-analyzer/values-thanos.yaml
```diff
@@ -51,11 +51,15 @@ prometheus:
 thanos:
   store:
     enabled: true
+    grpcSeriesMaxConcurrency: 1
+    blockSyncConcurrency: 1
     resources: 
       requests:
-        memory: "2Gi"
+        memory: "2.5Gi"
```
mbolt35: This was a changed that @AjayTripathy and I had talked about before. Since `chunk-pool-size` is 2GB and `index-cache-size` is 250MB, we want our requests to fully cover that total + a little breathing room.

cost-analyzer/values-thanos.yaml
```diff
@@ -51,11 +51,15 @@ prometheus:
...
   query: 
     enabled: true
+    timeout: 3m
+    maxConcurrent: 10
```
mbolt35: Let the timeout be 3m for longer running queries from `thanos-query` -> `thanos-store` to ensure that we can appropriately queue using less concurrent request handling.

PR #196, opened 2020-01-13
Add pricing configs

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -221,3 +221,25 @@ grafana:
 
 serviceAccount:
   create: true
+
+# These configs are settable from either the /settings.html page or value.yaml
```
dwbrown2: values.yaml -- missing 's'

cost-analyzer/values.yaml
```diff
@@ -221,3 +221,25 @@ grafana:
...
+# This block overrides set configs on /settings.html on pod restart
```
dwbrown2: 'configs set' instead of 'set configs'?

cost-analyzer/values.yaml
```diff
@@ -221,3 +221,25 @@ grafana:
...
+#kubecostProductConfigs: 
+#  defaultModelPricing:
```
dwbrown2: is it worth adding a comments above this block? proposal is below.

`# default (monthly) model prices are used predominately for on-prem clusters`

dwbrown2 (approved):

PR #198, opened 2020-01-13
Add "persistentVolume.storageClass"

matthope:


AjayTripathy:
Thanks for this PR! Looks like a great addition, let us test and get this submitted.

AjayTripathy:
This LGTM! Thanks for your contribution @matthope and review @egeland

cep21:
Should we add this to all the persistent volumes?

dwbrown2:
@cep21 I actually think it should be available on all now! Which PV are you looking to change storageclass on?

cep21:
I think it sets it on `kubecost-cost-analyzer`, but not the other things it makes like `kubecost-prometheus-server` and `kubecost-prometheus-alertmanager`

dwbrown2:
Those can actually be set in our sub-charts. For example, in our values you can set `prometheus.alertmanager.persistentVolume.storageClass`. Let me know if I can share more!

PR #201, opened 2020-01-16
Add kubecost-turndown to helm chart

mbolt35:
Adding turndown server and configuration for endpoint to helm chart. If enabled, turndown will add an endpoint to our cost-analyzer nginx config for the `/turndown` route.


cost-analyzer/templates/kubecost-turndown-template.yaml
```diff
@@ -0,0 +1,212 @@
...
+  labels:
+    app: {{ template "kubecost.turndownName" . }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "kubecost.turndownName" . }}
+  labels:
+    app: {{ template "kubecost.turndownName" . }}
+rules:
```
dwbrown2: can we add a very visible comment to say that all of these permissions are optional and not enabled by default? 


i propose something like: "These permissions are disabled by default and are optionally enabled with the Kubecost turndown feature." And then link to docs...

PR #213, opened 2020-01-28
saml fixes

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -30,9 +30,18 @@ saml: # enterprise key required to use
...
-    userTypeAdminValue: "admin"
-
+    groups:
+      - name: admin
+        enabled: false # if admin is disabled, all SAML users will be able to make configuration changes to the kubecost frontend
+        assertionName: "http://schemas.auth0.com/userType" # a SAML Assertion, one of whose elements has a value that matches on of the values in assertionValues
+        assertionValues:
+          - "admin"
+          - "superusers"
+      - name: readonly
```
dwbrown2: do you want to include this group in our default saml? proposal is that we could just describe this option in our docs, in an effort to keep simpler for users not using.

PR #264, opened 2020-01-30
Thanos Env + Adding Role and RoleBinding for Pods/Log

mbolt35:
Add Role and RoleBinding to our existing service account to access pod logs for the product Release.Namespace


cost-analyzer/templates/cost-analyzer-cluser-role-template.yaml
```diff
@@ -1,3 +1,19 @@
+# Log Access restricted to kubecost release namespace
+apiVersion: rbac.authorization.k8s.io/v1
```
dwbrown2: Lgtm, but can we also potentially add a comment on why we are getting these? Maybe something like this:

```
# Logs Access used for Kubecost bug report feature -- can be removed if desired 
```

mbolt35: Going to move to a values.yaml config so we can document better and cleanly control enabled/disabled.

AjayTripathy (approved):

PR #267, opened 2020-02-01
Access-Control-Allow-Origin * to headers via nginx

mbolt35:
This seems appropriate, and also adds to 400/500 responses. Does this conflict with SAML work?

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -41,6 +41,8 @@ data:
         add_header Cache-Control "max-age=300";
         add_header Cache-Control "must-revalidate";
         add_header ETag "{{ $.Chart.Version }}";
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -41,6 +41,8 @@ data:
         add_header Cache-Control "max-age=300";
         add_header Cache-Control "must-revalidate";
         add_header ETag "{{ $.Chart.Version }}";
+        add_header 'Access-Control-Allow-Origin' * always;
```
dwbrown2: @AjayTripathy immediate use case for this is to do things like determining if turndown is enabled from the frontend. Given the current config, all 404s are blocked by CORS policy.

AjayTripathy:
I'll test with the SAML integration, but LGTM. It shouldn't change anything there.

AjayTripathy (approved):

PR #272, opened 2020-02-04
bump version, revert nginx headers

AjayTripathy:


AjayTripathy:
Brought headers back after a bit more testing.

AjayTripathy:
That header breaks multicluster frontend view, as the backend server already has the header attached.

PR #273, opened 2020-02-04
Allow passing Spot Instance Configuration in values.yaml

gopisaba:


AjayTripathy:
Thanks for this PR! I'm going to go ahead and merge it-- but, I will need to remove the secret from the configmap and values. That will still need to be manually supplied until we make it readable from a kubernetes secret-- expect that in an upcoming release.

PR #278, opened 2020-02-06
Remove fixed time range from panel

minac:
The `Deployment/Statefulset/Daemonset utilization metrics` panel, has by default a range of time on Feb 2019, which is misleading.

This changes that time range to the last 24h.

The spacing is changed because of my browser's settings. Feel free to implement this with your own PR if you agree with the idea.


dwbrown2:
Thanks for this PR!! Our product actually overrides these windows with url parameters so I hadn't noticed. Let me just test this + your other PR quickly, but in general lgtm.

PR #280, opened 2020-02-10
support renaming container->container_name for compatibility with all…

AjayTripathy:
… versions of cadvisor

Testing notes.
No effect on 1.14. No effect on 1.15. Note that in 1.15, both container and container_name exist for backwards compatibility. I don't have access to a 1.16+ cluster yet, so I set [ cluster ] to a different label that exists and noted that the replace was happening.


dwbrown2 (approved):
LGTM, can you share how you've tested this change?

AjayTripathy:
Added testing notes to commit message.

Issue #207, opened 2020-01-22
Remove cluster.local from Prom address given to Grafana


dwbrown2:
```
- access: proxy
      isDefault: true
      name: Prometheus
      type: prometheus
      url: http://kubecost-prometheus-server.kubecost.svc.cluster.local
```

dwbrown2:
@mbolt35 is this a quick fix?

PR #289, opened 2020-02-14
Remove cluster.local

mbolt35:
Fix for https://github.com/kubecost/cost-analyzer-helm-chart/issues/207


dwbrown2:
@mbolt35 looks good so far! but we actually just had a user hit this same issue in thanos. I saw two references in the file below. Any chance we can pull these in to this PR? There may be others but I just found these.

https://github.com/kubecost/cost-analyzer-helm-chart/blob/e2f1972e312fee9f6983126ae38e797efdc49271/cost-analyzer/charts/thanos/templates/query-deployment.yaml#L68

PR #292, opened 2020-02-18
Add labels to job

minac:
Kubecost jobs show up like this:
`kubecost      cost-analyzer-checks-1582047000-txzzm                 0/1     Completed   0          25m`

This PR is to add labels to those jobs so that they don't show up in the `__unallocated__` bucket.


AjayTripathy:
Thanks for this PR!

PR #307, opened 2020-02-26
support custom cron schedule

AjayTripathy:
This makes the cronjob schedule configurable. However, note that if it isn't running around UTC midnight servertime on Sunday, the weekly update email will not be sent.


dwbrown2 (approved):

mikesplainsonos (approved):
Great thanks!

PR #312, opened 2020-03-02
Add support for thanos query url

vishesh92:


AjayTripathy:
Hi @vishesh92, can you help us understand what you are trying to do here? We don't offer support for durable storage in our free version, but our enterprise product does. There’s also a good bit of other configuration work needed to support Thanos, and querying against an arbitrary Thanos version may not work.

vishesh92:
I thought this was available in free version as well. I was thinking of using my own thanos setup instead of the one setup by helm chart.

dwbrown2:
Hi @vishesh92 this unfortunately isn't supported in our free version. Want to send an email to team@kubecost.com to see if there's another way to support your needs?

dwbrown2:
Hi @vishesh92, would you want to discuss different paths forward?

dwbrown2:
Closing now due to inactivity. Let me know if you would like to discuss.

PR #323, opened 2020-03-09
Update the PriorityClass apiVersion

mbolt35:


dwbrown2:
Just wanted to highlight that it looks like this goes back to 1.11 
https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/

PR #329, opened 2020-03-10
Add namespace option for grafana dashboard cms

gracedo:
Adds an option to set a namespace for grafana dashboard configmaps.

For our specific use case, we are using a separate grafana instance that is running in a different ns than kubecost. The grafana that we are running is also run as part of the prometheus-operator helm chart, which is unfortunately a bit more limited than the standalone grafana chart - the dashboards sidecar is unable to scan the entire cluster for dashboards, just the same ns that it is running in. This means that we would need the option to deploy the kubecost grafana dashboards into a different ns.


AjayTripathy:
Thanks again for this improvement!

Issue #327, opened 2020-03-10
Allow users to specify existing EBS volumes


michaelajr:
Allow users to specify existing EBS volumes for the Grafana and Cost Analyzer volumes instead of dynamically provisioning them when the charts are installed. This can be accomplished by providing a volume id in values and conditionally adding the volume id in the volume template.

This is important as we don't want to use ephemeral storage for things that contain lots of configuration (like our dashboards and analyzer settings). Dynamically created EBS volumes are ephemeral since they are destroyed when the chart is uninstalled.

dwbrown2:
@michaelajr, it looks like Grafana already supports using an existing PVC claim: 
https://github.com/helm/charts/blob/master/stable/grafana/values.yaml#L198

Does this same approach work for our configs? If so, it should be straightforward to support.

michaelajr:
I think in the past I have seen passing an EBS volumeId in the values - and the chart then creates the PV using the existing volume id. But I suppose this would work too. Just want to use a known pattern so folks are familiar with what to do. Definitely need something so we do not lose everything when the chart release is deleted and recreated. Thanks.

AjayTripathy:
Hi @michaelajr -- I'm looking into this now, and would suggest that the kubecost chart add support for storing its configs in an existingClaim, as grafana and prometheus already do.

The proposed workflow would be as follows:
Outside of the kubecost helm chart:
1) Provision volumes with EBS nodes as you normally do.
2) Create claims for these volumes

Inside the kubecost helm chart:
1) Pass these claims in prometheus.server.persistentVolume.existingClaim  and a newly created persistentVolume.existingClaim for our product configs

The work todo on our end here is support persistentVolume.existingClaim in our kubecost values.yaml, as this is already supported in our prometheus subchart. I'm proceeding with this approach and should have a PR soon, and merged in the next release, but would like to keep you in the loop and make sure this solution works for you. Thanks!

michaelajr:
@AjayTripathy Thanks for jumping on this. This seems to follow the pattern that Grafana and Prometheus use.

I was hoping we could just pass a volume id and the PV and PVC would be created by the chart. I am a strong believer that the chart should make all its needed resources - and removing a chart release - removes those resources. Doing it this way requires some set up before running the chart - and as such - complicates our automation process. But since this is the patterns used by the dependent charts, it would not make sense to introduce a new pattern. 

We will try this out as soon as it is merged.

Thanks again,
M

[...]

PR #331, opened 2020-03-11
fix disabling pvs, add support for custom pvc

AjayTripathy:
* Fixes issue where settings PVC: disabled was doing nothing
* Allows using an existing PVC


AjayTripathy:
Supports #327

PR #339, opened 2020-03-18
opensource basic tls

AjayTripathy:


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -54,7 +54,13 @@ data:
         add_header Cache-Control "max-age=300";
         add_header Cache-Control "must-revalidate";
         add_header ETag "{{ $.Chart.Version }}";
-
+{{- if .Values.kubecostFrontend.tls }}
+{{- if .Values.kubecostFrontend.tls.enabled }}
+        ssl_certificate     /etc/ssl/certs/kc.crt;
+        ssl_certificate_key /etc/ssl/certs/kc.key;
+        listen 443 ssl;
```
nikovacevic: Consider moving `listen 9090;` to an else block here to prevent http access on 9090.

nikovacevic (approved):
Looks good. It'd be a nice bonus to auto redirect http to https so that we don't send 400 Bad Request unless you manually type http://... but this doesn't seem to work:
```nginx
if ($scheme = http) {
  return 301 https://$host$request_uri;
}
```
Kind of makes sense, given that this nginx is behind a port-forward or a service, so perhaps redirecting requests like that would require using the nginx Ingress Controller. Not certain about that.

PR #340, opened 2020-03-18
Add namespace option for grafana datasource cm

gracedo:
Doing the same as #329 but for grafana datasource cm


gracedo:
Just saw https://github.com/kubecost/cost-analyzer-helm-chart/pull/343 - doing the same in the latest commit as well

AjayTripathy:
Awesome! Thank you so much for this!

PR #341, opened 2020-03-18
upgrade prometheus

AjayTripathy:


gracedo (approved):
Ran a diff, the only significant changes I saw were to the deployment templates
```
-  {{- if eq .Values.server.strategy.type "Recreate" }}
-  rollingUpdate: null
-  {{- end }}
```
and values
```
--- a/stable/prometheus/values.yaml
+++ b/stable/prometheus/values.yaml
@@ -524,6 +524,9 @@ server:
   enabled: true
   name: server
   sidecarContainers:
+  strategy:
+    type: Recreate
+    rollingUpdate: null
```

Do you mind explaining this a bit, since it looks like it's accomplishing the same thing? Otherwise this lgtm, thanks for the upgrade!

AjayTripathy:
@gracedo I think there is a yaml spacing bug in stable/prometheus (and that rollingUpdate: null change is very recent). I'm collecting info to file an issue there but this seemed to work in the meantime, and I don't want to block this upgrade on that, so it seemed easier to fork for now

PR #344, opened 2020-03-19
Thanos Query Service Override

mbolt35:
Update cost-analyzer deployment to allow a thanos query service override without enabling thanos


mbolt35:
The only other way I could see this working is if we add something like:

```yaml
global:
  externalThanos:
    enabled: true
    queryService: http://my-external-thanos
```

This seemed unnecessary though.

PR #345, opened 2020-03-20
Templatize prom server/alertmanager service names

gracedo:
And use `fullnameOverride` if supplied in config

`{{ .Release.Name }}-prometheus-server` needs to be truncated if it is >63 characters


AjayTripathy:
Thanks for this Grace!

PR #368, opened 2020-04-02
Add configmap_key_ref option to use configmap to set cluster_id

gracedo:
Instead of a predefined cluster_id value, there should be an option to set this dynamically. I've added the option here to use a user-generated configmap which should contain the `CLUSTER_ID` key used to set the external label. User is responsible for generating the configmap prior to installation. Let me know if this makes sense or if there is a better way to work this in!


AjayTripathy:
LGTM, thanks for this!

PR #394, opened 2020-04-16
Add Port and TargetPort to Service Values

mbolt35:
Add `.Values.service.port` and `.Values.service.targetPort` -- propagate through templates.


dwbrown2:
@quangngotan95 does this look good on your end?!

AjayTripathy:
LGTM!

PR #410, opened 2020-04-22
[bug fix] Move clusterIDConfigmap out of prom global config

gracedo:
```
err="error loading config from \"/etc/config/prometheus.yml\": couldn't load configuration (--config.file=\"/etc/config/prometheus.yml\"): parsing YAML file /etc/config/prometheus.yml: yaml: unmarshal errors:\n  line 2: field clusterIDConfigmap not found in type config.plain"
```
Looks like Prometheus will error if there are unsupported options included in the configuration file. I moved the problematic `clusterIDConfigmap` key up one more level so it shouldn't be unintentionally pulled into any prometheus configs. Feel free to move it elsewhere (potentially outside of `prometheus` block if it makes more sense and is less confusing).

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -248,15 +248,15 @@ spec:
             {{- end }}
             {{- end }}
             {{- end }}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
I'm fine to leave it in the prometheus block-- LGTM!

PR #411, opened 2020-04-22
Add matchLabels option to thanos services

gracedo:
I will [upstream](https://github.com/banzaicloud/banzai-charts/pull/1086) this as well. Added `matchLabels` option to all the thanos services to be able to configure extra selectors. This is necessary for our use case because we have multiples of the same thanos components running in a namespace.


AjayTripathy:
LGTM!

PR #418, opened 2020-04-24
Ajay tripathy saml upload

AjayTripathy:


nikovacevic (commented):
Not sure, but looks like there are non-SAML changes here

AjayTripathy:
Ah, this was based off an earlier branch and happened to have both. Those already made it into master. See new diff.

nikovacevic (approved):
LGTM

PR #424, opened 2020-04-28
Add matchLabels option for thanos deployments

gracedo:
@AjayTripathy @dwbrown2 

Upstreamed in https://github.com/banzaicloud/banzai-charts/pull/1086


gracedo:
@AjayTripathy sorry forgot to apply one fix, did it in upstream but forgot to make them here, updated!

PR #454, opened 2020-05-08
support thanos as datasource

AjayTripathy:


AjayTripathy:
cc @gracedo -- don't think this interferes with the way you're planning on attaching grafana but wanted to double check.

PR #462, opened 2020-05-12
Ajay tripathy addpsp

AjayTripathy:
Adding a pod security policy for network costs. The proposal is that this forms the template for a pod security policy for the main cost-analyzer pods as well.

A few questions for consideration:
1) What, if any of these contexts need to be configurable? Note that the network costs pod must run as root, or be disabled.
2) Any issues with using the same service account for network costs and the main kubecost pods, and controlling policies via the binding?


AjayTripathy:
Moved to #463

PR #463, opened 2020-05-12
pod security policy templates

AjayTripathy:


cost-analyzer/templates/cost-analyzer-psp.template.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.podSecurityPolicy }}
+{{- if .Values.podSecurityPolicy.enabled }}
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+    name: kubecost-cost-analyzer-psp
+spec:
+    privileged: true
```
dwbrown2, line 8: is this needed for cost-analyzer pod?

dwbrown2 (approved):

cost-analyzer/templates/cost-analyzer-psp.template.yaml
AjayTripathy: It may not need to be that high, however in some instances (for port binding to 80) it is. We can look to support lowering conditionally (if no root priverleges are needed) this at a later date.

PR #468, opened 2020-05-18
Add ETL enable/disable flag

nikovacevic:
Depends on https://github.com/kubecost/kubecost-cost-model/pull/137

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -83,6 +83,7 @@ kubecostModel:
   imagePullPolicy: Always
   warmCache: true
   warmSavingsCache: true
+  etl: true
   resources:
     requests:
       cpu: "200m"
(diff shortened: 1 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -83,6 +83,7 @@ kubecostModel:
   imagePullPolicy: Always
   warmCache: true
   warmSavingsCache: true
+  etl: true
```
dwbrown2, line 86: what about etlEnabled? I'm supportive either way, but just wanted to throw out the suggestion.

nikovacevic: I'm open to other names, but I think we should be consistent with "Enabled" as a suffix and add that to all of the enable/disable flags, or else have it on none of them.

dwbrown2: Your call... happy to keep as is

PR #474, opened 2020-05-20
Update values.yaml to show node-exporter enabled flag

dwbrown2:
Make it more clear how to disable node-exporter.


mbolt35 (approved):
LGTM

PR #491, opened 2020-05-28
Ajay tripathy saml fix

AjayTripathy:
So-- if you do not use the grafana frontend proxy, ie have byo-grafana, this fails because this was dumped in that block. It does not fail right away because nginx needs to restart to pick up new configmap changes.

Testing as follows:
set global.grafana.proxy = false
helm upgrade
note that things still work
restart the cost-analyzer-pod
note that /api is now unreachable
Apply the fix
helm upgrade
/api still unreachable
restart the cost-analyzer-pod
/api now reachable.


nikovacevic (approved):
LGTM

PR #500, opened 2020-06-03
Ajay tripathy cache

AjayTripathy:
Testing:
1. Manually inspect headers of requests
2. Look at page load times
3. Check that pages load from disk cache where appropriate
4. Upgrade, refresh (to force revalidation) and note the etag changes/page is redownloaded.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -72,7 +72,7 @@ data:
         listen {{ $nginxPort }};
 {{- end }}
         location ~* \.(js|html|css|png|jpg|jpeg|gif)$ {
-            add_header Cache-Control "max-age=43200";
+            add_header Cache-Control "max-age=604800";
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -71,6 +71,15 @@ data:
 {{- else }}
         listen {{ $nginxPort }};
 {{- end }}
+        location ~* \.(js|html|css|png|jpg|jpeg|gif)$ {
+            add_header Cache-Control "max-age=43200";
```
dwbrown2, line 75: This is 30 days correct? Seems fine, but could also probably go longer...

dwbrown2 (approved):

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
AjayTripathy: This is actually 12h. We can increase it though.

PR #501, opened 2020-06-04
Ajay tripathy gcp key

### LINKED ISSUE #498: Support adding GCP credentials at install time
We should support the upload of gcp keys at install time-- today they still needs to be added via the frontend. The uploaded JSON file should be created and read from a secret.

AjayTripathy:
Fixes #498 

Add a secret to the same name as kubecost like this:

kubectl create secret generic gcp-secret --from-file compute-viewer-kubecost-key.json -n kubecost

the secret has form 

```
{
  "type": "service_account",
  "project_id": "",
  "private_key_id": "07ea38ce0e4969ec8be4883d9ffe6941c478165c",
  "private_key": "-----BEGIN PRIVATE KEY-----foo-----END PRIVATE KEY-----\n",
  "client_email": "compute-viewer-kubecost@guestbook-227502.iam.gserviceaccount.com",
  "client_id": "103549887095203286466",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/compute-viewer-kubecost%40guestbook-227502.iam.gserviceaccount.com"
}
```

and can be created via the following steps:

```
export PROJECT_ID=$(gcloud config get-value project)
gcloud iam service-accounts create compute-viewer-kubecost --display-name "Compute Read Only Account Created For Kubecost" --format json
gcloud projects add-iam-policy-binding $PROJECT_ID --member serviceAccount:compute-viewer-kubecost@$PROJECT_ID.iam.gserviceaccount.com --role roles/compute.viewer
gcloud projects add-iam-policy-binding $PROJECT_ID --member serviceAccount:compute-viewer-kubecost@$PROJECT_ID.iam.gserviceaccount.com --role roles/bigquery.user
gcloud projects add-iam-policy-binding $PROJECT_ID --member serviceAccount:compute-viewer-kubecost@$PROJECT_ID.iam.gserviceaccount.com --role roles/bigquery.dataViewer
gcloud iam service-accounts keys create ./compute-viewer-kubecost-key.json --iam-account compute-viewer-kubecost@$PROJECT_ID.iam.gserviceaccount.com
```

Tested by adding a secret and checking it is placed in the correct filepath.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -380,7 +380,7 @@ kubecostProductConfigs:
 #  athenaTable: "athena_test1"
 #  projectID: "123456789"
 #  gcpSecretName: gcp-secret # Name of a secret representing the gcp service key
-#  biqQueryBillingDataDataset: billing_data.gcp_billing_export_v1_01AC9F_74CF1D_5565A2
+#  bigQueryBillingDataDataset: billing_data.gcp_billing_export_v1_01AC9F_74CF1D_5565A2
 #  labelMappingConfigs:  # names of k8s labels used to designate different allocation concepts
 #    enabled: true
 #    owner_label: "owner"
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -333,7 +333,7 @@ serviceAccount:
 # These configs can also be set from the Settings page in the Kubecost product UI
 # Values in this block override config changes in the Settings UI on pod restart
 #
-# kubecostProductConfigs:
+kubecostProductConfigs:
 # An optional list of cluster definitions that can be added for frontend access. The local
 # cluster is *always* included by default, so this list is for non-local clusters.
 # Ref: https://github.com/kubecost/docs/blob/master/multi-cluster.md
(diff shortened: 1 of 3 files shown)


cost-analyzer/values.yaml
```diff
@@ -333,7 +333,7 @@ serviceAccount:
 # These configs can also be set from the Settings page in the Kubecost product UI
 # Values in this block override config changes in the Settings UI on pod restart
 #
-# kubecostProductConfigs:
+kubecostProductConfigs:
```
dwbrown2, line 336: is this intentional? actually feels like a potential improvement if it doesn't cause errors

dwbrown2 (approved):

jdn-za:
If the secret is applied after the initial deployment of the cost-analyzer chart, would a simple restart of the kubecost-cost-analyzer be required?

AjayTripathy:
Yes-- a restart is required to pick up the new secret.

PR #503, opened 2020-06-07
add useEtlCache

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -85,6 +85,7 @@ kubecostModel:
   imagePullPolicy: Always
   warmCache: true
   warmSavingsCache: true
+# useEtlCache: false
```
dwbrown2, line 88: I don't see any reason to add this, but otherwise LGTM

dwbrown2 (approved):

PR #511, opened 2020-06-11
remove init pod

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -26,6 +26,14 @@ global:
   podAnnotations: {}
     # iam.amazonaws.com/role: role-arn
 
+pricingCsv:
```
dwbrown2, line 29: would it help to make this more generic? e.g. `customPricingProvider`

AjayTripathy: Think the specific is preferred? customPricingProvider is an implementation detail.

PR #517, opened 2020-06-18
Use default-kubecost datasource in grafana dashboards

### LINKED ISSUE #516: Default datasource can overlap with other default datasources
"default" datasource may collide with people using other default datasources. Instead let's change to "kubecost-default" or similar and dynamically set the name of the datasource we'd like to use:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/grafana-datasource-template.yaml#L47

cc @gracedo

gracedo:
Fixes #516 
Also added tags to the attached-disks.json dashboard and fixed all the `null` datasource references.

@dwbrown2 @AjayTripathy

### DIFF
--- a/cost-analyzer/attached-disks.json
+++ b/cost-analyzer/attached-disks.json
@@ -356,7 +356,11 @@
   ],
   "schemaVersion": 16,
   "style": "dark",
-  "tags": [],
+  "tags": [
+    "cost",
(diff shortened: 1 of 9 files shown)


AjayTripathy:
LGTM! Thanks

PR #526, opened 2020-06-30
check if thanos enabled before setting it's value

omerlh:
before this change, it will always set this values


AdamStack18:
@omerlh  Closing because the issue has been resolved. Please reopen if needed.

PR #536, opened 2020-07-06
Max Query Concurrency

mbolt35:
* Add configuration option for max query concurrency to cost-model values


dwbrown2 (approved):
LGTM! Any reason to believe setting this to 5 could have performance impact with our current Prom configs?

caarlos0 (approved):

PR #540, opened 2020-07-08
Fix env vars values in CronJob to make sure it points to the correct Service name

lebenitza:
Found this problem in `cost-analyzer-checks-template.yaml`:
```
value: {{ template "cost-analyzer.fullname" . }}.{{ .Release.Namespace  }}:9003
```
which resulted in:
```
Error: getaddrinfo ENOTFOUND cost-analyzer.kubecost cost-analyzer.kubecost:9003
```
caused by `cost-analyzer-service-template.yaml` having a totally different name:
```
kind: Service
apiVersion: v1
metadata:
  name: {{ template "cost-analyzer.serviceName" . }}
```

Removed serviceName template and any references to it and replaced them with full name as it was expected in cost-analyzer-checks.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-checks-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-checks-template.yaml
@@ -36,12 +36,12 @@ spec:
 {{ toYaml .Values.kubecostChecks.resources | indent 14 }}
             env:
             - name: COST_ANALYZER_SERVER_ENDPOINT
-              value: {{ template "cost-analyzer.fullname" . }}.{{ .Release.Namespace  }}:9001
+              value: {{ template "cost-analyzer.serviceName" . }}.{{ .Release.Namespace  }}:9001
             - name: COST_ANALYZER_MODEL_ENDPOINT
               {{- if .Values.saml.enabled }}
-              value: {{ template "cost-analyzer.fullname" . }}.{{ .Release.Namespace  }}:9004
+              value: {{ template "cost-analyzer.serviceName" . }}.{{ .Release.Namespace  }}:9004
               {{- else }}
-              value: {{ template "cost-analyzer.fullname" . }}.{{ .Release.Namespace  }}:9003
+              value: {{ template "cost-analyzer.serviceName" . }}.{{ .Release.Namespace  }}:9003
               {{ end }}
             {{- if .Values.kubecostChecks }}
             {{- if .Values.kubecostChecks.debug }}


dwbrown2:
Hi @MihaiAnei, thanks for this description and PR! I would love to learn a bit more about this problem and understand it's cause. I'm not currently able to reproduce.

Can you share what version of Kubernetes you are running? Also, what DNS provider? Anything else you think that would be impacting service discovery here?

MihaiAnei:
Hi @dwbrown2 . I don't think those details matter. The chart is installed with releaseName = cost-analyzer. This results in the Service name being:
```
Name:              cost-analyzer-cost-analyzer
Namespace:         kubecost
Labels:            app=cost-analyzer
                   app.kubernetes.io/instance=cost-analyzer
                   app.kubernetes.io/managed-by=Helm
                   app.kubernetes.io/name=cost-analyzer
                   helm.sh/chart=cost-analyzer-1.61.2
```
but the env vars in the job are
```
    Environment:
      COST_ANALYZER_SERVER_ENDPOINT:     cost-analyzer.kubecost:9001
      COST_ANALYZER_MODEL_ENDPOINT:      cost-analyzer.kubecost:9003
```
this is caused by using 2 different templates (functions) which do not have the same output all the time but we need them to:
```
            env:
            - name: COST_ANALYZER_SERVER_ENDPOINT
              value: {{ template "cost-analyzer.fullname" . }}.{{ .Release.Namespace  }}:9001
```
```
metadata:
  name: {{ template "cost-analyzer.serviceName" . }}
```
which are
```
{{- define "cost-analyzer.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```
```
{{- define "cost-analyzer.serviceName" -}}
{{- printf "%s-%s" .Release.Name "cost-analyzer" | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```
Because serviceName seems a custom function not usually found in helm charts and only used in the `cost-analyzer-service-template.yaml`'s name, I replaced that with `fullName`, as it was expected in COST_ANALYZER_SERVER_ENDPOINT, and removed `serviceName` and any reference to it.

cost-analyzer/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: "1.61.2"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: 1.61.2
+version: 1.61.3
```
AjayTripathy, line 6: We don't want to bump here-- we batch things together for releases and bump all at once.

cost-analyzer/templates/_helpers.tpl
```diff
@@ -62,10 +62,6 @@ Create the fully qualified name for Prometheus alertmanager service.
 {{- end -}}
 {{- end -}}
 
-{{- define "cost-analyzer.serviceName" -}}
```
AjayTripathy, line 65: We'd want to make the change in the cronjob template, to use cost-analyzer.serviceName, not in the helper/by changing all the other files-- so this should be left as-is and these should be changed: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-checks-template.yaml#L38

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -1,4 +1,4 @@
-{{- $serviceName := include "cost-analyzer.serviceName" . -}}
+{{- $serviceName := include "cost-analyzer.fullname" . -}}
```
AjayTripathy, line 1: See above, the change should be made on the cronjob template side.

cost-analyzer/templates/cost-analyzer-ingress-template.yaml
```diff
@@ -1,7 +1,7 @@
 {{- if .Values.ingress -}}
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "cost-analyzer.fullname" . -}}
-{{- $serviceName := include "cost-analyzer.serviceName" . -}}
+{{- $serviceName := include "cost-analyzer.fullname" . -}}
```
AjayTripathy, line 4: See above, the change should be made on the cronjob template side.

cost-analyzer/templates/cost-analyzer-service-template.yaml
```diff
@@ -3,7 +3,7 @@
 kind: Service
 apiVersion: v1
 metadata:
-  name: {{ template "cost-analyzer.serviceName" . }}
+  name: {{ template "cost-analyzer.fullname" . }}
```
AjayTripathy, line 6: See above, the change should be made on the cronjob template side.

cost-analyzer/values.yaml
```diff
@@ -174,7 +174,7 @@ prometheus:
       scheme: http
       dns_sd_configs:
       - names:
-        - {{ template "cost-analyzer.serviceName" . }}
+        - {{ template "cost-analyzer.fullname" . }}
```
AjayTripathy, line 177: See above, the change should be made on the cronjob template side.

MihaiAnei:
> Is there a reason you chose to do it this way?

Not a strong reason. I am used to having `fullName` used everywhere and not a specific template for Service name, for simplicity mostly and same name with the other resources, just like `helm create <chart-name>` does. 

I will try to modify this MR.

PR #544, opened 2020-07-09
fix datasource default

AjayTripathy:
If you have multiple default datasources, behavior on which datasource is actually added to grafana is undefined. We actually supply a default datasource elsewhere if this is set to true, so it is sometimes causing grafana to break.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -327,8 +327,8 @@ grafana:
       label: grafana_dashboard
     datasources:
       enabled: true
-      defaultDatasourceEnabled: true
-      dataSourceName: Prometheus
+      defaultDatasourceEnabled: false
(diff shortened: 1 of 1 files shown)


cost-analyzer/values.yaml
```diff
@@ -327,8 +327,8 @@ grafana:
       label: grafana_dashboard
     datasources:
       enabled: true
-      defaultDatasourceEnabled: true
-      dataSourceName: Prometheus
```
dwbrown2, line 331: how was this line alone not breaking grafana given that we no longer create this datasource?

AjayTripathy: Setting default to false is all that's needed now. dataSourceName would only be for BYO grafana going forward.

nikovacevic (approved):
LGTM

PR #546, opened 2020-07-10
nginx: cache static files for maximum duration

nikovacevic:
## Changes
- Remove global max-age header
- Add expires (equivalent header with nicer syntax) to individual server blocks
- Add catch-all file server location block to the end of the server block with `expire max`

### Do we get the right headers?
Yes, static file caching is maximized
Screenshot from 2020-07-10 14-19-25

Yes, /model/ and /api/ responses are still 5m
Screenshot from 2020-07-10 14-22-20

### Does Grafana work?
Yes
Screenshot from 2020-07-10 14-18-40

### Does ETag bust the cache?
Need to check

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -148,11 +148,11 @@ data:
             proxy_set_header  X-Real-IP  $remote_addr;
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
-{{- if .Values.global.grafana.proxy }}
+    {{- if .Values.global.grafana.proxy }}
         location /grafana/ {
-{{- if .Values.saml.enabled }}
+        {{- if .Values.saml.enabled }}
             auth_request /auth;
-{{- end }}
+        {{- end }}
             proxy_pass http://grafana/;
             proxy_redirect off;
             proxy_http_version 1.1;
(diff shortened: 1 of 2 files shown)


dwbrown2:
Right now, I know we rely on ETAGs to break cache after an app upgrade. Any concerns of that by impacted with this change?

nikovacevic:
@dwbrown2 No, I expect that behavior to be maintained. The ETag header is untouched. We can (and should) still test that, though.

@AjayTripathy I believe the problem we had before with Grafana had to do with ordering of location blocks. Putting the static file server at the end prevents the `location ~ .(js|css|...)` block from preempting the `location /grafana/` block. The `expires 5m` is just nicer syntax for applying the same `CacheControl max-age=...` header we were using.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -171,4 +171,8 @@ data:
             proxy_pass http://model/isAdminAuthenticated;
         }
     {{- end }}
+        location / {
+            expires max;
```
AjayTripathy, line 175: I know we didn't change the ETAG but doesnt this mean we never check an ETag bump unless you explicitly tell Chrome, ie hard refresH?

nikovacevic: I will test. Based on [this](https://web.dev/http-cache/) you are probably right. In that case, what should we do? Make the cache 24 hours? Ultimately, cache busting in React will eliminate this issue altogether, but in the meantime we'll have an interesting balance to strike.

dwbrown2: Niko, what'd you learn in testing?

nikovacevic: Didn't test today, but it's on the list for the week. Happy to move it up in priority.

nikovacevic:
For whenever we get back to this, the plan is: undo `expires max`, but investigate doing **[etag on](http://nginx.org/en/docs/http/ngx_http_core_module.html#etag)** or some other cache revalidation 304

AdamStack18:
@nikovacevic Closing because of age. Please reopen if needed.

PR #547, opened 2020-07-13
Ajay tripathy support ints

### LINKED ISSUE #495: Support custom prices as integers
The monthly to hourly conversion calculations (referenced below) do not work correctly for integers, only floats. We should support both formats.

https://github.com/kubecost/cost-analyzer-helm-chart/blob/966468bfb98a36be7b8633b04a18455bdd9c443b/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml#L15

AjayTripathy:
Fixes #495


mbolt35 (approved):
LGTM

PR #550, opened 2020-07-15
Enabled Helm Values Reporting

mbolt35:
Uses some helm templating wizardry to filter out any `.Values` leaf nodes that are set to `enabled: false`

This:
```yaml
foo:
  enabled: true
  bar: 123
  inner:
    value: 123
    innerInner:
      enabled: false
      anotherThing: 123
    thing: true
```

Is filtered to this:
```yaml
foo:
  enabled: true
  bar: 123
  inner:
    value: 123
    thing: true
```

There is also now a single flag under `reporting`, namely `reporting.valuesReporting` which controls whether or not this template is used. Any users who have issues can simply disable it via `--set reporting.valuesReporting=false`


nikovacevic (approved):
LGTM

PR #554, opened 2020-07-16
Add resources to the init container

AjayTripathy:


dwbrown2:
Can you share how you tested?

Also, I personally do not feel that the TODO is necessary to include.

AjayTripathy:
Tested by helm upgrading these values with/without a resources: request block. I can remove the TODO

cost-analyzer/values.yaml
```diff
@@ -103,10 +103,6 @@ kubecostModel:
     #  cpu: "800m"
     #  memory: "256Mi"
 
-kubecostInit:
```
dwbrown2, line 106: is there any way this can cause errors, i.e. someone still passing this value?

if not then, LGTM

dwbrown2 (approved):
One question below

cost-analyzer/values.yaml
AjayTripathy: Passing extraneous values is fine.

PR #558, opened 2020-07-20
Thanos Query Offset Configuration

mbolt35:
Add THANOS_QUERY_OFFSET to values and cost-model deployments


dwbrown2:
@gracedo, thought you might like this one! Based on your feedback :) Any input/concerns?

PR #570, opened 2020-07-31
Increase query.max-samples value

dwbrown2:
We've seen a number of users hit this threshold. Increasing should not have a major impact on max memory usage, given our query concurrency. 

This PR still needs to be tested.


AjayTripathy:
LGTM

PR #596, opened 2020-08-31
Ajay tripathy checks quay

AjayTripathy:
Reasoning: These are imagepullpolicy: Always and run ever 10minutes we're getting charged a decent amount for container egress on GKE. I propose we start by migrating checks to quay by default.

Tested by manually pushing, looking at cronjob file:

...
              valueFrom:
                configMapKeyRef:
                  key: prometheus-alertmanager-endpoint
                  name: kubecost-cost-analyzer
            image: quay.io/kubecost1/checks:prod-1.64.0
            imagePullPolicy: Always
            name: cost-analyzer-checks
            resources:
              limits:
                cpu: 100m
                memory: 200Mi
....

And noting:

cost-analyzer-checks-1598904600-qplvx               0/1     Completed   0          5m13s

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -60,7 +60,7 @@ saml: # enterprise key required to use
 
 kubecostChecks:
   enabled: true
-  image: "gcr.io/kubecost1/checks"
+  image: "quay.io/kubecost1/checks"
   resources:
     requests:
       cpu: "20m"


dwbrown2:
Will we now push to both GCR & Quay in our pipeline? That feels ideal in case we need to switch back or users want to keep using that repo / have it hard coded in their values.

AjayTripathy:
> Will we now push to both GCR & Quay in our pipeline? That feels ideal in case we need to switch back or users want to keep using that repo / have it hard coded in their values.

Sure, we can do that. I'll change our internal buildscripts to do both.

dwbrown2 (approved):
LGTM -- but I worry this may cause issues for users that are setting imageVersion directly to older releases.

PR #601, opened 2020-09-02
Add 5m Interval to Savings Recording Rules

mbolt35:


dwbrown2:
let's discuss this one! looks like there is one recording rule with `kube_pod_container_resource_requests_memory_bytes` that we may want to convert to an average over time query.

mbolt35:
> let's discuss this one! looks like there is one recording rule with `kube_pod_container_resource_requests_memory_bytes` that we may want to convert to an average over time query.

Not sure what you mean? This is just an interval update. The rules are already in?

mbolt35:
This change had adverse effects coupling with the resolutions we were using. For now, the default sample interval is fine.

PR #606, opened 2020-09-10
pkey secret

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -489,6 +489,7 @@ serviceAccount:
 #  serviceKeySecretName: "" # Use an existing AWS or Azure secret with format as in aws-service-key-secret.yaml or azure-service-key-secret.yaml. Leave blank if using createServiceKeySecret 
 #  createServiceKeySecret: true # Creates a secret representing your cloud service key based on data in values.yaml. If you are storing unencrypted values, add a secret manually
 #  sharedNamespaces: ""
-#  productKey: # apply business or enterprise product license
+  productKey: # apply business or enterprise product license. Supply as plaintext (this is never used to access your data) or as an existing secret
 #    key: ""
 #    enabled: false
+    secretname: productkeysecret # create a secret of format
```
dwbrown2, line 495: Is this an existing secret? If so, should we just call it `existingProductKeySecret`?

Also, I'm assuming we want this empty of commented out?

AjayTripathy: "existing" is a bit inconsistent with how we do this for serviceKeySecretName. Definitely needs to be commented by default though.

dwbrown2: I personally think how this is currently done is not intuitive.

AjayTripathy:
Ajay Tripathy  1:55 PM
anyway, mind reviewing https://github.com/kubecost/cost-analyzer-helm-chart/pull/606/files to double check all that?

Matt Bolt  2:34 PM
Yeah, that -- that seems fine, but it's assuming a pre-created secret right?

Matt Bolt  2:39 PM
LGTM

PR #616, opened 2020-09-22
Disable priorityclass by default

dwbrown2:
Helm does not handle this API version change properly + PriorityClass is less relevant without Prometheus init block.


AjayTripathy:
LGTM

PR #623, opened 2020-09-28
Ajay tripathy disable ksm secrets

AjayTripathy:
Reduce scope of the default chart permissions by removing KSM secret read permission.


dwbrown2 (approved):
Confirming this will disabled secrets access in this block https://github.com/kubecost/cost-analyzer-helm-chart/blob/966468bfb98a36be7b8633b04a18455bdd9c443b/cost-analyzer/charts/prometheus/charts/kube-state-metrics/templates/clusterrole.yaml#L138-L142

PR #628, opened 2020-09-30
Add PV for kubecost-cost-analyzer-db

nikovacevic:


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -162,6 +178,22 @@ spec:
               mountPath: /var/configs
           securityContext:
             runAsUser: 0
+{{ end }}
+{{- if .Values.supportNFS }}
```
AjayTripathy, line 182: Do you feel strongly about opening up another init container to do this? I'd just do this in a few lines of bash on the same init container.

AjayTripathy (approved):
LGTM!

Issue #622, opened 2020-09-25
Allow datasource file name to be customizable in grafana-datasource configmap


staniondaniel:
Currently the name of the file in the grafana-datasource template is  defined as :
```kubectl get cm grafana-datasource -o yaml
apiVersion: v1
data:
  datasource.yaml: |-
```

When used in conjunction with prometheus-operator  the configuration is ignored because  they use the same name for the datasource filename:
```/usr/share/grafana $ ls -al  /etc/grafana/provisioning/datasources/

-rw-r--r--    1 grafana  grafana        149 Sep 25 15:31 datasource.yaml
```

Depending on the name of the configmap this can also override the default prometheus operator config file with the one from kubecost because the files are processed in alphabetical order 

```
[2020-09-25 15:31:18] Unique filenames will not be enforced.
[2020-09-25 15:31:18] Working on configmap: default/grafana-datasource
[2020-09-25 15:31:18] Working on configmap: default/prometheus-operator-grafana-datasource
```

I would like to be able to customize the name of the file in the configmap so it doesn't conflict with other installations.
 
This can be done by adding a new option in value file and by altering the this template to use the new value
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/grafana-datasource-template.yaml

This problem became more visible after you changed the name of the datasource in Grafana json templates

AjayTripathy:
@staniondaniel this seems reasonable to me. I've opened #640 ... does that satisfy what you need here?

staniondaniel:
> @staniondaniel this seems reasonable to me. I've opened #640 ... does that satisfy what you need here?

Hey! Sorry for the delay. 
Yes this is what i need.

staniondaniel:
Hello 
I thought this was closed. I don't need anything else.  Thank you and closing the ticket

PR #640, opened 2020-10-12
Ajay tripathy allow configurable grafana fname

AjayTripathy:
See #622 for motivation

testing:
helm template kubecost ./cost-analyzer -n kubecost > out.txt generates this when the value is set:

```---
# Source: cost-analyzer/templates/grafana-datasource-template.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: grafana-datasource
  labels:

    helm.sh/chart: cost-analyzer-1.67.1
    app: cost-analyzer
    kubecost_grafana_datasource: "1"
data:
  foo.yml: |-
    apiVersion: 1
    datasources:
    - access: proxy
      name: "default-kubecost"
      isDefault: false
      type: prometheus
      url: http://kubecost-prometheus-server.kubecost```

### DIFF
--- a/cost-analyzer/templates/grafana-datasource-template.yaml
+++ b/cost-analyzer/templates/grafana-datasource-template.yaml
@@ -21,7 +21,7 @@ metadata:
     {{- end }}
     {{- end }}
 data:
-  datasource.yaml: |-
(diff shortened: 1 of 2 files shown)


mbolt35 (approved):
LGTM

PR #641, opened 2020-10-13
Add a choice of NodePort

RossyWhite:
I want to add an option to specify `nodePort` value explicitly.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-service-template.yaml
@@ -37,13 +37,28 @@ spec:
       {{- if .Values.kubecostFrontend.tls.enabled }}
       port: 443
       targetPort: 443
+      {{- if (eq .Values.service.type "NodePort") }}
+      {{- if .Values.service.nodePort }}
+      nodePort: {{ .Values.service.nodePort }}
+      {{- end }}
+      {{- end }}
       {{- else }}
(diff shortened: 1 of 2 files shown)


dwbrown2:
@RossyWhite thanks so much for this! @AjayTripathy

mbolt35 (approved):
Looks good to me!

cost-analyzer/templates/cost-analyzer-service-template.yaml
```diff
@@ -1,5 +1,6 @@
 {{- $nginxPort := .Values.service.targetPort | default 9090 -}}
 {{- $servicePort := .Values.service.port | default 9090 -}}
+{{- $nodePort := .Values.service.nodePort | default 32000 -}}
```
mbolt35, line 3: I think supplying a `nodePort` is optional if `type: NodePort` -  Probably want to avoid specifically choosing a value. Maybe use this as the conditional:
```
{{- if (eq .Values.service.type "NodePort") }}
{{- if .Values.service.nodePort }}
nodePort: .Values.service.nodePort
{{- end }}
{{- end }}

RossyWhite: fixed it !!

AjayTripathy:
Thank you so much for this PR @RossyWhite ! We will get this pushed in our next release!

PR #644, opened 2020-10-13
Use templated query component name instead of release name

gracedo:
When referencing the query service name, use the templated query component name (which is how the query service's name is also generated in its template). This will ensure that name overrides are respected if they are set.


AjayTripathy:
Thanks for the catch @gracedo , LGTM!

PR #666, opened 2020-11-09
fixed typo in reading ERROR_REPORTING_ENABLED config flag value

itscalvinwang:
Description: fixed typo that prevented read of ERROR_REPORTING_ENABLED flag in `cost-analyzer-deployment-template.yaml` from `cost-analyzer/values.yaml`

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -359,7 +359,7 @@ spec:
             - name: PRODUCT_ANALYTICS_ENABLED
               value: {{ (quote .Values.reporting.productAnalytics ) | default (quote true) }}
             - name: ERROR_REPORTING_ENABLED
-              value: {{ (quote .Values.reporting.errorReprting ) | default (quote true) }}
+              value: {{ (quote .Values.reporting.errorReporting ) | default (quote true) }}
             - name: VALUES_REPORTING_ENABLED
               value: {{ (quote .Values.reporting.valuesReporting) | default (quote true) }}
             {{- if .Values.reporting.errorReporting }}


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -359,7 +359,7 @@ spec:
             - name: PRODUCT_ANALYTICS_ENABLED
               value: {{ (quote .Values.reporting.productAnalytics ) | default (quote true) }}
             - name: ERROR_REPORTING_ENABLED
-              value: {{ (quote .Values.reporting.errorReprting ) | default (quote true) }}
+              value: {{ (quote .Values.reporting.errorReporting ) | default (quote true) }}
             - name: VALUES_REPORTING_ENABLED
```
dwbrown2, line 363: Good catch! @mbolt35 are we using the incorrect spelling on the backend?

AjayTripathy: We use VALUES_REPORTING_ENABLED as an env variable to check this, so I doubt it, but worth reading through.

itscalvinwang: I think you mean `ERROR_REPORTING_ENABLED`, but this maps to that in the back end. It just means that this value never gets assigned for the back end to read, so the `env.Get` call won't return anything.

mbolt35: It returns the default, which is set to true above, and true in code as well.

mbolt35: Just so it's clear, in it's current state, this doesn't directly cause anything to break with respect to the actual error reporting on the backend. For example, if users had set errorReporting: false in values, the above value would've been wrong, but further in the template, we have these lines:

```
{{- if .Values.reporting.errorReporting }}
            - name: SENTRY_DSN
              value: "https://71964476292e4087af8d5072afe43abd@o394722.ingest.sentry.io/5245431"
            {{- end }}
```

Which uses the correct spelling, and therefore, sentry would've been disabled (which is the only error reporting we have on the backend). 

However, if the frontend uses clusterInfo.errorReporting to determine whether or not it reports errors, then this would be "true" in all scenarios.

AjayTripathy (approved):

Issue #668, opened 2020-11-10
Please add support for imagePullSecrets K8s feature


esteban1983cl:
**What problem are you trying to solve?**
I need pull images using imagePullSecret kubernetes feature

**Describe the solution you'd like**
Due docker increase pull limits errors, sometimes my pods can't work

**Describe alternatives you've considered**
Add imagePullSecrets in chart template files

**How would users interact with this feature?**
Updating values.yaml file

dwbrown2:
Thanks for this request, Esteban! What containers do you need this for? 

It should be supported for Prometheus, kube-state-metrics, and other deps. Here are current references: https://github.com/kubecost/cost-analyzer-helm-chart/search?q=imagePullSecrets%3A

esteban1983cl:
HI @dwbrown2 I need support for kubecost containers, not prometheus, for example I have the follow containers

gcr.io/kubecost1/checks:prod-1.53.2
gcr.io/kubecost1/frontend:prod-1.53.2
gcr.io/kubecost1/server:prod-1.53.2
gcr.io/kubecost1/cost-model:prod-1.53.2

Thanks

dwbrown2:
@esteban1983cl happy to look at this with you. Also, we'd welcome a contribution if you are interested.

dwbrown2:
@itscalvinwang I believe this is now supported, correct?

Any chance you can add a PR #? 

Also, @esteban1983cl let us know if there's anything else we can get you!

dwbrown2:
My apologies, I see https://github.com/kubecost/cost-analyzer-helm-chart/issues/668 referenced.

Esteban, let us know if this meet yours needs!

PR #685, opened 2020-11-23
support imagePullSecrets on kubecost images

itscalvinwang:
Issue: https://github.com/kubecost/cost-analyzer-helm-chart/issues/668
Description: Add imagePullSecrets field to cost-analyzer helm chart `values.yaml`
Testing: ran `helm template ./cost-analyzer -n kubecost > test.yaml` and verified that the `imagePullSecrets` field is below the `containers` field at the same indentation

Sample outputs: See [comment](https://github.com/kubecost/cost-analyzer-helm-chart/pull/685#issuecomment-732459373) below.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -62,7 +62,7 @@ saml: # enterprise key required to use
         assertionvalues:
           - "readonly"
 
-imagePullSecrets:
+# imagePullSecrets:
 # - name: "image-pull-secret"
 
 kubecostChecks:
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -62,6 +62,9 @@ saml: # enterprise key required to use
         assertionvalues:
           - "readonly"
 
+imagePullSecrets:
+- name: "image-pull-secret"
+
 kubecostChecks:
   enabled: true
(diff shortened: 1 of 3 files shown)


AjayTripathy:
Preferably when we test this way, what we want we want to see is: 
1. the yaml before the change
2. the yaml after the change

Let's also  be thorough on our tests and paste the deployment yaml to a pastebin if it doesn't fit here.

cost-analyzer/values.yaml
```diff
@@ -62,6 +62,9 @@ saml: # enterprise key required to use
         assertionvalues:
           - "readonly"
 
+imagePullSecrets:
```
AjayTripathy, line 65: This should be commented out by default. Can you also test to make sure nothing regresses when this is completely missing?

itscalvinwang:
[before.yaml
](https://gist.github.com/itscalvinwang/632a085f7a154b27e3cf55a274d932d4)
[after.yaml](https://gist.github.com/itscalvinwang/00b1725f0e72c36b2d29e2583f4bf734)

I also ran diff checker on the before.yaml and the generated template with the `imagePullSecrets` field commented out and there is no difference, and the file lengths (by line number) are the same.

AjayTripathy:
LGTM!

PR #691, opened 2020-12-02
pspAppArmor conditional check 

nrolando46:
* Update values.yaml

* Adding to avoid pspAppArmor issues


dwbrown2:
Thanks for the PR, @nrolando46!

Giving to @AjayTripathy for triage/review.

nrolando46:
In our current environment, we don't support the use of AppArmor. As a result, grafana was not getting deployed in our clusters properly. To fix this solution I had to make the changes referenced in the PR to get our deployments working again. 

More importantly, In the latest grafana charts, this change is already in place.

AjayTripathy:
Hi @nrolando46 there's a couple reasons we can't immediately upgrade to the newest grafana charts, but we're discussing internally. Thanks for this change in the meantime, LGTM!

PR #692, opened 2020-12-04
rename configmap key to load db's into grafana properly

AjayTripathy:
Because configmaps are loaded in a random order, sometimes this dashboard would load and sometimes the pod-utilization dashboard would load, because the data key conflicted.

Testing done: Install grafana five or six times to make sure all dashboards load every time


mbolt35 (approved):
LGTM

PR #705, opened 2020-12-14
add read_only flag

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -466,6 +466,9 @@ serviceAccount:
   annotations: {}
   # name: kc-test
 
+
+readonly: false # disable updates to kubecost from the frontend UI
```
dwbrown2, line 470: isn't this more than just disabling from FE? it also disabled update APIs, correct? if so, can we update comment

AjayTripathy: added.

PR #707, opened 2020-12-15
Set readonly to false, commented by default

AjayTripathy:


dwbrown2:
Seems like you could do either comment or set to false, but LGTM

PR #714, opened 2021-01-04
GODEBUG=madvdontneed=1

mbolt35:
* Add flags for thanos store, query, query-frontend
* Add flags for cost-analyzer, network-costs


AjayTripathy:
The code changes themselves LGTM but were you able to test on any scaled environments and notice a difference?

AjayTripathy (approved):

Issue #696, opened 2020-12-08
Pass additional environment variables to pods


AjayTripathy:
**What problem are you trying to solve?**
Support using a proxy for HTTP requests 

**Describe the solution you'd like**
go respects proxy environment variables: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml

**How would users interact with this feature?**
Set environment variables for proxy from helm. Make sure this works even when these values are missing from helm

PR #717, opened 2021-01-06
add http proxy environment variable

calvinwang:
Issue: https://github.com/kubecost/cost-analyzer-helm-chart/issues/696
Description: Support HTTP and HTTPS proxy environment variables.

Testing: helm template for correct generation, ssh into pod and check that the environment variable is correctly set

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -131,8 +131,8 @@ saml: # enterprise key required to use
           - "readonly"
 
 systemProxy:
-  enabled: true
-  useHttps: true
+  enabled: false
+  useHttps: false
   proxyIP: "127.0.0.1"
   proxyPort: "8888"
 
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -130,6 +130,12 @@ saml: # enterprise key required to use
         assertionvalues:
           - "readonly"
 
+systemProxy:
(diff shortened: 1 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -130,6 +130,12 @@ saml: # enterprise key required to use
         assertionvalues:
           - "readonly"
 
+systemProxy:
```
dwbrown2, line 133: Can we reference docs or add comments?

dwbrown2 (commented):
@AjayTripathy would it be helpful to share with Bob to get feedback before merging?

cost-analyzer/values.yaml
itscalvinwang: I've added comments on the behavior (and I've also tested all four cases to ensure correct behavior), i.e., 
1. systemProxy.enabled = false, systemProxy.useHttps = false -- doesn't show up in template
2. systemProxy.enabled = false, systemProxy.useHttps = true -- doesn't show up
3. systemProxy.enabled = true, systemProxy.useHttps = false -- HTTP_PROXY is set
4. systemProxy.enabled = true, systemProxy.useHttps = true -- HTTPS_PROXY is set

let me know if Bob has any feedback to share

AjayTripathy (approved):

PR #721, opened 2021-01-07
support nodeselectors throughout

AjayTripathy:


mbolt35 (approved):
LGTM

kesor:
@AjayTripathy This created a problem for us since we want kubecost to run on specific nodes, but the network daemonset must not use the same selector as the model.

AjayTripathy:
@kesor thanks for the feedback. We'll create a seperate nodeselector for that daemonset.

AjayTripathy:
I've filed #772 to address- meanwhile can you use the old method of taints/tolerations for this to work?

kesor:
> I've filed #772 to address- meanwhile can you use the old method of taints/tolerations for this to work?

meanwhile I removed the `nodeSelector` for network analyzer

PR #724, opened 2021-01-12
add cost change alerting template

itscalvinwang:
Description: add cost change custom alert template to values.yaml, remove `{{ if .Values.global.notifications.alertConfigs.enabled }}` check from configmap so diagnostics endpoint will show alerts enabled as false on helm upgrade

KCM PR: https://github.com/kubecost/kubecost-cost-model/pull/285

Testing: helm template, integration testing with back end changes.


cost-analyzer/values.yaml
```diff
@@ -57,10 +58,14 @@ global:
           window: weekly # or 7d
           aggregation: namespace
           filter: kubecost
-          ownerContact: # ownerContact(s) should be the same for the same namespace, otherwise the last namespace alert overwrites
-            - owner@example.com
-            - owner2@example.com
-
+          # Cost Change Alert
```
dwbrown2, line 61: i'm not a fan of this name. i actually like something like `anomaly` better. do you feel strongly about this current name?

reminder to add to docs once merged!

cost-analyzer/values.yaml
```diff
@@ -57,10 +58,14 @@ global:
           window: weekly # or 7d
           aggregation: namespace
           filter: kubecost
-          ownerContact: # ownerContact(s) should be the same for the same namespace, otherwise the last namespace alert overwrites
```
dwbrown2, line 60: has this changed?

cost-analyzer/values.yaml
itscalvinwang: Sorry it was a working title that I stuck with until now haha. I'm not a fan of CostChange either but I worry that something with "Anomaly" would be confusing for international/ESL users. How do you feel about `spendChange` like in the initial feature design doc?

dwbrown2: I'm on board with that!

itscalvinwang: cool. and sounds good; documentation is the first thing I work on after this gets released!

cost-analyzer/values.yaml
itscalvinwang: it has not; every alert will look to send to its individually scoped recipients before defaulting to globalAlertEmails

itscalvinwang: oh I just removed the example from the ones after the first one to avoid repeating it too much. That behavior will be noted in the documentation.

dwbrown2: all good, you can remove! thanks for clarifying

AjayTripathy (approved):

PR #730, opened 2021-01-13
commented out alertConfigs block for default behavior

calvinwang:
Commented out alertConfigs block in values.yaml to prevent overwrite of UI configured alert settings.


itscalvinwang:
Transform already checks if alerts are enabled before mapping. Closing this as it is unnecessary.

PR #731, opened 2021-01-14
support SAML NameIDFormat config

AjayTripathy:
Testing done: run helm template and note the env variable is set, to be consumed by cost-analyzer pod
helm template ./cost-analyzer > out.txt
cat out.txt
```       ...
            - name: UTC_OFFSET
              value: "+00:00"
            - name: SAML_ENABLED
              value: "true"
            - name: IDP_URL
              value: https://dev-elu2z98r.auth0.com/samlp/metadata/c6nY4M37rBP0qSO1IYIqBPPyIPxLS8v2
            - name: SP_HOST
              value: http://localhost:9090
            - name: NAME_ID_FORMAT
              value: urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified
            - name: CLUSTER_ID
              value: cluster-one
            - name: SQL_ADDRESS
            ...
```

This env variable is consumed by kubecost when it is acting as a Service Provider to set the NameIdFormat.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -470,6 +470,10 @@ spec:
             - name: AUDIENCE_URI
               value: {{ .Values.saml.audienceURI }}
             {{- end }}
+            {{- if .Values.saml.nameIDFormat }}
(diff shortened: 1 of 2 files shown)


mbolt35 (approved):
LGTM

PR #733, opened 2021-01-14
default run as nonroot

AjayTripathy:
Run as nonroot by default.


mbolt35:
Assuming that we've verified the user changes work, and that privileged isn't necessary on the PodSecurityPolicy, this seems fine!

mbolt35 (approved):

PR #735, opened 2021-01-19
Add env var to enable OOC in cost model

michaelmdresser:
See https://github.com/kubecost/kubecost-cost-model/pull/310

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -393,8 +393,8 @@ spec:
               value: "https://71964476292e4087af8d5072afe43abd@o394722.ingest.sentry.io/5245431"
             {{- end }}
             {{- end }}
-            - name: OUT_OF_CLUSTER_ENABLED
-              value: {{ (quote .Values.kubecostModel.outOfClusterEnabled) | default (quote false) }}
+            - name: OUT_OF_CLUSTER_PROM_METRICS_ENABLED
+              value: {{ (quote .Values.kubecostModel.outOfClusterPromMetricsEnabled) | default (quote false) }}
             - name: CACHE_WARMING_ENABLED
               value: {{ (quote .Values.kubecostModel.warmCache) | default (quote true) }}
             - name: SAVINGS_CACHE_WARMING_ENABLED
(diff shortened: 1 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -181,6 +181,7 @@ kubecost:
 kubecostModel:
   image: "gcr.io/kubecost1/cost-model"
   imagePullPolicy: Always
+  outOfClusterEnabled: false
```
dwbrown2, line 184: Looks good, but I propose we make this name a tad more specific. We often refer to outOfCluster as a broad group of functionality. Something like `outOfClusterPromEnabled` or `outOfClusterPromMetricsEnabled` could work?

@AjayTripathy or @michaelmdresser, you may have much better ideas!

AjayTripathy: outOfClusterPromMetricsEnabled seems the right level of specificity, since we have a few different ways of getting/reading out of cluster metrics.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -393,6 +393,8 @@ spec:
               value: "https://71964476292e4087af8d5072afe43abd@o394722.ingest.sentry.io/5245431"
             {{- end }}
             {{- end }}
+            - name: OUT_OF_CLUSTER_ENABLED
```
AjayTripathy, line 396: this env variable name needs to be a bit more specific, as noted below

cost-analyzer/values.yaml
michaelmdresser: Updated.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
michaelmdresser: Updated.

cost-analyzer/values.yaml
```diff
@@ -181,6 +181,7 @@ kubecost:
 kubecostModel:
   image: "gcr.io/kubecost1/cost-model"
   imagePullPolicy: Always
+  outOfClusterPromMetricsEnabled: false
```
AjayTripathy, line 184: Quick doc or comment on what this means?

AjayTripathy (approved):

cost-analyzer/values.yaml
michaelmdresser: Done.

PR #739, opened 2021-01-21
remove health metrics

AjayTripathy:


michaelmdresser:
Is this still a change we want to make?

PR #748, opened 2021-01-25
stop using hostnetwork by default

AjayTripathy:
Using hostnetwork by default here is undesirable as certain k8s flavors don't allow it.


dwbrown2:
What are other implications for changing? I'm imaging it was defaulted to true for a reason?

AjayTripathy:
yeah, actually this appears to be unneeded and undesirable as a default after looking into the issue more.

PR #749, opened 2021-01-26
Net Costs 14.1 and Disable Thanos Sidecar Discoverability by Default

mbolt35:
Update network-costs to 14.1, disable sidecar discovery from thanos querier.


AjayTripathy (approved):

AjayTripathy:
Approved, thank you. Let's look at disabling the sidecar service discovery as well as the sidecar service, but between the two I prefer disabling the service discovery. Ideally we do both but it can be hard to predict who is using what.

PR #751, opened 2021-01-26
switch default auth role

AjayTripathy:


dwbrown2:
This means that users cannot add, edit, or delete data sources, correct?

Webb
Kubecost, Co-founder

ᐧ

AjayTripathy:
I'm testing to see what functionality is disabled, but generally yes.

mbolt35 (approved):
Looks good.

PR #754, opened 2021-01-27
enable diagnostics alerts

AjayTripathy:
Testing done:
When true: I0127 19:05:59.544163       1 log.go:47] [Info] Diagnostics[DpteS]: Running diagnostics checks
When false: no logs


nikovacevic (approved):
LGTM!

PR #760, opened 2021-02-02
Simplify proxying

AjayTripathy:
Use a simpler method of providing a string to the env variables HTTP_PROXY, HTTPS_PROXY, NO_PROXY and make sure to support all three simultaneously.


bob-lockwood (changes requested):
Ajay - this is still not going to work because of the "and" logic in your if statements and the inclusion of .Values.systemProxy.useHttps.  When true this logic will generate two env variables and when false only one.

Please simply further by only having one if block that control all three env variables and remove the "and" logic that is testing .Values.systemProxy.useHttps

Thanks,
Bob

mbolt35 (approved):
LGTM

PR #764, opened 2021-02-05
Add values and secret for Azure storage config

Sean-Holcomb:
added values and secret for accessing Azure Storage service
Tested with configuration ssecret name set in values.yaml
Create secret with command
kubectl create secret generic <SECRET_NAME> --from-file=<STORAGE_CONFIG_JSON>

JSON file should look like:
{
	"azureStorageAccount": "<STORAGE_ACCOUNT_NAME>",
	"azureStorageAccessKey": "<STORE_ACCESS_KEY>",
	"azureStorageContainer": <CONTAINER_NAME>
}

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -613,11 +613,11 @@ serviceAccount:
 #  azureClientID: f2ef6f7d-71fb-47c8-b766-8d63a19db017
 #  azureTenantID: 72faf3ff-7a3f-4597-b0d9-7b0b201bb23a
 #  azureClientPassword: fake key # Only use if your values.yaml are stored encrypted. Otherwise provide an existing secret via serviceKeySecretName
-#  azureStorageAccount: "kubecoststore"
-#  azureStorageAccessKey: ""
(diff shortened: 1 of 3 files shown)


cost-analyzer/values.yaml
```diff
@@ -613,6 +613,11 @@ serviceAccount:
 #  azureClientID: f2ef6f7d-71fb-47c8-b766-8d63a19db017
 #  azureTenantID: 72faf3ff-7a3f-4597-b0d9-7b0b201bb23a
 #  azureClientPassword: fake key # Only use if your values.yaml are stored encrypted. Otherwise provide an existing secret via serviceKeySecretName
+#  azureStorageAccount: "kubecostazurestore"
```
dwbrown2, line 616: should we can comment these fields?

AjayTripathy (commented):
Can you add the exact kubectl commands you used to create a secret for the case where an existing secret is used (maybe with a fake key) to the top of the PR? Then we can easily add that to docs.

AjayTripathy (approved):

PR #765, opened 2021-02-08
Add back podSecurityPolicy.enabled checks

gracedo:
This was removed in https://github.com/kubecost/cost-analyzer-helm-chart/pull/733. With the removal of the checks, the PSP is now always deployed where in the past it was defaulted to false. This has broken our build after upgrading to 1.73.0; in cases where two instances of kubecost are deployed on a cluster, there is a collision of the PSP name - perhaps a fix can be to generate unique names using `{{ template "cost-analyzer.fullname" . }}`, happy to add that in this pr or another if that makes sense.

I've added the checks back in to use the value, I'm not sure if this was intentionally removed or not. Let me know if the value should be defaulted to `true` accompanying this change if the intention was to always deploy it.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-psp.template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-psp.template.yaml
@@ -1,3 +1,5 @@
+{{- if .Values.podSecurityPolicy }}
+{{- if .Values.podSecurityPolicy.enabled }}
 apiVersion: {{ include "cost-analyzer.podSecurityPolicy.apiVersion" . }}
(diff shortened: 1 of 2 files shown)


gracedo:
@AjayTripathy PTAL thank you 🙏

AjayTripathy (approved):
Thanks for this Grace! Expected in our release later this week (1.74.0, currently scheduled for Tuesday evening PST.)

AjayTripathy:
(The default value is intended to be true-- we got feedback from users that they want to make sure this is always deployed)

PR #773, opened 2021-02-11
Template PSP names 

gracedo:
Followup to https://github.com/kubecost/cost-analyzer-helm-chart/pull/765
To prevent name collisions if multiple instances of kubecost may be running in a cluster, template out the name so that they can be unique across multiple charts.


AjayTripathy:
Thanks for this @gracedo !

AjayTripathy (approved):

PR #779, opened 2021-02-18
Disable Alertmanager by default

dwbrown2:
The majority of users are not leveraging Alertmanager and are instead using our in-product alerts. This proposal is to disable it by default.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -327,7 +327,7 @@ prometheus:
     #    value: "value"
     #    effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
   alertmanager:
-    # enabled: false
+      enabled: false
     persistentVolume:
       enabled: true
   nodeExporter:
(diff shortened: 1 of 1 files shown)


dwbrown2:
@AjayTripathy can you help with testing? Also, are there any unintended consequences from this change?

We should very clearly highlight in the associated release.

AjayTripathy:
I'm happy to help with testing the long tail of any unintended consequences, but can you please share what the expected behavior is and what testing has been done so far if any? Have at least these 2 been done?
1. alertmanager is no longer installed by default
2. alertmanager is installed when this is set to true (no regression)

dwbrown2:
I have tested by confirming the following:

1) the current values do not cause any noticeable regressions in the core kubecost application and do disable AlertManager as expected
2) both commenting the line out again and setting this value to true both cause alertmanger to be deployed

@AjayTripathy  please feel free to test any other cases you can think of but this seemed sufficient to me

AjayTripathy (approved):

AjayTripathy:
Yep agreed. we'll get this merged and rolled into our integration tests.

PR #781, opened 2021-02-21
disable legacy api option

AjayTripathy:


cost-analyzer/values.yaml
```diff
@@ -183,6 +183,8 @@ kubecost:
 kubecostModel:
   image: "gcr.io/kubecost1/cost-model"
   imagePullPolicy: Always
+  # Disable the legacy out of cluster costs API. Set enabled for very large cost and usage reports to see cached data on reports.html
+  legacyOutOfClusterAPIDisabled: false
```
dwbrown2, line 187: Does this really belong in default values? I worry this file is being bloated with config parameters that are will be used by very few teams. Would referencing in our docs be better?

AjayTripathy: Ok, removed.

Sean-Holcomb:
LGMT

Sean-Holcomb (approved):

PR #782, opened 2021-02-22
Document kubecostToken

dwbrown2:
Still needs to be tested thoroughly


AjayTripathy:
I'm ok to take over testing but can you please share which features need to be tested here and what the expected behavior should be?

dwbrown2:
It's only regression tested that's needed.

AjayTripathy:
We'll catch it in integration tests then.

PR #787, opened 2021-02-23
Update grafana-datasource-template

Sean-Holcomb:
Cleaned up Grafana datasource template, the main issue however was that .Values.grafana.datasources.dataSourceName was set to default-kubecost which would already be the name of another datasource if that value was being used


Sean-Holcomb:
Screen Shot 2021-02-24 at 4 47 10 PM

cost-analyzer/values.yaml
```diff
@@ -534,7 +534,7 @@ serviceAccount:
 # These configs can also be set from the Settings page in the Kubecost product UI
 # Values in this block override config changes in the Settings UI on pod restart
 #
-#kubecostProductConfigs:
+kubecostProductConfigs:
```
dwbrown2, line 537: this an accident?

cost-analyzer/values.yaml
```diff
@@ -617,7 +617,7 @@ serviceAccount:
 #  azureStorageAccount: "kubecoststore" # Name of Azure Storage account where cost report is being saved
 #  azureStorageAccessKey: "" # Azure Storage account access key found in Azure Storage portal for account where cost report is being saved
 #  azureStorageContainer: "test" # Name of daily month-to-date cost report for Azure
-#  azureStorageSecretName: "azure-storage-config" # Name of Kubernetes Secret where Azure Storage Configuration is stored
+  azureStorageSecretName: "azure-storage-config" # Name of Kubernetes Secret where Azure Storage Configuration is stored
```
dwbrown2, line 620: same comment as above

cost-analyzer/values.yaml
Sean-Holcomb: yep

dwbrown2 (approved):
LGTM, but you may get another review from eng team

Issue #770, opened 2021-02-11
Missing Prometheus annotation inside Network Costs template


pierluigilenoci:
**Describe the bug**

I open this as a result of the tests done on the new [release](https://github.com/kubecost/cost-analyzer-helm-chart/releases/tag/v1.74.0). 

Prometheus does not scrap the Network Costs pod.

Ref: https://github.com/kubecost/cost-analyzer-helm-chart/issues/741#issuecomment-777627784 

**Proposed solution**

Here:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/47af25f922a6bf0c329f25137f5f8817b2cc91d6/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml#L16-L18
is missing this:
```yaml
annotations:
  prometheus.io/scrape: 'true'
```

Manually added to the Daemonset everything started working perfectly.

dwbrown2:
@pierluigilenoci thanks for the great report and suggestion! most teams directly add this scrape when using BYO Prom instead of using service discovery. I definitely see your point though.

@mbolt35 thoughts? seems like a very easy addition, but would this lead to double scraping with our bundled Prom? any other unintended consequences?

mbolt35:
> @mbolt35 thoughts? seems like a very easy addition, but would this lead to double scraping with our bundled Prom? any other unintended consequences?

Yeah, does seem specific to bring your own prom, since our helm chart is configured for the network costs scrape. Does prometheus not recognize the same source? I'll check this out.

dwbrown2:
@pierluigilenoci here's how you can do this today: https://github.com/kubecost/cost-analyzer-helm-chart/blob/6c0975614b4a6854be602d1a6f9506ce8b80abdc/cost-analyzer/values.yaml#L294-L301

@mbolt35 if we do this via service discovery, then I believe we would want to remove the config above.

pierluigilenoci:
@dwbrown2 I prefer a more general solution lowering the impact of external tool configuration changes.
IMHO, a tool should be as autonomous as possible in its configuration, minimizing the impact on the ecosystem in which it is integrated.

pierluigilenoci:
@dwbrown2 however, the solution you are using now is not documented on the [page](https://github.com/kubecost/docs/blob/master/custom-prom.md) for the use of custom Prometheus.

mbolt35:
[...]

Issue #747, opened 2021-01-22
Deprecated cadvisor metrics


pierluigilenoci:
**Describe the bug**
According to [Kubernetes instrumentation guidelines](
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/instrumentation.md) the `pod_name` and `container_name` metrics are deprecated since almost 2 years and not available since K8s 1.16. 

Ref:
* https://github.com/kubernetes/kubernetes/issues/66790
* https://github.com/kubernetes/kubernetes/pull/69099
* https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#metrics-changes

In your integration guide, you suggest changing the default Prometheus configuration to generate them again. https://github.com/kubecost/docs/blob/master/custom-prom.md#cadvisor-metric-labels

I find it wrong to have to force Prometheus configuration to generate the old metrics otherwise the software won't work. It makes more sense that the software uses the official metrics and, in case someone uses a legacy version of Kubernetes (<1.16), they introduce this change in their Prometheus installation. 

Doing so forces those who use a separate installation of Prometheus to use a non-default configuration compared to the official chart.

**Expected behavior**
Do not force the change of third party software configurations to make Kubecost work

dwbrown2:
We introduced these changes to continue supporting legacy versions. I agree we should drop this requirement / move away from this practice.

@nikovacevic @AjayTripathy

AjayTripathy:
Hi @pierluigilenoci -- we will look into this. The main complexity is we want to support both versions seamlessly, as not all of our users can upgrade and it would also be a bad experience to upgrade and for these to all stop working.

pierluigilenoci:
Hi @AjayTripathy, I can understand that you want to keep backward compatibility but it shouldn't be done at the expense of new versions. In the charts, it is the norm, when some changes break the compatibility to announce the breaking changes and a migration guide. Anyone who wants to upgrade the software but still remains at the legacy version of the cluster can apply the workarounds. 

In your chart instead, you impose workarounds and over-configurations on those who keep their clusters updated to favor legacy versions. In this way, all those who want to start using the software in updated clusters make a bad experience.

AjayTripathy:
Yes, we are ok to take this on, just trying to give context on why this change isn't a simple one-liner to relabel.

SergeyKons:
Hi! Is there any progress?

[...]

PR #795, opened 2021-02-24
Network Costs Prometheus Discovery Scrape Annotations

mbolt35:
https://github.com/kubecost/cost-analyzer-helm-chart/pull/795#issuecomment-799735549 for testing done

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
@@ -14,6 +14,13 @@ spec:
       app: {{ template "cost-analyzer.networkCostsName" . }}
   template:
     metadata:
+      annotations:
+{{- if .Values.networkCosts.prometheusScrape }}
+        prometheus.io/scrape: "{{ .Values.networkCosts.prometheusScrape }}"
+        prometheus.io/port: {{ (quote .Values.networkCosts.port) | default (quote 3001) }}
+{{- else }}
+        {}
+{{- end }}
       labels:
         app: {{ template "cost-analyzer.networkCostsName" . }}
         {{- if .Values.networkCosts.additionalLabels }}
@@ -32,7 +39,7 @@ spec:
 {{- end }}
 {{- if .Values.networkCosts.resources }}
         resources:
-{{ toYaml .Values.networkCosts.resources | indent 10 }}
+          {{ toYaml .Values.networkCosts.resources | indent 10 }}
 {{- end }}
         env:
         {{- if .Values.networkCosts.hostProc }}
--- a/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
@@ -14,13 +14,10 @@ spec:
       app: {{ template "cost-analyzer.networkCostsName" . }}
   template:
     metadata:
+      {{- with .Values.networkCosts.annotations }}
       annotations:
-{{- if .Values.networkCosts.prometheusScrape }}
-        prometheus.io/scrape: "{{ .Values.networkCosts.prometheusScrape }}"
-        prometheus.io/port: {{ (quote .Values.networkCosts.port) | default (quote 3001) }}
-{{- else }}
-        {}
-{{- end }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       labels:
         app: {{ template "cost-analyzer.networkCostsName" . }}
         {{- if .Values.networkCosts.additionalLabels }}
--- a/cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml
@@ -1,0 +1,26 @@
+{{- if .Values.networkCosts }}
+{{- if .Values.networkCosts.enabled }}
+{{- if .Values.networkCosts.prometheusScrape }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "cost-analyzer.networkCostsName" . }}
+  annotations:
+    prometheus.io/scrape: "{{ .Values.networkCosts.prometheusScrape }}"
+    prometheus.io/port: {{ (quote .Values.networkCosts.port) | default (quote 3001) }}
+  labels:
+    app: {{ template "cost-analyzer.networkCostsName" . -}}
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+spec:
+  clusterIP: None
+  ports:
+  - name: metrics
+    port: {{ .Values.networkCosts.port | default 3001 }}
+    protocol: TCP
+    targetPort: {{ .Values.networkCosts.port | default 3001 }}
+  selector:
+    app: {{ template "cost-analyzer.networkCostsName" . }}
+  type: ClusterIP
+{{- end }}
+{{- end }}
+{{- end }}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -400,6 +400,8 @@ networkCosts:
     enabled: false
   image: gcr.io/kubecost1/kubecost-network-costs:v15.0
   imagePullPolicy: Always
+  # For existing Prometheus Installs, enable the scrape annotation for network-cost metrics discovery 
+  prometheusScrape: false 
   # Traffic Logging will enable logging the top 5 destinations for each source
   # every 30 minutes.
   trafficLogging: true
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -403,7 +403,10 @@ networkCosts:
     enabled: false
   image: gcr.io/kubecost1/kubecost-network-costs:v15.1
   imagePullPolicy: Always
-  # For existing Prometheus Installs, enable the scrape annotation for network-cost metrics discovery 
+  # For existing Prometheus Installs, create a Service which generates Endpoints for each of the network-costs pods.
+  # This Service is annotated with prometheus.io/scrape: "true" to automatically get picked up by the prometheus config.
+  # NOTE: Setting this option to true and leaving the above extraScrapeConfig "job_name: kubecost-networking" configured will cause the 
+  # NOTE: pods to be scraped twice. 
   prometheusScrape: false 
   # Traffic Logging will enable logging the top 5 destinations for each source
   # every 30 minutes.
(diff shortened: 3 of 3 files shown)


dwbrown2:
cc @michaelmdresser because he's thinking about this in a similarish situation

dwbrown2:
Bolt, were you able to determine if enabling leads to double scraping when our job was already added? 

LGTM, if you are comfortable with how this was tested

cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
```diff
@@ -32,7 +39,7 @@ spec:
 {{- end }}
 {{- if .Values.networkCosts.resources }}
         resources:
-{{ toYaml .Values.networkCosts.resources | indent 10 }}
+          {{ toYaml .Values.networkCosts.resources | indent 10 }}
```
michaelmdresser, line 42: Is this right? There's already an `indent` here.

mbolt35: That's a good catch, going to blame auto format here, but my tests don't seem to generate something improperly formatted after running `helm template`. 

```yaml
      containers:
      - name: release-name-network-costs
        image: gcr.io/kubecost1/kubecost-network-costs:v15.0
        imagePullPolicy: Always
        resources:
          requests:
            cpu: 50m
            memory: 20Mi
          
        env:
        - name: NODE_NAME
```

I was pretty sure that this was going to be an issue, but both helm2 and helm3 generated the yaml with proper indention. Thoughts?

mbolt35: Also worth noting that the `indent` command indents _all_ the output from the pipe source. In this case, all of the yaml gets indented by 4, but I would've still expected the first line to be indented further.

michaelmdresser: Oh, huh. Well, all's well that ends well.

michaelmdresser:
> Bolt, were you able to determine if enabling leads to double scraping when our job was already added?
> 
> LGTM, if you are comfortable with how this was tested

dwbrown2:
@mbolt35 anything left on this PR?

pierluigilenoci:
@mbolt35 when will this PR be released? Because it is not inside the v1.76.0 like you said.

mbolt35:
> @mbolt35 when will this PR be released? Because it is not inside the v1.76.0 like you said.

@pierluigilenoci  Apologies if I mentioned it would make the release. I have some follow up questions here. Based on our helm chart, our prometheus dependency does _not_ have a configuration in place that would consume network costs metrics using these annotations. We would like to add this config, otherwise adding this option to our chart doesn't make sense outside of your specific use case. May I ask what your specific scrape_config looks like to scrape the pods for this annotation?

I drafted the following:
```yaml
- job_name: 'kubernetes-pods'
  kubernetes_sd_configs:
  - role: pod
  relabel_configs:
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]
    action: keep
    regex: true
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
    action: replace
    target_label: __scheme__
    regex: (https?)
  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_path]
    action: replace
    target_label: __metrics_path__
    regex: (.+)
  - source_labels: [__address__, __meta_kubernetes_service_annotation_prometheus_io_port]
    action: replace
    target_label: __address__
    regex: ([^:]+)(?::\d+)?;(\d+)
    replacement: $1:$2
```

Does your configuration look similar?

pierluigilenoci:
Hi @mbolt35,
my Prometheus configuration is the default one of Prometheus
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L1575-L1607

I don't understand why to ask to add a specific config to Prometheus when the vanilla configuration can be used instead. Forcing users to change the configuration of existing installations instead of adapting the software to work across all configurations doesn't seem like the right approach. 

What I asked you to add is the default annotation that Prometheus uses to read metrics from pods. Why not use it instead of complicating life with configurations that are more difficult to maintain?

pierluigilenoci:
For example, if Kubocost stopped using CAdvisor's deprecated metrics, Prometheus' vanilla configuration would work perfectly for you as well. https://github.com/kubecost/cost-analyzer-helm-chart/issues/747 (Please read the whole discussion carefully.)

mbolt35:
> For example, if Kubocost stopped using CAdvisor's deprecated metrics, Prometheus' vanilla configuration would work perfectly for you as well. #747 (Please read the whole discussion carefully.)

@pierluigilenoci Understood. There are several different approaches that are currently in the mix, and we need to choose the best solution for our out of box install. Unfortunately, it's not as simple given the backwards compatibility requirement we have to support for existing users. It appears that another way this is done is by leveraging the `endpoints` config for node-exporter, which we'd need to move to using a `Service` template for network-costs (annotated with `prometheus.io/scrape`). We're still discussing how this can be implemented. Thanks for your patience!

mbolt35:
> Hi @mbolt35,
> my Prometheus configuration is the default one of Prometheus
> https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L1575-L1607
> 
> I don't understand why to ask to add a specific config to Prometheus when the vanilla configuration can be used instead. Forcing users to change the configuration of existing installations instead of adapting the software to work across all configurations doesn't seem like the right approach.
> 
> What I asked you to add is the default annotation that Prometheus uses to read metrics from pods. Why not use it instead of complicating life with configurations that are more difficult to maintain?
> 
> P.S.: you mentioned here [#770 (comment)](https://github.com/kubecost/cost-analyzer-helm-chart/issues/770#issuecomment-785470253)

Our default prometheus config doesn't have a pod scrape target, which is why I asked. It's also why it's important that we move existing users who use our helm installed prometheus to receive the same update.

pierluigilenoci:
@mbolt35 as I wrote somewhere in another comment I prefer a more general solution lowering the impact of external tool configuration changes. A tool should be as autonomous as possible in its configuration, minimizing the impact on the ecosystem in which it is integrated.

I will quote myself again. 😛 

I can understand that you want to keep backward compatibility but it shouldn't be done at the expense of new versions. In the charts, it is the norm, when some changes break the compatibility to announce the breaking changes and a migration guide. Anyone who wants to upgrade the software but still remains at the legacy version of the cluster can apply the workarounds.

In your chart instead, you impose workarounds and over-configurations on those who keep their clusters updated (or using vanilla configuration) to favor legacy versions. In this way, all those who want to start using the software in updated clusters make a bad experience.

mbolt35:
> @mbolt35 as I wrote somewhere in another comment I prefer a more general solution lowering the impact of external tool configuration changes. A tool should be as autonomous as possible in its configuration, minimizing the impact on the ecosystem in which it is integrated.
> 
> I will quote myself again. 😛
> 
> I can understand that you want to keep backward compatibility but it shouldn't be done at the expense of new versions. In the charts, it is the norm, when some changes break the compatibility to announce the breaking changes and a migration guide. Anyone who wants to upgrade the software but still remains at the legacy version of the cluster can apply the workarounds.
> 
> In your chart instead, you impose workarounds and over-configurations on those who keep their clusters updated (or using vanilla configuration) to favor legacy versions. In this way, all those who want to start using the software in updated clusters make a bad experience.

Thanks for your insight.

mbolt35:
Test:
```
$helm template ./cost-analyzer -f ./cost-analyzer/values.yaml --set networkCosts.enabled=true --set networkCosts.prometheusScrape=true > k.yaml
```

Check Output:
```yaml
$ grep "cost-analyzer-network-costs-service*" k.yaml -A 100
# Source: cost-analyzer/templates/cost-analyzer-network-costs-service.yaml

apiVersion: v1
kind: Service
metadata:
  name: release-name-network-costs
  annotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "3001"
  labels:
    app: release-name-network-costs
    helm.sh/chart: cost-analyzer-1.76.0
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Tiller
    app: cost-analyzer
spec:
  clusterIP: None
  ports:
  - name: metrics
    port: 3001
    protocol: TCP
    targetPort: 3001
  selector:
    app: release-name-network-costs
  type: ClusterIP
---
```

This configuration will be picked up by our existing default prometheus configuration: 
```
- job_name: 'kubernetes-service-endpoints'
```

dwbrown2:
@mbolt35 I added a corresponding docs change here: https://github.com/kubecost/docs/pull/73

AjayTripathy (approved):

pierluigilenoci:
❤️

anthosz:
> 



> Test:
> 
> ```
> $helm template ./cost-analyzer -f ./cost-analyzer/values.yaml --set networkCosts.enabled=true --set networkCosts.prometheusScrape=true > k.yaml
> ```
> 
> Check Output:
> 
> ```yaml
> $ grep "cost-analyzer-network-costs-service*" k.yaml -A 100
> # Source: cost-analyzer/templates/cost-analyzer-network-costs-service.yaml
> 
> apiVersion: v1
> kind: Service
> metadata:
>   name: release-name-network-costs
>   annotations:
>     prometheus.io/scrape: "true"
>     prometheus.io/port: "3001"
>   labels:
>     app: release-name-network-costs
>     app.kubernetes.io/name: cost-analyzer
>     helm.sh/chart: cost-analyzer-1.76.0
>     app.kubernetes.io/instance: release-name
>     app.kubernetes.io/managed-by: Tiller
>     app: cost-analyzer
> spec:
>   clusterIP: None
>   ports:
>   - name: metrics
>     port: 3001
>     protocol: TCP
>     targetPort: 3001
>   selector:
>     app: release-name-network-costs
>   type: ClusterIP
> ---
> ```
> 
> This configuration will be picked up by our existing default prometheus configuration:
> 
> ```
> - job_name: 'kubernetes-service-endpoints'
> ```

Hello @mbolt35 ,

It's not really clear how it works.

On my side,  I added networkCosts.prometheusScrape=true + serviceMonitor.networkCosts.enabled=true (same for serviceMonitor.enabled + networkCosts.enabled (services selected)) but what do I need to add on the existing prometheus server to be able to see a cost? (If I replace prometheus by the one included in the kubecost helm chart, it works).

According to the doc, it supposed to be works and no other scrape are mentioned excluding kubecost one...

anthosz:
(fixed)

PR #802, opened 2021-03-04
Add ETL resolution configurations

nikovacevic:
Supports https://github.com/kubecost/cost-model/pull/719 and https://github.com/kubecost/kubecost-cost-model/pull/347

## Changes
- `kubecostModel.etlResolutionSeconds` (i.e. `ETL_RESOLUTION_SECONDS`) determines how high of a resolution ETL should use. Defaults to `60` (i.e. 1m) but should be set to `300` (i.e. 5m) or even `600` (i.e. 10m) for big clients.
- `kubecostModel.etlMaxBatchHours` (i.e. `ETL_MAX_BATCH_HOURS`) limits the window duration of expensive ETL queries, batching them out so that any one query will not time out. Defaults to `6` (i.e. 6h) but might need to be set to `3` (i.e. 3h) for big clients who want to keep high resolution.

## Testing
- Will do a clean install today with helm values set and report back

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -199,7 +199,7 @@ spec:
         {{- end }}
           resources:
 {{ toYaml .Values.initChownData.resources | indent 12 }}
-          {{- if and (.Values.kubecostModel.etlToDisk | default true) .Values.persistentVolume.dbPVEnabled }}          
+          {{- if and (.Values.kubecostModel.etlToDisk | default true) .Values.persistentVolume.dbPVEnabled }}
           command: ["sh", "-c", "/bin/chmod -R 777 /var/configs && /bin/chmod -R 777 /var/db"]
           {{- else }}
           command: ["sh", "-c", "/bin/chmod -R 777 /var/configs"]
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -436,6 +436,10 @@ spec:
             - name: ETL_STORE_DURATION_DAYS
               value: {{ (quote .Values.kubecostModel.etlStoreDurationDays) | default (quote 120) }}
             {{- end }}
+            - name: ETL_RESOLUTION_SECONDS
+              value: {{ (quote .Values.kubecostModel.etlResolutionSeconds) | default (quote 60) }}
```
michaelmdresser, line 440: What's our process for default in the template vs. default in values?

michaelmdresser (approved):
Looks fine.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
nikovacevic: Good question. I provide both to be "safe", but we should decide as a team. Do you have a preference?

AjayTripathy: It is best to provide both so people can easily read values to see the default and users who don't often update their values get a sensible default, in case they are installing from an old values.yaml

nikovacevic: Agreed with that. Only thing left to discuss is what the default values should be. Here, I've opted for ones that prioritize accuracy and hedge against timeouts for medium- to medium-large clusters. But a large customer might still want to make resolution 5m, or even batch 3h, to hedge further. I'd just hate to default to something that, for instance, misses all the checks pods. But open to opposing arguments.

nikovacevic: Agreed to discuss Monday with @dwbrown2 and @AjayTripathy

PR #825, opened 2021-03-22
Add cluster filter + cpu throttle to pod Grafana dashboard

dwbrown2:
Tested by deploying locally in a single cluster setup to confirm throttle panel and filter variable were present.


cost-analyzer/pod-utilization.json
```diff
@@ -781,13 +917,12 @@
     ]
   },
   "time": {
-    "from": "now-24h",
-    "to": "now"
+    "from": "2021-03-22T01:33:44.520Z",
```
AjayTripathy, line 920: Doesn't seem right. can we rever this to now-24h?

AjayTripathy (changes requested):

cost-analyzer/pod-utilization.json
dwbrown2: Should now be fixed

AjayTripathy (approved):

PR #828, opened 2021-03-22
use datasource dropdown in one more spot

andrewgeller:
It looks like one "default-kubecost" datasource was missed in https://github.com/kubecost/cost-analyzer-helm-chart/pull/807


Sean-Holcomb:
@andrewgeller Good catch, thank you!

Sean-Holcomb (approved):

PR #850, opened 2021-04-13
add protocol header to port names for istio support

garreeoke:
When kubecost was deployed on an istio-enabled cluster/namespace, kubecost was not functioning ... frontend would not connect to api. 

Per istio documenation, service ports should be named protocol-name. I did a test on one of my cluster using the changes I made in this PR and it work. I also kept the new port names and tested it successfully on a non-istio enabled cluster/namespace. My testing was minimal, we should definitely test more on the nightly builds.

### DIFF
--- a/cost-analyzer/charts/grafana/templates/service.yaml
+++ b/cost-analyzer/charts/grafana/templates/service.yaml
@@ -37,7 +37,7 @@ spec:
 {{ toYaml .Values.service.externalIPs | indent 4 }}
 {{- end }}
   ports:
-    - name: service
+    - name: tcp-service
       port: {{ .Values.service.port }}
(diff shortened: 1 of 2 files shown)


dwbrown2:
Thanks a ton for investigating, @garreeoke!!

cc @AjayTripathy for suggested next steps!

AjayTripathy (approved):

AjayTripathy:
LGTM!

dwbrown2:
This PR seems like it could have unintended consequences. Do we feel it's been tested sufficiently?

AjayTripathy:
Merging to develop gets this rolled to nightly and easier to pick up, so we'll keep an eye on those

PR #857, opened 2021-04-19
Bump thanos to 0.19

AjayTripathy:
Tested on scaled out cluster. No more duplicate labels series matches as in 0.18!

http://35.232.95.104:9090/diagnostics.html

There are a few timeouts I've been investigating, but on rebuild they disappear/go green.


mbolt35 (approved):
LGTM

PR #858, opened 2021-04-20
feat(network-costs): add affinity

mhulscher:
This PR allows configuring affinity for the network-costs pods. This can be useful for creating anti-affinity rules for certain nodes, for  example EKS Fargate.


AjayTripathy:
Thanks for this PR. @mbolt35 mind taking a look when you get a chance?

AjayTripathy (approved):

PR #862, opened 2021-04-23
Group -> allocationFilter mapping

nealormsbee:
Maybe the template should only run if SAML is enabled as well? This PR represents it as a standalone config setting.

### Testing done

Installed the chart as-is

App behavior: normal as expected

Then uninstalled, edited `values.yaml` to include

```
groupFilters:
  enabled: true
  allocationFilters:
    admin:
      namespace:
        - kubecost
  assetFilters: {}
```

and reinstalled.

App behavior: still normal as expected

Then edited the `auth.go` `Groups()` function to return `[]string{admin}` (short circuit to always get admin group). Built and deployed this image to the cluster.

App behavior: Now the Allocations view aggregated by namespace only shows the `kubecost` namespace + `__idle__` !

Have yet to do end-to-end testing with an actual SAML-based identity provider.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -144,6 +144,14 @@ saml:
         assertionvalues:
           - "readonly"
 
+# Define global data filters for specific user groups
+# No filters defined by default
+# BETA: This implementation is to collect user feedback and may change dramatically.
+# Do not rely on this configuration format remaining consistent.
+groupFilters:
+  allocationFilters: {} # allocation data filters per-group
+  assetFilters: {} # asset data filters per-group
+
 # Adds an httpProxy as an environment variable. systemProxy.enabled must be `true`to have any effect.
 # Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
 systemProxy:
@@ -545,13 +553,6 @@ serviceAccount:
   create: true # Set this to false if you're bringing your own service account.
   annotations: {}
   # name: kc-test
-
-# Define global data filters for specific user groups
-# Disabled by default
-groupFilters:
-  enabled: false
-  allocationFilters: {} # allocation data filters per-group
-  assetFilters: {} # asset data filters per-group
 
 # readonly: false # disable updates to kubecost from the frontend UI and via POST request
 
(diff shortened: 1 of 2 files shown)


dwbrown2:
This is somewhat hard for me to understand as a user. We probably want to document this well. 

Do we support any group name, or are we limited to admin, read-only? It seems teams will likely want many groups.

Could you filter by say a set of clusters and namespaces at the same time?

Also, have we gotten design input from any prospective users? That would likely be helpful.

SheepGotoHeaven:
We can support any group name and filter by clusters, namespaces, services, labels....any filter supported by the Allocation ETL API.

I get why it's hard to understand. Is the right place to document a companion doc in `kubecost/docs`?
And I don't think any users have given input on this design. My read is that this may be more of an initial target to let them react against than the last word.

dwbrown2:
SGTM, your call on easiest place to document... we could start with a Google docs or go straight to `kubecost/docs`. I do think the latter is the right long-term home but can be a little harder for early/fast collaboration.

SheepGotoHeaven:
Makes sense. Alright if I merge this today so we can use it as a strawman for customers?

dwbrown2:
No objections here, but just know that this YAML can be a little harder to change/migrate...

I may consider commenting this block as beta?

cost-analyzer/values.yaml
```diff
@@ -546,6 +546,12 @@ serviceAccount:
   annotations: {}
   # name: kc-test
 
+# Define global data filters for specific user groups
+# Disabled by default
+groupFilters:
```
dwbrown2, line 551: should we have this block near SAML since these are going to be used in conjunction often?

cost-analyzer/values.yaml
```diff
@@ -546,6 +546,12 @@ serviceAccount:
...
+  enabled: false
```
dwbrown2, line 552: Do we need this if the fields are empty? Just feels like more for users to manage...

SheepGotoHeaven: ^ That's a good point

SheepGotoHeaven:
> SGTM, your call on easiest place to document... we could start with a Google docs or go straight to `kubecost/docs`. I do think the latter is the right long-term home but can be a little harder for early/fast collaboration.

@dwbrown2 there's a first draft of a Google doc [here](https://docs.google.com/document/d/18x5KudKdzL1iqMgTz8DgoEwz8nglBU1kcCAxlo-18mY/edit?usp=sharing)

I'll make the other suggested adjustments and resubmit.

Adam-Stack-PM:

AjayTripathy:
I think this has changed enough to no longer be relevant

PR #864, opened 2021-04-25
Disable checks cronjob by default

AjayTripathy:
The checks cronjob is being deprecated in favor of moving our alerting system within the cost-model pod.


dwbrown2:
LGTM, but can you confirm how you tested

PR #865, opened 2021-04-25
Stop warming the legacy cache

AjayTripathy:
Stops warming the legacy cache, which made the default view on the legacy cost-allocation page load quickly. That page has been deprecated, though it is still available by navigating directly to /details.html

This will dramatically decrease prometheus resource utilization.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -195,7 +195,7 @@ kubecostModel:
   # kubecost_cloud_expense_total metrics
   outOfClusterPromMetricsEnabled: false
   # Build local cost allocation cache
-  warmCache: true
+  warmCache: false
   # Build local savings cache
   warmSavingsCache: true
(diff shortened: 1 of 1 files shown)


dwbrown2:
Can you add a bit more the description about the impact, background for this, etc. And can you confirm how this was tested? Let's just make sure no other parts of the application are using this, e.g. namespace page.

nikovacevic (commented):
Yeah, it'd be good to see a list of the places we know that this will affect. I'm happy to draw one up, but (besides detail.html) I'm curious where the known callers of the agg API are. And do users who've integrated with that API know that this is coming?

AjayTripathy:
namespace.html has dependencies on the old, uncached API. Let's pause this for now.

AjayTripathy:
After further examination and discussion, this should be good to go.

nikovacevic (approved):

PR #871, opened 2021-04-29
Correct ServiceMonitor endpoint object

avivgold098:
# Bug fix
The endpoint of the cost-analyzer `ServiceMonitor` points to a non-existing port in the `service` object.
This PR fixes the mismatch between the ServiceMonitor endpoint port, and the service ports names

### DIFF
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.79.0"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: 1.79.0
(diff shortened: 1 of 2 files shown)


avivgold098:
@adri @mbolt35 Hey would like to get your review :)

andreb89:
@AjayTripathy Please take a look as well. This PR needs a Hotfix release as soon as possible. Since the default service monitors of the chart currently do not work.

AjayTripathy (approved):

AjayTripathy:
Thanks all for the fix. We'll plan on releasing a patch tomorrow PST.

AjayTripathy:
Thank you for this PR. We have a patch available, v1.79.1

PR #887, opened 2021-05-07
update the thanos sidecar version to 0.19.0

garreeoke:
We should update sidecar version to match others.


AjayTripathy:
Thanks for this @garreeoke -- sidecar works closely with prometheus and we wanted to get prom up as well so, will require #905 , but pulled this PR into that one!

PR #889, opened 2021-05-11
Update pod-utilization Grafana dashboard

zdenekvicar:
Hello Kubecost maintainers,
I am creating this PR as a follow up after our customer call between Outreach and Kubecost.

At our company, we find the original `pod-utilization` dashboard problematic, because it does not provide an option to show data from all pods, rather just focusing on 1 specific. This becomes a problem when you want to observe requests/limits/usages of a workload containing hundreds of pods.

Suggested upgrade provides:
- Ability to use regex filter into `Pod filter` variable
- Ability to use regex filtering in `Container filter` variable
- Changes `rate` time in queries from `10m` to `3m` to provide more precise data
- Various cosmetic changes ( shared tooltip, legend adjustments, etc ... )

This dashboard is the one we are currently using, and so far our teams seems to be happy with that one - please check this out, maybe it can be useful also to other users & Kubecost community.

PS: I am adding also example of how this dashboard looks when displaying all `coredns` pods data:

Screen Shot 2021-05-11 at 15 01 04

### DIFF
--- a/cost-analyzer/pod-utilization.json
+++ b/cost-analyzer/pod-utilization.json
@@ -15,10 +15,640 @@
   "description": "Visualize your kubernetes costs at the pod level.",
   "editable": true,
   "gnetId": 9063,
-  "graphTooltip": 0,
+  "graphTooltip": 2,
+  "id": 10,
   "iteration": 1616382275479,
   "links": [],
   "panels": [
+    {
+      "aliasColors": {},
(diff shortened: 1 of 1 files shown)


AjayTripathy:
LGTM! Will merge and run our integration tests!

AjayTripathy (approved):

Issue #877, opened 2021-05-03
Grafana dashboard lost the datasource after upgrade to latest chart cost-analyzer-1.79.1


bin3377:
**Describe the bug**
After upgrade the chart to 1.79.1, the Grafana dashboards show no data and the datasource `default-kubecost` is not available. Looks like it's because of the grafana/k8s-sidecar image [version changed](https://github.com/kubecost/cost-analyzer-helm-chart/commit/3ce1905de8f2f62a780a7934f6fd0452188e53de#diff-08c195f9c26798d55c847bae88a9df2a4d8d42fbd31b10029395b9bf1ff7c300) recently. When I manually patched the deploy with earlier image version (7.1.1/0.1.144) the problem is gone.

**To Reproduce**
1. Go to 'Overview'
2. Click on 'CLUSTER METRICS >>' in 'Monthly cluster cost' panel
3. Nothing shown in the grafana dashboard

**Expected behavior**
The metrics are shown in the dasboard

**Screenshots**
![Screen Shot 2021-05-03 at 12 21 30 PM](https://user-images.githubusercontent.com/16766633/116922898-d0f06d00-ac0a-11eb-91b4-a9e808ce2af8.png)

**Collect logs (please complete the following information):**

 - Run `helm ls` and paste the output here:
```
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
kubecost        kubecost        8               2021-05-03 12:19:59.709006 -0700 PDT    deployed        cost-analyzer-1.79.1    1.79.1
```
 - If the pod is stuck in init, run `kubectl logs <kubecost-cost-analyzer pod name> -n kubecost -c cost-analyzer-init` and paste output here:

 - If a page is broken or not loading, open your javascript developer console and paste the ouput here:

 - Run `kubectl logs <kubecost-cost-analyzer pod name> -n kubecost -c cost-analyzer-server` and paste the output here:

 - Run `kubectl logs <kubecost-cost-analyzer pod name> -n kubecost -c cost-model` and paste the output here:

 - Run `kubectl logs <kubecost-cost-analyzer pod name> -n kubecost -c cost-analyzer-frontend` and paste the output here:

AjayTripathy:
Hi @bin3377 , I wasn't able to reproduce this.  Could you try to upgrade grafana and send logs from grafana and grafana-sc-datasources? Here's a sample:

[...]

PR #891, opened 2021-05-11
Add error_throttle_sleep for grafana.sidecar

zdenekvicar:
Should provide fix for https://github.com/kubecost/cost-analyzer-helm-chart/issues/877


AjayTripathy:
Thank you! I tested this for regression and found none-- I am still unable to repro the exact issue, but we will get this out in the next release and hope it helps for users having the issue.

AjayTripathy (approved):

PR #905, opened 2021-05-20
Ajay tripathy upgrade prom thanos

AjayTripathy:
Bump prometheus and thanos-store versions

### DIFF
--- a/cost-analyzer/values-thanos.yaml
+++ b/cost-analyzer/values-thanos.yaml
@@ -28,7 +28,7 @@ prometheus:
     enableAdminApi: true
     sidecarContainers:
     - name: thanos-sidecar
-      image: thanosio/thanos:v0.15.0
+      image: thanosio/thanos:v0.19.0
       securityContext:
         runAsNonRoot: true
         runAsUser: 1001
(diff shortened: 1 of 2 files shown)


cost-analyzer/values-thanos.yaml
```diff
@@ -28,7 +28,7 @@ prometheus:
     enableAdminApi: true
     sidecarContainers:
     - name: thanos-sidecar
-      image: thanosio/thanos:v0.15.0
+      image: thanosio/thanos:v0.19.0
```
mbolt35, line 31: Why just the sidecar here?

mbolt35: Ok, nevermind - I see that sidecar didn't get bumped in the previous update.

AjayTripathy: Yes. Two reasons:

Sidecar interoperates with prometheus closely, so prometheus being significantly behind sidecar seemed bad

Users don't upgrade all clusters simultaneously anyway, so we needed to support some level of interoperability anyway between 0.15 and 0.19.

mbolt35 (approved):
LGTM

PR #907, opened 2021-05-20
ETL: new environment variables

nikovacevic:
Required for https://github.com/kubecost/kubecost-cost-model/pull/400

NOTE: we need to test as many combinations of the new configurations as possible, but several have been unit tested (see PR above)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -442,14 +442,18 @@ spec:
...
-            {{- if .Values.kubecostModel.etlStoreDurationDays }}
-            - name: ETL_STORE_DURATION_DAYS
-              value: {{ (quote .Values.kubecostModel.etlStoreDurationDays) | default (quote 120) }}
-            {{- end }}
             - name: ETL_RESOLUTION_SECONDS
               value: {{ (quote .Values.kubecostModel.etlResolutionSeconds) | default (quote 300) }}
             - name: ETL_MAX_BATCH_HOURS
               value: {{ (quote .Values.kubecostModel.etlMaxBatchHours) | default (quote 6) }}
+            - name: ETL_DAILY_STORE_DURATION_DAYS
+              value: {{ (quote .Values.kubecostModel.etlDailyStoreDurationDays) | default (quote 91) }}
```
michaelmdresser, line 450: Any particular reason for this to default to 91 instead of the 121 that it defaults to here: https://github.com/kubecost/kubecost-cost-model/pull/400/files#diff-d83ea7f33bc8866981a821b46ab46510f0ddcdca122e64c5d71e81de360d03d9R65

michaelmdresser (approved):

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
nikovacevic: I thought we were defaulting to 120d of coverage (requires building 121 days back) but apparently we at some point changed to 90. I'll update the kcm default to match this. Kind of crappy having to keep those two in-sync, but thus is life.

michaelmdresser: > Kind of crappy having to keep those two in-sync, but thus is life.

Indeed :disappointed: config management is so hard.

nikovacevic: Completed here: https://github.com/kubecost/kubecost-cost-model/pull/400/commits/cc521e90eae2f7a5d0c07d8789e10fd1cbb753f5

mbolt35:
https://github.com/kubecost/cost-analyzer-helm-chart/pull/906 is needed with this

PR #910, opened 2021-05-24
Remove deprecated cost-analyzer-checks

AjayTripathy:
We don't use this anymore-- actually delete this instead of simply disabling it.


dwbrown2:
does it also make sense to remove this values block as part of this? https://github.com/kubecost/cost-analyzer-helm-chart/blob/32e8103a1a165b40ee97617def8ef5e70e870001/cost-analyzer/values.yaml#L171-L182

AjayTripathy:
> does it also make sense to remove this values block as part of this?
> 
> https://github.com/kubecost/cost-analyzer-helm-chart/blob/32e8103a1a165b40ee97617def8ef5e70e870001/cost-analyzer/values.yaml#L171-L182

Yep, makes sense. Done!

mbolt35 (approved):
Love it!

dwbrown2:
Conceptually looks good to me. I'll let @mbolt35 take review if that works. 

Also, can we document how we've tested? One question I have is that if you still pass the only config does that cause problems? I see example where users have forked the repo and may accidentally not sync their values to latest.

AjayTripathy:
Testing:

1. Install with file deleted and values unchanged, note the cronjob is deleted
```
atripathy@Ajays-MacBook-Pro builds % kubectl get cronjob -n kubecost
No resources found in kubecost namespace.
```
2. Install with values and file deleted, note upgrade is successful

PR #919, opened 2021-06-02
Replacement metrics to handle Prom v2 deprecation

michaelmdresser:
Adds:
- `kube_node_status_capacity`
- `kube_node_status_allocatable`
- `kube_pod_container_resource_limits`

Done because in KSM v2, the resource-specific versions of those metrics have been removed. We fortunately already had `kube_pod_container_resource_requests` whitelisted.

Also removed a duplicate entry of `kube_node_status_capacity_memory_bytes`

## Testing

Updated a GKE cluster with this chart and confirmed the new/continued existence of:
- [x] `kube_node_status_capacity`
- [x] `kube_node_status_capacity_cpu_cores`
- [x] `kube_node_status_capacity_memory_bytes`
- [x] `kube_node_status_allocatable`
- [x] `kube_node_status_allocatable_cpu_cores`
- [x] `kube_node_status_allocatable_memory_bytes`
- [x] `kube_pod_container_resource_limits`
- [x] `kube_pod_container_resource_limits_cpu_cores`
- [x] `kube_pod_container_resource_limits_memory_bytes`


michaelmdresser:
[Design doc](https://docs.google.com/document/d/1Mr0JpZ4PTjrOMQtOMgTo2PeveB1s2Ukd646uP3tJ3Cw/edit#) (internal)

mbolt35:
I think there are a few spots where we have the white list regex, just making sure that's not the case?

AjayTripathy:
> I think there are a few spots where we have the white list regex, just making sure that's not the case?

Well, there are two seperate whitelists, one for cadvisor and one for service endpoints. This change should only affect KSM/service endpoints so I think it's good to go!

michaelmdresser:
> I think there are a few spots where we have the white list regex, just making sure that's not the case?

The only repeat of the KSM metric whitelist regex is [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/64a25e86a47f75da866f6fde23dfdceedaeb1487/kubecost.yaml#L266-L269); that file is generated from `helm template` according to @AjayTripathy.

AjayTripathy (approved):

PR #923, opened 2021-06-03
Replaced old request metrics with the new metrics

michaelmdresser:
Replaced `kube_pod_container_resource_requests_cpu_cores`
and `kube_pod_container_resource_requests_memory_bytes`
with `kube_pod_container_resource_requests{resource="cpu", unit="core"}`
and `kube_pod_container_resource_requests{resource="memory", unit="byte"}`
respectively.

Necessary because in KSM v2.0+, the old metrics have been removed.
This particular migration is easy because we whitelisted
`kube_pod_container_resource_requests` many months ago, at the same
time we whitelisted the old metrics being replaced in this commit.

Achieved with:
```
fd --type f | xargs sed -i 's/kube_pod_container_resource_requests_memory_bytes{/kube_pod_container_resource_requests{resource="memory", unit="byte", /g'

fd --type f | xargs sed -i 's/kube_pod_container_resource_requests_memory_bytes/kube_pod_container_resource_requests{resource="memory", unit="byte"}/g'

fd --type f | xargs sed -i 's/kube_pod_container_resource_requests_cpu_cores{/kube_pod_container_resource_requests{resource="cpu", unit="core", /g'

fd --type f | xargs sed -i 's/kube_pod_container_resource_requests_cpu_cores/kube_pod_container_resource_requests{resource="cpu", unit="core"}/g'
```

## Testing
See https://github.com/kubecost/kubecost-cost-model/pull/427

PR #924, opened 2021-06-03
Escaped some label strings that were missed

michaelmdresser:
Missed in https://github.com/kubecost/cost-analyzer-helm-chart/pull/923


Sean-Holcomb (approved):
LGTM

PR #929, opened 2021-06-10
Added a section to describe asset reports

nealormsbee:
Template for creating asset reports and a default block in `values.yaml` to show examples

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -116,6 +116,18 @@ global:
         accumulate: true # entire window resolution
         filters: [] # if no filters, specify empty array
 
+  # Set saved report(s) accessible from reports.html
+  # Ref: http://docs.kubecost.com/saved-reports
+  assetReports:
+    enabled: false # If true, overwrites report parameters set through UI
+    - title: "Example Asset Report 0"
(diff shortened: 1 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -116,6 +116,18 @@ global:
         accumulate: true # entire window resolution
         filters: [] # if no filters, specify empty array
 
+  # Set saved report(s) accessible from reports.html
+  # Ref: http://docs.kubecost.com/saved-reports
+  assetReports:
+    enabled: false # If true, overwrites report parameters set through UI
```
SheepGotoHeaven, line 122: On another helm chart PR I opened, I think Webb asked if this were necessary. It doesn't seem like it to me, but the allocation reports and some other things use this pattern and I'm concerned there might be some usecase I'm not imagining.

I guess maybe it's handy to be able to quickly enable / disable things without removing them from your values.yaml?

AjayTripathy: I like these because I think they give you a clear indication that your UI will be overwritten/not used if these are set. But I guess they aren't strictly necessary.

AjayTripathy (approved):

PR #930, opened 2021-06-10
Values options for emitting kube_node/pod_labels

michaelmdresser:
See https://github.com/kubecost/cost-model/pull/830, https://github.com/kubecost/cost-model/pull/831, and https://github.com/kubecost/cost-model/pull/833

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -420,36 +420,9 @@ spec:
             - name: EMIT_NAMESPACE_ANNOTATIONS_METRIC
               value: {{ (quote .Values.kubecostMetrics.emitNamespaceAnnotations) | default (quote false) }}
             {{- end }}
-
-            # A little ugly, see https://github.com/helm/helm/issues/8026 for why it
-            # isn't cleaner.
             {{- if .Values.kubecostMetrics }}
-            - name: EMIT_KUBE_NODE_STATUS_CAPACITY_MEMORY_BYTES_METRIC
-              value: {{ (quote .Values.kubecostMetrics.emitKubeNodeStatusCapacityMemoryBytes) | default (quote true) }}
-            {{- else }}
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -437,6 +437,20 @@ spec:
             - name: EMIT_KUBE_NODE_STATUS_CAPACITY_CPU_CORES_METRIC
               value: "true"
             {{- end }}
+            {{- if .Values.kubecostMetrics }}
```
dwbrown2, line 440: no objections here, but could/should this be made more general? 

e.g. EMIT_KUBE_LABELS_METRIC or EMIT_KSM_METRICS

just trying to help avoid future config overload if possible

michaelmdresser: We could change to something like `EMIT_KSM_V1_METRICS`, which I think is a pretty good idea. I doubt there will be a situation where we want to emit only some of them. I'd like a third opinion, @AjayTripathy / @mbolt35 / @nikovacevic

michaelmdresser: Perhaps `EMIT_NECESSARY_KSM_V1_METRICS`

nikovacevic: `EMIT_KSM_V1_METRICS` is good with me. Seems like we might need a broader discussion around emitting metrics this way, as it could quickly become messy no matter what name we pick.

michaelmdresser: Yes, agreed. I'll get the ball rolling for `EMIT_KSM_V1_METRICS`. Should we discuss this week?

michaelmdresser: https://github.com/kubecost/cost-model/pull/833

nikovacevic (approved):

PR #936, opened 2021-06-15
Update default chunk pool size

AjayTripathy:
Updates the default chunk pool size to be a bit more in line with a mid-size user's needs. Unfortunately when the chunk pool is exhausted, thanos can silently fail so it is better to be a bit high here than too low.


michaelmdresser:
Do we want to review and merge this?

AdamStack18:
@AjayTripathy Closing because of inactivity and age. Please reopen if needed, and I will track down a review.

PR #953, opened 2021-06-25
Fix missing assetReports.reports key

DavidSpek:
The `global.assetReports.reports` key was missing causing the YAML to be invalid.


AjayTripathy:
Hi @DavidSpek thanks! I believe this bug made it into our develop branch last night, which is generally unstable. Super appreciate this though!

AjayTripathy (approved):

PR #961, opened 2021-07-02
Add multi-cloud config secret

Sean-Holcomb:
Required by: https://github.com/kubecost/kubecost-cost-model/pull/445
Related to : https://github.com/kubecost/cost-model/pull/856

Update to mount secret containing multi-cloud-config. Enabling multi-cloud and multi-account cloud resources and reconciliation.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -671,4 +671,4 @@ serviceAccount:
 #    key: ""
 #    enabled: false
 #    secretname: productkeysecret # create a secret out of a file named productkey.json of format { "key": "kc-b1325234" }
-#  multiCloudSecret: "multi-cloud-config"
+#  multiCloudSecret: "multi-cloud-config"#  cloudIntegrationSecret: "cloud-integration"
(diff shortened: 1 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -671,3 +671,4 @@ serviceAccount:
 #    key: ""
 #    enabled: false
 #    secretname: productkeysecret # create a secret out of a file named productkey.json of format { "key": "kc-b1325234" }
+#  multiCloudSecret: "multi-cloud-config"
```
dwbrown2, line 674: Nice! 

Should this this be more generic, i.e. `cloudIntegrationSecret`, and then make this the recommended path going forward for all integrations? 

Have docs yet on how this can/should be used?

Sean-Holcomb: Currently the method of using this secret is separated in such a way that it would be easy to feature gate it. Some things could be moved around to make that more possible, I could definitely change up the name here though. 

I have a start on the docs, but they are not ready yet.

michaelmdresser (approved):

PR #976, opened 2021-07-14
wrap all of the server block in SAML

AjayTripathy:
Kubecost's HTML will load when directly visiting certain views-- no data or APIs will be accessible, but the HTML will load from behind a SAML endpoint. This PR should fix this.

Testing done: Load kubecost with SAML and note no more flashing of unprotected HTML pages.


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -75,8 +75,21 @@ data:
...
+{{- if .Values.saml.enabled }}
+        add_header Cache-Control "max-age=0";
+        location / {
+            auth_request /auth;
+            proxy_redirect off;
+            proxy_http_version 1.1;
+            proxy_set_header Connection "";
+            proxy_set_header  X-Real-IP  $remote_addr;
+            proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
+            error_page 401 = /login;
```
nikovacevic, line 88: Do you know if we want or need `403` to take you to `/login` as well?

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -75,8 +75,21 @@ data:
         root /var/www;
         index index.html;
         large_client_header_buffers 4 32k;
-        add_header Cache-Control "max-age=300";
         add_header Cache-Control "must-revalidate";
+{{- if .Values.saml.enabled }}
+        add_header Cache-Control "max-age=0";
+        location / {
```
nikovacevic, line 81: This feels like a very strange set-up, adding a new `location` (especially a "catch-all" prefix location) inside a conditional like this. That said, I don't know why it wouldn't work. But it may be worth revisiting nginx sooner than later to make something that wraps everything in SAML so that it's more clear how this all works.

nikovacevic (approved):
We don't have a test SAML environment, so I can't test it. But if you've tested thoroughly, then I'm willing to approve the code part of things.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
AjayTripathy: 401 deals with authentication and 403 deals with authorization. 403 implies un-authentication but 403 implies un-authorization, 403s should be reserved for fine-grained authorization failing on already logged-in users. so i think just 401s for now.

PR #977, opened 2021-07-14
Ajay tripathy 1.22 compat

AjayTripathy:
Adds kubeVersion and notes that we currently support all k8s features up to 1.22


mbolt35 (approved):
LGTM

PR #999, opened 2021-07-27
Kubemetrics Addition

mbolt35:
* Allows KSM replacement emitter pod to be deployed optionally
* Added auto-discovery via service annotations
* Added optional service-monitor
* Removes recording rules that are no longer required for savings.


dwbrown2:
Are we hoping to push this soon or is this just for discussion purposes?

cost-analyzer/values.yaml
```diff
@@ -202,6 +202,50 @@ kubecost:
...
+  # emitKsmV1Metrics: true
+  
+  # Optional
+  # The metrics emitter is a separate deployment and service (for prometheus scrape auto-discovery)
+  # which emits metrics cost-model relies on. Enabling this deployment also removes the KSM dependency
+  # from the cost-model. If the deployment is not enabled, the metrics will continue to be emitted from
+  # the cost-model.
+  emitter:
+    enabled: false
+    image: "gcr.io/kubecost1/kube-metrics:v0.0.1"
```
dwbrown2, line 220: I'd be reluctant to introduce another name, i.e. "kube-metrics" isn't this the "kubecost_exporter" that has a "kube-metrics" or "ksm" module?

cost-analyzer/values.yaml
```diff
@@ -202,6 +202,50 @@ kubecost:
     #  cpu: "100m"
     #  memory: "256Mi"
 
+# Kubecost Metrics deploys a separate pod which will emit kubernetes specific metrics required
```
dwbrown2, line 205: aren't we deploying in our current container first? if so, should we wait on values updates?

cost-analyzer/values.yaml
```diff
@@ -409,16 +453,6 @@ prometheus:
               record: kubecost_savings_memory_allocation_bytes
               labels:
                 daemonset: "true"
-            - expr: label_replace(sum(kube_pod_status_phase{phase="Running",namespace!="kube-system"} > 0) by (pod, namespace), "pod_name", "$1", "pod", "(.+)")
```
dwbrown2, line 412: nice! can all recording rules go at this point, or other still being used?!

cost-analyzer/templates/kubecost-metrics-service-template.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.kubecostMetrics }}
+{{- if .Values.kubecostMetrics.emitter }}
+{{- if .Values.kubecostMetrics.emitter.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "kubecost.kubeMetricsName" . }}
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: {{ (quote .Values.kubecostMetrics.emitter.port) | default (quote 9005) }}
```
dwbrown2, line 10: isn't the standard term that this is just a prom "exporter"? here's an interesting resource if you haven't seen: https://prometheus.io/docs/instrumenting/writing_exporters/

cost-analyzer/values.yaml
mbolt35: Not really, trying avoid name conflict here until we have parity between what we currently call the "exporter" (headless cost-model).

cost-analyzer/values.yaml
mbolt35: This is disabled by default, and since the internal version requires updates to env vars, we need to at least template the values.

cost-analyzer/values.yaml
mbolt35: Great question. I'm removing the ones that I know were used in Abandoned Workloads and Request Sizing. I'm not sure where the others are in use.

AjayTripathy: I believe there's a small scale test that needs to be run to double check max() on CPU usage. I don't anticipate that that's a large query though, since we now have the 1h progressive builds :)

PR #1003, opened 2021-07-28
bug fix: solve large headers, cookie problem with nginx

MichaelMeelisRTL:
If the kubecost frontend receives a large header or cookie it will return a "400 request header or cookie too large". By adding the large client header buffer 4 32k this problem will be solved.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -74,6 +74,7 @@ data:
         server_name _;
         root /var/www;
         index index.html;
+        large_client_header_buffers 4 32k;
         add_header Cache-Control "max-age=300";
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -74,6 +74,7 @@ data:
         server_name _;
         root /var/www;
         index index.html;
+        large_client_header_buffers 4 32k;
```
michaelajr, line 77: Should the size be configurable via values?

kirbsauce: Agreed, we should follow up with improved flexibility.  Hopefully this first step alleviates the need for you to maintain a fork.

AjayTripathy (approved):

PR #1013, opened 2021-08-03
add generic cost-analyzer network policy template

benfain:
### Background
The [current cost-analyzer network policy template](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-network-policy.yaml) isn't sufficient for our use case. It's also somewhat non-standard because rather than targeting cost-analzyer pods, it conditionally can target prometheus pods which I haven't seen too often.

From what I gather, the current policy template supports:
* Restricting egress to inside the cluster only
* Restricting ingress to self (cost-analzyer pods)
* Allowing ingress access to prometheus from cost-analzyer pods

For the above, I don't believe you can combine them as well.

We needed to append a few ingress/egress rules to the template and so I wasn't sure where to start with adding to the pre-existing template. My solution in this PR is to add a new template where the rules are generic and the target pods are cost-analyzer specific, leaving the old template for backwards compat purposes.

The general concept was _borrowed_ from https://github.com/ameijer/k8s-as-helm/blob/master/charts/networkpolicy/templates/networkpolicy.yaml -- although I did update the metadata to be specific to this chart.

### Changes
* Add a new network policy template called `cost-analyzer-network-policy-template.yaml` that is disabled by default which dynamically defines ingress/egress rules from values in values.yaml files.
* Add related defaults in `values.yaml`
* Update readme

### Testing
Testing input vars:
```
networkPolicy:
  costAnalyzer:
    enabled: true
    annotations:
      foo: bar
    additionalLabels:
      foo: bar
    ingressRules:
      - selectors: # allow ingress from self on all ports
        - podSelector:
            matchLabels:
    egressRules:
      - selectors: # allow egress to self on all ports
        - podSelector:
            matchLabels:
      - selectors: # allow egress access to prometheus
        - namespaceSelector:
            matchLabels:
              name: prometheus
          podSelector:
            matchLabels:
              app: prometheus
        ports:
          - protocol: TCP
            port: 9090
```

Resulting rendered template:
```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: kubecost-cost-analyzer
  annotations:
    foo: bar
  namespace: kubecost
  labels:
    helm.sh/chart: cost-analyzer-1.83.2
    app: cost-analyzer
    foo: bar
spec:
  podSelector:
    matchLabels:
      app: cost-analyzer
  policyTypes:
    - Ingress
    - Egress
  egress:
    - to:
       - podSelector:
           matchLabels:
      ports:
         null
    - to:
       - namespaceSelector:
           matchLabels:
             name: prometheus
         podSelector:
           matchLabels:
             app: prometheus
      ports:
         - port: 9090
           protocol: TCP
  ingress:
    - from:
       - podSelector:
           matchLabels:
      ports:
         null
```

Null port entries in these policies maps to `<any>` in the final policy from my testing which is desired:
```
      ports:
         null
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-policy-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-policy-template.yaml
@@ -1,0 +1,47 @@
+{{- if .Values.networkPolicy -}}
+{{- if .Values.networkPolicy.costAnalyzer.enabled -}}
+kind: NetworkPolicy
+apiVersion: {{ include "cost-analyzer.networkPolicy.apiVersion" . }}
+metadata:
+  name: {{ template "cost-analyzer.fullname" . }}
+{{- if .Values.networkPolicy.costAnalyzer.annotations }}
+  annotations:
+{{ toYaml .Values.networkPolicy.costAnalyzer.annotations | indent 4}}
+{{- end }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{- if .Values.networkPolicy.costAnalyzer.additionalLabels }}
+{{ toYaml .Values.networkPolicy.costAnalyzer.additionalLabels | indent 4 }}
+{{- end }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "cost-analyzer.selectorLabels" . | nindent 6 }}
+  policyTypes:
+{{- if .Values.networkPolicy.costAnalyzer.ingressRules }}
+    - Ingress
+{{- end }}
+{{- if .Values.networkPolicy.costAnalyzer.egressRules }}
+    - Egress
+{{- end }}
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/cost-analyzer-network-policy-template.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.networkPolicy -}}
+{{- if .Values.networkPolicy.costAnalyzer.enabled -}}
```
AjayTripathy, line 2: Just a quick note that for backwards compatibility we check every level of the object so we'd also want a check for 
.Values.networkPolicy.constAnalyzer, but I can add it for you.

AjayTripathy (approved):

AjayTripathy:
Thanks so much for this @benfain !

PR #1014, opened 2021-08-03
In Zone Documentation Addition for 10.0.0.1/8

mbolt35:


kirbsauce (approved):
Agreed, @mbolt35 , nothing too advanced, but will certainly help those using Multi-AZ Kubernetes realize there is something to consider here.

dwbrown2:
@mbolt35 we still need this for single-AZ clusters though? Seems like it could be dynamically detected... I can't remember the details but I seem to recall we had a specific reason to include.

mbolt35:
> we still need this for single-AZ clusters though? Seems like it could be dynamically detected... I can't remember the details but I seem to recall we had a specific reason to include.

We included this by default because it meets the standards for RFC private address ranges, which typically involve local hardware, but that's a fairly broad assumption now with distributed architecture (Might be unintentionally dating myself). 

I think this also had the advantage of categorizing local traffic we maybe weren't identifying accurately in the early releases of network-costs. It seems that this might be actually causing more issues than it's preventing these days, but I will need to look into it a bit more. It may be as simple as just removing this categorization entry completely.

Issue #1016, opened 2021-08-04
Add namespace metadata to all resources


cten:
**What problem are you trying to solve?**
I do not use Helm, but instead to use `helm template` to output the manifests, then use kustomize to overlay changes. It would be very helpful if the resources each had its namespace metadata set. I can't use the kustomize namespace operator bc I do create other resources that need to go in namespaces other than kubecost. Also unless I deploy kubecost to the monitoring namespace all the grafana dashboards wouldn't be able to be loaded. 

**Describe the solution you'd like**
Set the namespace in each individual resource, this wouldn't be hard to have a variable/key in the values file that can be used to set the namespace for grafana dashboards different from namespace for kubecost specific resources. 

**Describe alternatives you've considered**
Using kustomize doesn't work, or is very risky. Using sed/awk is also very unreliable/risky.

**How would users interact with this feature?**
Same way they do with all settings in the values.yaml file.

gz#649

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-32) by [Unito](https://www.unito.io)

PR #1020, opened 2021-08-05
added ability to set namespace for cost-analyzer resources and overwr…

cten:
added ability to set namespace for cost-analyzer resources and overwrite namespace for thanos resources


AjayTripathy:
Thanks @cten -- looks like this fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1016 . We'll test on our end and look to get this merged soon!

cten:
My team currently is using this branch, it should still support those who use Helm normally. The logic is to explicitly take the namespace used by Helm and set it on all resources. We used this to deploy our kubecost and worked recently to upgrade from 1.84 to 1.85.

kbrwn:
@cten This change did not work for all resources deployed by the cost-analyzer helm chart. I tested several `--set` variations and the desired namespace was not used for prometheus resources.

keithhand:
> @cten This change did not work for all resources deployed by the cost-analyzer helm chart. I tested several `--set` variations and the desired namespace was not used for prometheus resources.

From what I see, the Thanos chart looks changed, but it seems like the Grafana, Prometheus, and KSM sub-charts all still need to be updated before all the resources will deploy correctly.

Adam-Stack-PM:

PR #1039, opened 2021-08-18
Document and enable ETL filestore by default.

AjayTripathy:
Document and enable ETL filestore by default.

@mbolt35 did some benchmarking on our scale test cluster-- could you summarize your results of that here when you get a moment?

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -222,6 +222,8 @@ kubecostModel:
   etlHourlyStoreDurationHours: 49
   # max number of concurrent Prometheus queries
   maxQueryConcurrency: 5
+  # If running kubecost from a PV, greatly reduce memory usage for a small increase in data loading times.
+  etlFileStoreEnabled: true
   # utcOffset represents a timezone in hours and minutes east (+) or west (-)
   # of UTC, itself, which is defined as +00:00.
   # See the tz database of timezones to look up your local UTC offset:


cost-analyzer/values.yaml
```diff
@@ -222,6 +222,8 @@ kubecostModel:
   etlHourlyStoreDurationHours: 49
   # max number of concurrent Prometheus queries
   maxQueryConcurrency: 5
+  # If running kubecost from a PV, greatly reduce memory usage for a small increase in data loading times.
```
dwbrown2, line 225: what happens if you're running Kubecost without an attached PV?

mbolt35:
# Scale Cluster
The scale cluster contained 40 days of data, 4000 nodes, 250k pods. ETL stores (per hour/day) were ~93mb on disk. 

## Memory Usage
General memory usage during queries was observed using the metrics-server and `kubectl top`

### MemoryStore
- 7d Query: 20-30GB
- 30d Query: 40-60GB

### FileStore
- 7d Query: 2-4GB
- 30d Query: 4-6GB

## Query Times
We sacrifice query performance when using file store since it must actively decode entries into memory (instead of cloning) prior to performing operations. 

### Memory Store
7d Query: 3.8s
30d Query; 25.7s

### File Store (Before String Table Optimizations):
7d Query: 8.2s
30d Query: 43.1s

### File Store (After String Table Optimizations):
7d Query: 5.0s
30d Query: 31.3s

## Summary
Generally, smaller clusters may still benefit from using `MemoryStore`, but using `FileStore` should not have any dramatic reduction in performance. `FileStore` does not support `S3` (yet), so that may also be a limitation requiring users to move back to `MemoryStore`.

PR #1043, opened 2021-08-24
chore: Use consistent tags in deployment utilization dashboard

gracedo:
## What does this PR change?
Adds the same "cost", "utilization", "metrics" tags to the `Deployment/Statefulset/Daemonset utilization metrics` grafana dashboard, just to be consistent with the other cost dashboards.
image

No

Different tags displayed in the grafana UI, little to no impact

## How was this PR tested?
Manually imported the updated dashboard json
image


AjayTripathy (approved):
LGTM -- thanks Grace!

Issue #1046, opened 2021-08-26
v1beta1 ClusterRoleBinding and v1beta1 Ingress are unavailable in v1.22+


pierluigilenoci:
**Describe the bug**

It is not possible to install the chart in K8s v1.22.
This is because the chart uses APIs that have been deprecated for some time and removed in version v1.22.

Specifically the Ingress (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122) and the ClusterRoleBinding (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122)

```
rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRoleBinding
networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
``` 

**To Reproduce**

1. Try to install the chart in a cluster running K8s v1.22

**Expected behavior**

Be able to install the chart.

**Screenshots**

Not relevant

**Collect logs (please complete the following information):**

Not relevant

gz#733

pierluigilenoci:
Is this issue fixed?

kirbsauce:
yes, this should be fixed in 1.86.0.

pierluigilenoci:
@kirbsauce I would definitely say no:

<img width="1187" alt="Screenshot 2021-09-09 at 18 39 25" src="https://user-images.githubusercontent.com/36044817/132727129-5a880999-a6b9-43fb-974c-499e3c61d26c.png">

kbrwn:
@pierluigilenoci Closing this out as it should be fixed by a previous release. If you notice otherwise please feel free to reopen. Often times these "is deprecated in" messages are warnings that do not prevent kubecost from proceeding to install with the correct apiVersion type.

PR #1047, opened 2021-08-28
Ajay tripathy 1.22 subchart support

AjayTripathy:
## What does this PR change?
Adds support for k8s 1.22 in subcharts

Users will be able to install the full kubecost stack in 1.22

https://github.com/kubecost/cost-analyzer-helm-chart/issues/1046

## How was this PR tested?
Run helm template and note no regressions, and that an ingress with the right apiversion shows up:
```
# Source: cost-analyzer/charts/prometheus/templates/server-ingress.yaml
apiVersion: networking.k8s.io/v1
```


cost-analyzer/charts/grafana/templates/role.yaml
```diff
@@ -1,6 +1,6 @@
 {{ if .Values.global.grafana.enabled }}
 {{- if .Values.rbac.create }}
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
michaelmdresser, line 3: Will changes like this cause problems on older versions of K8s? I know we technically support quite old versions.

michaelmdresser: Or, to phrase this another way, when did rbac go from `v1beta1` to `v1`?

michaelmdresser (approved):
One small question, otherwise looks good!

cost-analyzer/charts/grafana/templates/role.yaml
AjayTripathy: In this particular case, no, since we don't switch on v1beta for rbac.authorization.k8s.io anywhere else in the chart.

Issue #816, opened 2021-03-12
Inaccurate date offset on kubernetes data


Sean-Holcomb:
User reports offset in data by -1 day on assets page. From user description queries grab correct data, for example in the picture below the custom date range is. from 8th March to 12th March (today). But it shows from 7th - 11th. 

![](https://user-images.githubusercontent.com/78923723/110919956-1a17f500-831d-11eb-9da5-8d9d207fa822.png)

User also notes they are in CET timezone and have the kubecostModel.utcOffset value set to "+01:00" for extended discussion of this bug up to this point see @andreb89 's comments on https://github.com/kubecost/cost-analyzer-helm-chart/issues/414

Sean-Holcomb:
@andreb89 we have ticket for this issue and will be working on it some time soon. For now you should unset the kubecostModel.utcOffset value. If this makes a significant change for you. Please report it back here. Thank you.

andreb89:
@Sean-Holcomb So I removed the kubecostModel.utcOffset value from the deployment. To be sure, I also completely removed the current kubecost installation and run a new fresh installation afterwards. So the good news is that now there are no "empty" days anymore. 
The dashboard now shows data for all days. The current and last day (15th and 16th) show only the internal Kubernetes costs and no out-of-Azure costs like storages, caches, etc.. The day before yesterday (14th) is missing the postgres database costs block (red blocks in the screenshot). 
So overall it's better than before. Is the missing of some costs because of the 48-hour window, that you mentioned before? Is this now the correct behavior or is still something not the way it should be?

![Screenshot from 2021-03-16 14-09-26](https://user-images.githubusercontent.com/78923723/111314260-84080580-8661-11eb-8a59-feaad291a8d1.png)

dwbrown2:
This does seem a little longer than I would expect, but it may be normal given the average Azure billing data delays... @Sean-Holcomb can you confirm based on what you've seen so far?

Is this issue accurately titled given what we've learned?

Sean-Holcomb:
@andreb89 There is actually no window for OOC cost, they are displayed as soon as the they are available in your export CSV,    can you rebuild your ETL and open your most recent cost export CSV and confirm for me that there are line items for the previous day while it shows no OOC costs. In the case of your last image you would be looking for line items on 2021-3-16.

On face value it looks to me like the Azure costs are still being offset by one day. but it may have to do with when Azure is releasing the CSVs and what data is in them.

[...]

PR #1081, opened 2021-09-22
Deprecate UTC Offset

AjayTripathy:
## What does this PR change?
Deprecate UTC offset. It is not uniformly supported across all cloud provider CURs and is leading to confusion.

No

Removes the ability to configure viewing kubecost data in different timezones. Timezone skew is not supported in many cloudprovider's data, making comparisons across CUR data more complex.

https://github.com/kubecost/cost-analyzer-helm-chart/issues/816

## How was this PR tested?
This is the default and we're removing a feature.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -222,11 +222,6 @@ kubecostModel:
   etlHourlyStoreDurationHours: 49
   # max number of concurrent Prometheus queries
   maxQueryConcurrency: 5
-  # utcOffset represents a timezone in hours and minutes east (+) or west (-)
-  # of UTC, itself, which is defined as +00:00.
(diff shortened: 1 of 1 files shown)


nikovacevic (approved):
This will certainly achieve the stated goal. For the record, I'm agnostic about that goal, and open to a conversation about it. But if this has been decided already, then I feel comfortable that this will prevent users from setting the UTC offset. (We'll likely want to go clean up other vestigial code if this gets merged.)

AjayTripathy:
Given that it's broken right now, has been for quite some time, and even if it were working perfectly would still cause confusion when comparing data from the cur on every major cloud providers, which uses it, I think we're good to pass on it here.

PR #1090, opened 2021-09-29
Update cost-analyzer ingress

hescmi:
## What does this PR change?
Consolidates the logic for picking the Ingress API version into `cost-analyzer-ingress-template.yaml` and removes the related lines in `_helpers.tpl` which become unneeded.

Also adds if/else to the path definition portion of `cost-analyzer-ingress-template.yaml` to base the yaml used off the Ingress API version.  The two versions of the Ingress API (`v1` and `v1beta1`) use different path values and not accounting for this causes errors.

Example difference between API versions:
`serviceName` -> `service.name`
`servicePort` -> `service.port.name` OR `service.port.name` (depending on the type of value)
https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-rules

Here's a random stackoverflow example of what one would hit changing the Ingress API version but not updating the path:
https://stackoverflow.com/questions/64125048/get-error-unknown-field-servicename-in-io-k8s-api-networking-v1-ingressbacken

No.

This should allow for ingress to work for kubecost cost-analyzer with regardless of the version of Kubernetes.

N/A

## How was this PR tested?
Did a dry-run of the updated helm chart to confirm the result from `cost-analyzer/templates/cost-analyzer-ingress-template.yaml` would produce the correct result.  Only have a `1.19` cluster though.

Don't believe this should require a documentation update.  This change should require a change in values that a user has to provide.

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -149,19 +149,6 @@ Return the appropriate apiVersion for networkpolicy.
 {{- end -}}
 
 {{/*
-Return the appropriate apiVersion for Ingress.
-*/}}
-{{- define "cost-analyzer.ingress.apiVersion" -}}
-{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-{{- print "extensions/v1beta1" -}}
-{{- else if semverCompare "^1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-{{- print "networking.k8s.io/v1beta1" -}}
-{{- else if semverCompare "^1.19-0" .Capabilities.KubeVersion.GitVersion -}}
-{{- print "networking.k8s.io/v1" -}}
-{{- end -}}
-{{- end -}}
-
-{{/*
 Return the appropriate apiVersion for podsecuritypolicy.
 */}}
 {{- define "cost-analyzer.podSecurityPolicy.apiVersion" -}}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
@hescmi thank you for this PR! We will get a review here asap

hescmi:
Cool, thanks 👍

AjayTripathy (approved):

AjayTripathy:
LGTM! Thanks for this, expected in our 1.88.0 release!

viallikavoo:
Hi,
I am not sure if this is the right place to comment(please let me know if I should create a separate issue for this), but I am writing here because I suspect this change has caused the issue  I am seeing on our Kubernetes cluster.
We are running on kubernetes `1.18.10`
Upgrading to the cost-analyzer chart `1.88.0` has broken the apiVersion of our ingress object.
The helm chart renders the apiVersion as `networking.k8s.io/v1`, but since we are running on kubernetes `1.18.10` that is not a valid version for the ingress object for us.

viallikavoo:
Hi,
I am not sure if this is the right place to comment(please let me know if I should create a separate issue for this), but I am writing here because I suspect this change has caused the issue I am seeing on our Kubernetes cluster.
We are running on kubernetes 1.18.10
Upgrading to the cost-analyzer chart 1.88.0 has broken the apiVersion of our ingress object.
The helm chart renders the apiVersion as networking.k8s.io/v1, but since we are running on kubernetes 1.18.10 that is not a valid version for the ingress object for us.
Also when I list the `api-resources` on our cluster, I can see 
`
kubectl --context <redacted> api-versions | grep networking
networking.istio.io/v1alpha3
networking.istio.io/v1beta1
networking.k8s.io/v1
networking.k8s.io/v1beta1
`
I Suppose that is why the capability returns true?

AjayTripathy:
Hi @viallikavoo, sorry for the late response-- another issue would be great, but I'll get this looked into. I'm not sure what's up with the capability being listed but not existing during the deployment, can you provide the exact version and provider with `kubectl version` ? Are you using a managed provider? If so which one?

Issue #1097, opened 2021-10-07
Support custom priorityClass in cost-analyzer deployment manifest


AshutoshNirkhe:
**What problem are you trying to solve?**
Currently I can't pass my own (pre-created) priority class to the helm chat values.

**Describe the solution you'd like**
Currently helm takes care of create a new priority class if we set 'priority.enabled' to true. But we can't pass our pre-created priority class. We just need to add an 'else' condition to https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L806 to support custom priority class name/value and should be good.

**Describe alternatives you've considered**

**How would users interact with this feature?**

gz#892

AjayTripathy:
Hi @AshutoshNirkhe we're a bit backed up on supporting additional features, but we'd gladly review a PR here that meets your needs! Otherwise, we can expect to get to this in a few releases.

PR #1101, opened 2021-10-11
Add support to use already existing prority class to address #1097

AshutoshNirkhe:

- 
- 

- 
-

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -805,7 +805,11 @@ spec:
     {{- end }}
       {{- if .Values.priority }}
       {{- if .Values.priority.enabled }}
+      {{- if .Values.priority.createNew }}
       priorityClassName: {{ template "cost-analyzer.fullname" . }}-priority
+      {{- else }}
+      priorityClassName: {{ .Values.priority.name }}
+      {{- end }}
       {{- end }}
       {{- end }}
       {{- with .Values.nodeSelector }}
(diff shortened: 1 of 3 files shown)


nikovacevic (changes requested):
This version looks better to me. Thank you! Unfortunately, my original feedback included syntax that Helm doesn't support. My apologies @AshutoshNirkhe. Here's the actual syntax for those two boolean expressions:
```helm
{{- if eq (len .Values.priority.name) 0 }}
```
```helm
{{- if gt (len .Values.priority.name) 0 }}
```

The other thing is that, if `.Values.priority.enabled` is set to `true`, but `name` is not given at all, we'll get an error. It must be `name: ""`. Thus the default should likely be `""`.

Do you have concerns about any of this, @AjayTripathy?

AjayTripathy:
> The other thing is that, if .Values.priority.enabled is set to true, but name is not given at all, we'll get an error. It must be name: "". Thus the default should likely be "".

Since these are both new fields, it should be ok to just set the default to "" and .Values.priority.enabled to true, so I don't have concerns there.

AshutoshNirkhe:
> This version looks better to me. Thank you! Unfortunately, my original feedback included syntax that Helm doesn't support. My apologies @AshutoshNirkhe. Here's the actual syntax for those two boolean expressions:
> 
> ```
> {{- if eq (len .Values.priority.name) 0 }}
> ```
> 
> ```
> {{- if gt (len .Values.priority.name) 0 }}
> ```
> 
> The other thing is that, if `.Values.priority.enabled` is set to `true`, but `name` is not given at all, we'll get an error. It must be `name: ""`. Thus the default should likely be `""`.
> 
> Do you have concerns about any of this, @AjayTripathy?

@nikovacevic updated as per your comment. Could you please review/merge if all is good now ?

nikovacevic (approved):
LGTM. Thank you @AshutoshNirkhe.

PR #1106, opened 2021-10-20
emit ksm v1 node metrics only

AjayTripathy:
## What does this PR change?
Duplicate only the minimal KSM metrics for accurate node pricing.

Metrics missing from default KSM v2 installations are 
```kube_pod_labels, kube_node_labels, kube_node_status_capacity_*```

These are re-added to better support running kubecost and KSM v2 with minimal metric duplication. Running KSM v2 next to kubecost has always been supported, but users are finding it difficult to dedupe existing queries by "job=kubecost".

Testing done: Run https://github.com/kubecost/cost-model/pull/970 with these flags and note that these metrics are still emitted.


mbolt35 (approved):
LGTM

PR #1127, opened 2021-11-03
Update prometheus ingress

korjek:
## What does this PR change?
Fix the issue with Prometheus Ingress spec.

No
- 
- 

No impact

- 
- 

## How was this PR tested?
With `helm update` for running cost-analyzer helm release.

Not needed

### DIFF
--- a/cost-analyzer/charts/prometheus/templates/server-ingress.yaml
+++ b/cost-analyzer/charts/prometheus/templates/server-ingress.yaml
@@ -5,7 +5,9 @@
 {{- $serviceName := include "prometheus.server.fullname" . }}
 {{- $servicePort := .Values.server.service.servicePort -}}
 {{- $extraPaths := .Values.server.ingress.extraPaths -}}
+{{- $apiV1 := false -}}
 {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+{{- $apiV1 = true -}}
 apiVersion: networking.k8s.io/v1
 {{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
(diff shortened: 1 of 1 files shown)


korjek:
guys, could we add this to 1.88.0 release as well?
This blocks us from upgrading.

AjayTripathy:
Hi @korjek ... thank you so much for the PR. I'll get it reviewed and see if we can do a 1.88.1 patch release or just re-release 1.88.0

cost-analyzer/charts/prometheus/templates/server-ingress.yaml
```diff
@@ -34,10 +36,19 @@ spec:
 {{ if $extraPaths }}
 {{ toYaml $extraPaths | indent 10 }}
 {{- end }}
+          {{- if $apiV1 }}
+          - path: /{{ rest $url | join "/" }}
+            backend:
+              service:
+                name: {{ $serviceName }}
+                port:
```
AjayTripathy, line 44: Are we sure about this indent? Isn't port supposed to be outdented here?

AjayTripathy (changes requested):

AjayTripathy:
@korjek can you share a subset of the generated YAML file from `helm template` or the prometheus ingress yaml manifest from your testing  ? I'm not sure how this PR would work with the "port" block indented like that.

AjayTripathy (approved):

AjayTripathy:
Hi @korjek just letting you know this should be in our staging build but we can't add it retroactively to 1.88.0 

https://github.com/kubecost/docs/blob/main/staging.md for instructions-- otherwise it will be in our next major release. Thanks.

PR #1128, opened 2021-11-03
fix: default Ingress pathType

mgoodness:
Signed-off-by: Michael Goodness <michael.goodness@mlb.com>

## What does this PR change?

Set default Ingress pathType to ImplementationSpecific, which should work for any Ingress controller. Can be overridden by end-user.

No

Bug fix - v1.88.0 breaks Ingress if class: gce-internal (other classes almost certainly also affected)

## How was this PR tested?
Applied to our environment

No

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
@@ -3,6 +3,7 @@
 {{- $fullName := include "cost-analyzer.fullname" . -}}
 {{- $serviceName := include "cost-analyzer.serviceName" . -}}
 {{- $ingressPaths := .Values.ingress.paths -}}
+{{- $ingressPathType := .Values.ingress.pathType -}}
 {{- $apiV1 := false -}}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Thanks for this!

cost-analyzer/templates/cost-analyzer-ingress-template.yaml
```diff
@@ -3,6 +3,7 @@
 {{- $fullName := include "cost-analyzer.fullname" . -}}
 {{- $serviceName := include "cost-analyzer.serviceName" . -}}
 {{- $ingressPaths := .Values.ingress.paths -}}
+{{- $ingressPathType := .Values.ingress.pathType -}}
```
AjayTripathy, line 6: Can we get a default here if the user does not supply anything? ImplementationSpecific should be fine.

mgoodness: If a user doesn't override from their own values file, the chart will use the value from the `values.yaml` file that ships with the chart. So the default is already `ImplementationSpecific`.

AjayTripathy (approved):

PR #1159, opened 2021-12-07
Mount GCP Service Key Secret Correctly

### LINKED ISSUE #1135: gcpSecretName not propagated into the application correctly
**Describe the bug**
When specifying the secret name for GCP integration in the helm chart via `gcpSecretName`, the key doesn't get properly propagated into the application, even though all the prerequisites seem to exist in the container.

```
/ $ ls /models/key.json
/models/key.json
/ $ echo $GOOGLE_APPLICATION_CREDENTIALS
/models/key.json
```

The error from `curl ../model/etl/status | jq`

```
....
    "cloud": {
      "redacted": {
        "cloudConnectionStatus": "Failed Connection",
        "cloudAssets": {
          "coverage": {
            "start": "2021-08-03T00:00:00Z",
            "end": "2021-11-15T00:00:00Z"
          },
          "lastRun": "2021-11-08T11:55:42.174587027Z",
          "nextRun": "2021-11-08T14:55:42.174169884Z",
          "progress": 1,
          "refreshRate": "3h0m0s",
          "resolution": "24h0m0s",
          "startTime": "2021-11-08T11:55:42.172201827Z"
....
```

The key and other settings are, however, correct - when I copy the contents of the key mounted to `/models/key.json` in cost-model container and paste it in the form at `/keyinstructions.html`, everything works out well and the cloud connection reports success once rebuilt.

`curl /model/etl/asset/cloud/rebuild?commit=true`
`curl ../model/etl/status | jq`

```
...
    "cloud": {
      "redacted": {
        "cloudConnectionStatus": "Connection Successful",
        "cloudAssets": {
          "coverage": {
            "start": "2021-10-24T00:00:00Z",
            "end": "2021-11-08T12:06:10Z"
          },
          "lastRun": "2021-11-08T12:06:11.331263027Z",
          "nextRun": "2021-11-08T12:06:10.643527397Z",
          "progress": 0,
          "refreshRate": "3h0m0s",
          "resolution": "24h0m0s",
          "startTime": "2021-11-08T12:06:10.643527282Z"

...
```

**EDIT: It seems that, by default, kubecost is reaching for a legacy key path (`/var/configs/key.json`), however I haven't been successful finding this reference in the code base of cost-model. If I copy `/models/key.json` to `/var/configs/key.json` and run rebuild, everything seems to work correctly.**

```
cost-model E1108 13:42:04.881686       1 log.go:17] [Error] ETL: CloudAsset[heu-infra-global]: Build[SjwMP]: QueryCloudAssets error: bigquery: constructing client: cannot read credentials file: open /var/configs/key.json: no such file or directory
```

**To Reproduce**
Detailed in the description

**Expected behavior**
The key should be propagated into the application from the secret automatically

```
$ helm ls -n kubecost
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
kubecost        kubecost        1               2021-11-08 12:52:28.392449 +0100 CET    deployed        cost-analyzer-1.88.0    1.88.0
```

gz#1002

mbolt35:
## What does this PR change?
* The location in which the GCP service key is mounted: `/var/secrets/service-key.json` 
* Do not change key location in configuration, GCP auth handling code already exists (it's just looking in the incorrect place). 

* Allows use of `gcpSecretName` in the helm chart. 

- Fix: https://github.com/kubecost/cost-analyzer-helm-chart/issues/1135

## How was this PR tested?
- Updated our local GCP cluster with chart updates

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -93,7 +93,7 @@ spec:
             secretName: {{ .Values.kubecostProductConfigs.gcpSecretName }}
             items:
             - key: compute-viewer-kubecost-key.json
-              path: key.json
+              path: service-key.json
         {{- end }}
         {{- end -}}
         {{- if .Values.kubecostProductConfigs.serviceKeySecretName }}
@@ -285,13 +285,13 @@ spec:
             {{- end }}
             {{- if .Values.kubecostProductConfigs.gcpSecretName }}
             - name: gcp-key-secret
-              mountPath: /models
+              mountPath: /var/secrets
             {{- end }}
-            {{- if or .Values.kubecostProductConfigs.azureStorageSecretName .Values.kubecostProductConfigs.azureStorageCreateSecret}}
+            {{- if or .Values.kubecostProductConfigs.azureStorageSecretName .Values.kubecostProductConfigs.azureStorageCreateSecret }}
             - name: azure-storage-config
               mountPath: /var/azure-storage-config
             {{- end }}
-            {{- if .Values.kubecostProductConfigs.cloudIntegrationSecret}}
+            {{- if .Values.kubecostProductConfigs.cloudIntegrationSecret }}
             - name: cloud-integration
               mountPath: /var/cloud-integration
             {{- end }}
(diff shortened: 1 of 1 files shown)


Sean-Holcomb (approved):
This should correctly load the secret into `/var/configs/service-key.json` where it can be accessed by the GCP provider and moved to `/var/configs/key.json`. One thing I would like to point out is that the contents and the amount of the configuration contained by `/var/configs/service-key.json` and what is done with it varies by provider. AWS and Azure take the values and move them into the provider object to be used on instantiation, while GCP moves it into a file which the GCP SDK accesses during configuration. Additionally Azures secret contains the entire azure configuration, while GCP and AWS contain only the credentials and not the location of the billing data (BigQuery or Athena)

mbolt35:
> This should correctly load the secret into `/var/configs/service-key.json` where it can be accessed by the GCP provider and moved to `/var/configs/key.json`. One thing I would like to point out is that the contents and the amount of the configuration contained by `/var/configs/service-key.json` and what is done with it varies by provider. AWS and Azure take the values and move them into the provider object to be used on instantiation, while GCP moves it into a file which the GCP SDK accesses during configuration. Additionally Azures secret contains the entire azure configuration, while GCP and AWS contain only the credentials and not the location of the billing data (BigQuery or Athena)

I think this is probably ok, I don't think we'll have parity across service provider auth formatting. We _should_ at least use similar code pathways to process these secrets, which I think we do currently. As long as the providers can receive the auth secret in a familiar format, I'm ok with the current approach.

PR #1193, opened 2022-01-05
release-1.89.1

AjayTripathy:
## What does this PR change?
Patch release for panic fixes.

### DIFF
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -1,10 +1,10 @@
 
 apiVersion: v2
-appVersion: "1.89.0"
+appVersion: "1.89.1"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: 1.89.0
+version: 1.89.1
 annotations:
   "artifacthub.io/links": |
     - name: Homepage


pierluigilenoci:
Is this not considered a Breaking change?

AjayTripathy:
Hi @pierluigilenoci I'm not sure I follow. v1.89.0 had a few panics on certain environments which are addressed in 1.89.1 , so if you were experiencing those, you should attempt an upgrade to 1.89.1

pierluigilenoci:
@AjayTripathy when I do the updates I always check the differences between the various chart versions in order to understand if there are any configuration changes. When I did between 1.89.0 and 1.89.1 yesterday I found changes different from the current ones. I have seen that you have done release 1.89.1 three times, maybe it was in one of the previous releases. 

Everything is fine, sorry for the inconvenience.

pierluigilenoci:
@AjayTripathy found it. I was referring to this https://github.com/kubecost/cost-analyzer-helm-chart/pull/1132 
This is tagged with v1.89.1

kbrwn:
@pierluigilenoci The option to set a className for ingress has not be removed. It has been turned off by default via a comment as not all ingress providers support `ingress.className`. You can still use this configuration option in your values.yaml.

pierluigilenoci:
Ok, you make backward compatibility only when it suits you. Good to know.
I still think this change is "breaking change".

dwbrown2:
@pierluigilenoci I would ask you to please be respectful to those in our community 

This change was not in this patch release. It was documented [here](https://github.com/kubecost/cost-analyzer-helm-chart/releases/tag/v1.89.0).

We had no intention for this to be a breaking change to anyone. It was a community fix in this PR https://github.com/kubecost/cost-analyzer-helm-chart/pull/1132

Any more you can share on the issue it caused for you? We absolutely do our best to maintains backwards compatibility.

pierluigilenoci:
@dwbrown2 I have absolutely no disrespect to the community. 

My criticism is about the lack of communication regarding how this PR has been integrated into the master.

It is documented in release 1.89.0 (without specifying that this is a breaking change) but the commit in question https://github.com/kubecost/cost-analyzer-helm-chart/pull/1132/commits/d10e2783f8b2e06bf1cf758903deb63113613d4d is tagged `1.89.1`.

Screenshot 2022-01-13 at 12 09 39

So in addition to the lack of communication there is also a bit of mess in the repo with tags.

pierluigilenoci:
Maybe it's not clear why it's a breaking change.

By removing that configuration from the ingress, if the Nginx controller does not have a default "class" it removes the Kubecost ingress from the configuration and Kubecost becomes unreachable.

kbrwn:
@pierluigilenoci We have to set sensible configuration defaults for a wide variety of clusters and ingress controllers. Not all users of kubecost use nginx-ingress. Setting `className` by default caused several tickets and slack threads where we had to assist users with the very problem you describe here: Kubecost becomes unreachable due to an invalid className parameter in the ingress resource. 

As I mentioned previously you still have the option to use this configuration in the values.yaml file if you are using an ingress controller that supports className.

pierluigilenoci:
@kbrwn I'm not questioning the technical choice (which I might as well share).

Obviously, I do not agree with the choice of "preferring" to make unreachable the ingresses of users who use Nginx than those who do not. They too had the option of specifying a className. 😉 

The problem is that if you put a default parameter and then remove it all those who used it then find themselves without that parameter configured. 

It would have been enough to communicate this in a more evident way in the Release Note.

Issue #1150, opened 2021-11-27
YAML Key-duplication


lkhomenk:
**Describe the bug**
YAML Key duplication in https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml#L12 conflicts with `app` label from commonLabels.
Helm itself can handle key-duplication but latest versions on [flux with kustomize v4](https://github.com/fluxcd/flux2/issues/1522). This is the only duplicated key in whole chart
`app: cost-analyzer` label is not needed in `cost-analyzer-network-costs-service` as PodMonitor uses another value
```
map[string]interface {}(nil): yaml: unmarshal errors:
mapping key "app" already defined
```

**To Reproduce**
1. Run yamllint against helm manifests of templates/cost-analyzer-network-costs-service-template.yaml
2. grep key-duplication

Or deploy chart via flux 1.23

gz#1084

kbrwn:
@lkhomenk Thanks for the report. Helm is the supported tool to deploy kubecost. We hope to maintain compatibility with other tools such as flux/kustomize etc but this may not always be the case as new versions of these other tools come out and have breaking changes. As helm can handle key-duplications without error this is unlikely to be a priority to fix right away. We are happy to review any PRs if you are up for the challenge.

PR #1198, opened 2022-01-07
feat: dedicated networkcosts labels

### LINKED ISSUE #1176: Network Costs Service 'app' label is overwritten by commonLabels
https://github.com/kubecost/cost-analyzer-helm-chart/blob/cb836cf5f765474509b65c94eeb182e6039b8b94/cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml#L12-13

The ordering of these two lines means that the `app` label is overwritten by the `commonLabels` from the helper template. This means the Service for network costs has a different `app` label to the pods in the DS.

```
## Service:
...
metadata:
  labels:
    app: cost-analyzer
...

## Pod
...
metadata:
  labels:
    app: kubecost-network-costs
...
```

Thus, when using a Prometheus-operator `ServiceMonitor`, the `labelSelector` doesn't work as it's different between the pods and the service.

gz#1157

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-56) by [Unito](https://www.unito.io)

lkhomenk:
## What does this PR change?

- Adds aligned labels for all stuff related to network costs - this stack of resources is to different and it doesn't need commonLabels
- Adds global.additionalLabels to commonLabels - kubecost relies on pod labels but couldn't labels all workloads itself
- Resolves [this](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1176 ) and [this](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1150) issues

No
- 
- 

https://github.com/kubecost/cost-analyzer-helm-chart/issues/1176
https://github.com/kubecost/cost-analyzer-helm-chart/issues/1150
- 
- 

## How was this PR tested?
```helm template --debug kubecost . --values values.yaml > kubecost.yaml && yamllint kubecost.yaml``` to exclude key-duplication + validate that all labels are at their places

no new parameters/variables were added

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-policy-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-policy-template.yaml
@@ -10,7 +10,7 @@ metadata:
 {{- end }}
   namespace: {{ .Release.Namespace }}
   labels:
-    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{ template "cost-analyzer.networkCostsLabels" merge (dict "extraLabels" .Values.networkCosts.additionalLabels) .  }}
 {{- if .Values.networkPolicy.costAnalyzer.additionalLabels }}
 {{ toYaml .Values.networkPolicy.costAnalyzer.additionalLabels | indent 4 }}
 {{- end }}
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -99,11 +99,22 @@ Create the name of the service account
 Create the common labels.
 */}}
 {{- define "cost-analyzer.commonLabels" -}}
+{{- if .Values.global.additionalLabels }}
+{{- toYaml .Values.global.additionalLabels -}}
+{{- end }}
 app.kubernetes.io/name: {{ include "cost-analyzer.name" . }}
 helm.sh/chart: {{ include "cost-analyzer.chart" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app: cost-analyzer
+{{- end -}}
+
+{{- define "cost-analyzer.networkCostsLabels" -}}
+{{- $labelChart := include "cost-analyzer.chart" $ -}}
+{{- $labelApp := include "cost-analyzer.networkCostsName" $ -}}
+{{- $labelName := include "cost-analyzer.name" $ -}}
+{{- $labels := dict "app.kubernetes.io/name" $labelName "app" $labelApp "helm.sh/chart" $labelChart "app.kubernetes.io/instance" .Release.Name  "app.kubernetes.io/managed-by" .Release.Service -}}
+{{ merge .extraLabels $labels | toYaml | indent 4 }}
 {{- end -}}
 
 {{/*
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by mbolt35** [comment]
Thanks for the PR! I haven't looked through all of this yet, but I do think your change to `commonLabels` will cause problems for cost-analyzer-deployment (which I'm pretty sure uses `commonLabels` and `global.additionalLabels` ). I'm not certain though - will need to pull it down and give it a shot. 

For reference: It looks like the pod spec for the deployment uses the `global.additionalLabels`:
https://github.com/parimatch-tech/cost-analyzer-helm-chart/blob/b364a57919e8ea7103440e87c8507f49a285f4ba/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L25-L31

And the deployment labels simply use the `commonLabels`:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/72c0a92215ddaf53365b7b278cc7b93542c3eb37/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L6-L7

It's not clear to me whether we want the additional labels to be included with common labels (although I do agree with your opinion of making it simpler). In a meeting currently, but I'll take a closer look when it's over. Thanks again for the help here!

File: cost-analyzer/templates/cost-analyzer-network-policy-template.yaml [hunk 1]
```diff
@@ -10,7 +10,7 @@ metadata:
 {{- end }}
   namespace: {{ .Release.Namespace }}
   labels:
-    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{ template "cost-analyzer.networkCostsLabels" merge (dict "extraLabels" .Values.networkCosts.additionalLabels) .  }}
```
mbolt35, line 13: This seems unrelated to network-costs labels.

mbolt35:
@lkhomenk Sorry for the delay here, I've taken a deeper look at this PR, and I had some issues in my tests:

I added the following `global.additionalLabels` in `values.yaml`:
```
  additionalLabels:
    foo: bar
    test1: true
```
Running `helm template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml  --set networkCosts.enabled=true > out.yaml`

If I look in `out.yaml` under `# Source: cost-analyzer/templates/cost-analyzer-deployment-template.yaml`
I expected to see 
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kubecost-cost-analyzer
  labels:
    foo: bar 
    test1: true 
    helm.sh/chart: cost-analyzer-1.89.1
    app: cost-analyzer
```
but instead the result I got was:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kubecost-cost-analyzer
  labels:
    
    
    helm.sh/chart: cost-analyzer-1.89.1
    app: cost-analyzer
```

The `matchLabels` also seem to be missing the additionalLabels
```
    matchLabels:

        app: cost-analyzer
```

It looks like `.Values.global` isn't in the scope `_helpers.tpl`?

Similar to the above situation, `networkCosts.additionalLabels` also seem to be ignored.

kbrwn:
@lkhomenk Please review the provided feedback on this PR.

Adam-Stack-PM:

PR #1208, opened 2022-01-21
Simplify Open Source Interoperability

AjayTripathy:
## What does this PR change?
Use a single flag to swap in the open source images.

https://github.com/kubecost/cost-model/pull/1028

Simplify deployment of open-source cost model from helm chart

## How was this PR tested?
Deployed against open source cost-model

No-- will do when this comes out of beta (better UI/UX on frontend still required)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -236,7 +236,9 @@ spec:
 {{ end }}
       containers:
         {{- if .Values.kubecostModel }}
-        {{- if .Values.kubecostModel.fullImageName }}
+        {{- if .Values.kubecostModel.openSourceOnly }}
```
dwbrown2, line 239: Would it be a simpler user experience to have a global flag equivalent to "openSourceOnly"? That way you would abstract away from each individual image being changed?

Also, we should think about naming here... `kubecost-cost-model` is that open source image? That seems confusing if cost-model is that plus other bits..

AjayTripathy: Yeah, the names have been backwards since we started naming the community edition just "cost-model". Let me investigate what it would take to change the names to reflect what they are.

michaelmdresser: I don't think we've had a canonical name for the full-OSS image because it isn't a heavily documented install path. Once we start going down this path, I expect the chosen name will stick.

I'll suggest `gcr.io/kubecost1/cost-model-oss` as a candidate name; it conveys reasonably well that its "equivalent" to the `cost-model` image but without all the excellent Kubecost special sauce.

PR #1225, opened 2022-01-31
migrate healthz endpoints outside of auth redirect to better support ingress healthchecks

AjayTripathy:
## What does this PR change?
Don't auth redirect the healthz endpoints

No

Because the healthz endpoint was responding with a redirect to the SAML provider, certain External Load Balancers and V2 Ingresses were failing. 

## How was this PR tested?
curl the /healthz endpoints and note that they no longer issue redirects.

No, this should work seamlessly.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -88,6 +88,18 @@ data:
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
             error_page 401 = /login;
         }
+        location /healthz {
+            return 200;
+            add_header Content-Type text/plain;
+        }
+        location /api/healthz {
+            return 200;
+            add_header Content-Type text/plain;
+        }
+        location /model/healthz {
+            return 200;
+            add_header Content-Type text/plain;
+        }
 {{- else }}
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -88,6 +88,18 @@ data:
...
+            add_header Content-Type text/plain;
+        }
+        location /api/healthz {
+            return 200;
+            add_header Content-Type text/plain;
+        }
+        location /model/healthz {
+            return 200;
+            add_header Content-Type text/plain;
+        }
```
michaelmdresser, line 102: I don't quite understand the implications of this change.

- Is NGINX serving requests considered suitable for "cost-analyzer-frontend is healthy"?
- Why do we need to do `/api/healthz` and `/model/healthz` on top of `/healthz`? AFAICT K8s only hits `/healthz` [for the frontend readinessProbe](https://github.com/kubecost/cost-analyzer-helm-chart/blob/ef9b541880070ccb2b33171abe3411ed916bb404/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L694)
- I assume this doesn't break the cost-model readiness probe because its a different container and K8s hits the container directly for healthz?

PR #1233, opened 2022-02-03
Permissions for request sizing application

michaelmdresser:
## What does this PR change?

Adds permissions to the cost-analyzer ClusterRoleBinding for application of request sizing recommendations.

N/A

Kubecost is given `update` and `get` verbs on StatefulSets, Deployments, DaemonSets, and ReplicaSets to support a new automation feature.

N/A

## How was this PR tested?

See PRs implementing core functionality. This chart edit was used to deploy those PRs.

N/A

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -206,6 +206,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: CLUSTER_ID
+          value: {{ .Values.prometheus.server.global.external_labels.cluster_id }}
         - name: TURNDOWN_NAMESPACE
           value: {{ .Release.Namespace }}
         - name: TURNDOWN_DEPLOYMENT


dwbrown2:
I actually think we want to be crazy thoughtful here. We have historically put all write edit/permissions in a separate pod/deployment. Specifically what is the turndown pod today. 

What do you think about expanding the scope of the turndown pod and calling that the broader Kubecost agent/controller? That feels like really clean isolation to me.

michaelmdresser:
> What do you think about expanding the scope of the turndown pod and calling that the broader Kubecost agent/controller?

I'm pretty convinced this is the correct play in the long term. If you think the additional permissions in this PR will prevent regular installs then I'm totally okay with pulling the plug on this PR and shifting the corresponding code in other PRs to a proper Kubecost "continuous automation" agent (for lack of official terminology).

dwbrown2:
I think it's worth a debate... In short, I feel like this would be a good change to consider with Kubecost 2.0 but that many users really appreciate Kubecost being read-only by default. 

Again, open to discussing this!

nealormsbee:
It seems like the two of you are totally agreed on a distinct / isolated workload for write permissions being clean architecture, as well as something our users will probably appreciate. That all seems right to me as well. Putting myself in the shoes of a user not ready to jump into automation, I'd feel good about restricting perms to read-only. Separation also limits our exposure to exploit risk, i.e. getting a pod that's only _intended_ to read, to write.

So I'm reading this as mostly as a question of where to invest time/focus in the near-term. We're taking the next ~2 months to focus on automation; if, at the end of that two months, _all_ we had accomplished was a really well-defined read/write agent separation, laying the groundwork for future development, I would be pleased as punch. So I'm all for doing that work now 👍

michaelmdresser:
I like that take, Neal.

One of the things that's tricky is having our own in-product permissions scheme. Putting write operations in the UI, which could be accessible to any number of people and isn't subject to user-specific cluster permissions (K8s RBAC), should be carefully considered and I hadn't thought about it until now. @dwbrown2 what do you think the story looks like here? I'm happy to take this to a different forum if you'd like.

Just to keep throwing ideas out there, what if we had a single Helm value that determined if Kubecost (cost-model or any associated pods, like turndown) has write permission to the cluster and then looped all automation features under that flag? I can see plenty of pitfalls here, but I'm interested if it will spark some conversation.

dwbrown2:
Love the discussion! I'd be interested if there are other perspectives. cc @mbolt35 @kaelanspatel @AjayTripathy

kaelanspatel:
I haven't had too much of a horse in this race so far, but I'm very supportive wrt the agent pod and the single "lynchpin" helm value that governs write permissions.

I think getting ahead of the curve here will be really conducive to reducing the bloat of our configs and their associated logic.

michaelmdresser:
> just want to get a sense if you're majorly blocked.

I can do plenty of internal testing for approaches to automation that isn't dependent on having these permission right now. The distance from the release means this can wait.

For some extra commentary on the separate agent approach, see Bolt's points [in the review of the API PR](https://github.com/kubecost/kubecost-cost-model/pull/633#issuecomment-1029412940).

dwbrown2:
So far all our write permissions are managed by this single controller flag:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/c795ba9f582def39381876d229fd531a4626eba9/cost-analyzer/values.yaml#L598

Do you like that paradigm? It feels to me like it's worked pretty well, but let me know if you see immediate opportunities for improvement.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -64,6 +64,8 @@ rules:
     verbs:
       - list
       - watch
+      - update
+      - get
```
mbolt35, line 68: I saw above in this file, there's a "deployments" under `apiGroups: ''` with `get, list, watch` only permissions. What's the difference between that and the `apiGroups: extensions` ?

michaelmdresser: Deployments had API versions: `extensions/v1beta1`, `apps/v1beta1`, and `apps/v1beta2` [until K8s 1.16](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/). Those patch notes include similar info about DaemonSets, StatefulSets, etc. So I think if we need to access deployments (etc.) in K8s versions pre-1.16 we would need the `apiGroups: extensions` permissions.

michaelmdresser:
I'm closing this PR because cluster-controller, where the functionality these permissions were for [now lives](https://github.com/kubecost/cluster-controller/pull/2), already has [these permissions](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/kubecost-cluster-controller-template.yaml#L92-L106).

I'm going to open a different PR for a small env var change that cluster-controller needs for that PR.

michaelmdresser:
Thanks for all the great comments! This was a good discussion and I think this thread is worth referring back to in the future when we talk about write-level permission in the cluster :smile:

Issue #1224, opened 2022-01-30
"Logout" button w/redirect


kirbsauce:
**What problem are you trying to solve?**
The ability for our users to logout/sign out of the UI and be redirected to a configurable url.

**Describe the solution you'd like**
A logout button which would clear the session and redirect the user to a configurable URL

**Describe alternatives you've considered**
none

**How would users interact with this feature?**
User would click the logout button and be redirected to a configurable url.

related to https://github.com/kubecost/cost-analyzer-helm-chart/issues/1223

kaelanspatel:
This has been implemented with the linked PR.

Issue #1223, opened 2022-01-30
Configurable Session Timeout


kirbsauce:
**What problem are you trying to solve?**
Our policy dictates a 15 minute inactivity session timeout.  

**Describe the solution you'd like**
I would like the ability to define a timeout length and a redirect URL.

**Describe alternatives you've considered**
none

**How would users interact with this feature?**
For security purposes, we would like to define this at deployment time via the values.yaml.

kaelanspatel:
This has been implemented with the linked PR.

PR #1246, opened 2022-02-10
Configurable SAML token expiration and SAML logout functionality

kaelanspatel:
## What does this PR change?
Adds configurable SAML cookie timeout and a logout button/endpoint that invalidates the cookie and redirects to a configurable URL.
image

- https://github.com/kubecost/cost-analyzer-frontend/pull/504
- https://github.com/kubecost/kubecost-cost-model/pull/654

Adds configurable SAML session timeout and logout redirect URL (with associated logout button on FE).

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1223
- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1224

## How was this PR tested?
Manually with Auth0.

No.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -153,6 +153,8 @@ saml:
   #metadataSecretName: "kubecost-authzero-metadata" # One of metadataSecretName or idpMetadataURL must be set. defaults to metadataURL if set
   idpMetadataURL: "https://dev-elu2z98r.auth0.com/samlp/metadata/c6nY4M37rBP0qSO1IYIqBPPyIPxLS8v2"
   appRootURL: "http://localhost:9090" # sample URL
+  authTimeout: 1440 # number of minutes the JWT will be valid
+  redirectURL: "https://dev-elu2z98r.auth0.com/v2/logout" # callback URL redirected to after logout
   # audienceURI: "http://localhost:9090" # by convention, the same as the appRootURL, but any string uniquely identifying kubecost to your samp IDP. Optional if you follow the convention
   # nameIDFormat: "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified" If your SAML provider requires a specific nameid format
   # isGLUUProvider: false # An additional URL parameter must be appended for GLUU providers
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -614,6 +614,14 @@ spec:
             - name: NAME_ID_FORMAT
               value: {{ .Values.saml.nameIDFormat }}
             {{- end}}
+            {{- if .Values.saml.authTimeout }}
+            - name: AUTH_TOKEN_TIMEOUT
+              value: {{ (quote .Values.saml.authTimeout) }}
+            {{- end}}
+            {{- if .Values.saml.redirectURL }}
+            - name: LOGOUT_REDIRECT_URL
+              value: {{ .Values.saml.redirectURL }}
+            {{- end}}
             {{- if .Values.saml.rbac.enabled }}
             - name: SAML_RBAC_ENABLED
               value: "true"
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -187,6 +187,18 @@ data:
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
 
+        location /logout {
+            proxy_connect_timeout       180;
+            proxy_send_timeout          180;
+            proxy_read_timeout          180;
+            proxy_pass http://model/logout;
+            proxy_redirect off;
+            proxy_http_version 1.1;
+            proxy_set_header Connection "";
+            proxy_set_header  X-Real-IP  $remote_addr;
+            proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
+        }
+
     {{- if .Values.global.grafana.proxy }}
         location /grafana/ {
         {{- if .Values.saml.enabled }}


cost-analyzer/values.yaml
```diff
@@ -153,6 +153,8 @@ saml:
   #metadataSecretName: "kubecost-authzero-metadata" # One of metadataSecretName or idpMetadataURL must be set. defaults to metadataURL if set
   idpMetadataURL: "https://dev-elu2z98r.auth0.com/samlp/metadata/c6nY4M37rBP0qSO1IYIqBPPyIPxLS8v2"
   appRootURL: "http://localhost:9090" # sample URL
+  authTimeout: 1440 # number of minutes the JWT will be valid
```
AjayTripathy, line 156: How is this being passed into the app? Does there need to be an environment variable or something?

cost-analyzer/values.yaml
```diff
@@ -153,6 +153,8 @@ saml:
...
+  redirectURL: "https://dev-elu2z98r.auth0.com/v2/logout" # callback URL redirected to after logout
```
AjayTripathy, line 157: Same as above.

cost-analyzer/values.yaml
kaelanspatel: There's an [existing configmap](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-saml-config-map-template.yaml) for SAML which is [already being mounted](https://github.com/kubecost/cost-analyzer-helm-chart/blob/4a6b07dfb1679983b6dc655e8aac2fb74b735cec/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L168-L172). 

Actually, I see now it doesn't actually get mounted unless rbac is enabled... is there a reason for this? Because this configmap contains things not related to rbac as well (not just this PR's changes). I think we should probably change this to always mount if Values.saml.enabled.

AjayTripathy: I'd avoid making such a foundational change and just  also pass/moun` these as env variables to not blow up the scope here. 

Once you do that, I'd also double check that authTimeout gets parsed correctly as an int when coming through an env variable since it won't be json serialized.

kaelanspatel: Is there a particular reason why we view changing this configmap mount condition as foundational? AFAICT this configmap is always created if SAML is enabled.

Otherwise I feel strongly like creating two more env variables to pass values that are already in a configmap redundant. Increasing our number of env variables to avoid mounting a configmap that we both already try to unmarshal if SAML is enabled AND mount anyway if RBAC is enabled seems incorrect.

AjayTripathy: It's beyond the scope of the PR to make that change, it's a significant change in functionality that may regress users since we'd be introducing a configmap and removing env variables. I would make the minimal change to unblock the feature then file a follow up issue if you feel strongly.

michaelmdresser: I'm a bit confused by the current state of the world, here. I'm going to try to summarize what I see here and ask some questions.

From doing a little bit of spelunking, it looks to me like we have [a few env vars related to SAML](https://github.com/kubecost/cost-analyzer-helm-chart/blob/4a6b07dfb1679983b6dc655e8aac2fb74b735cec/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L597-L622):
- `SAML_ENABLED` (makes sense, this turns the functionality on)

And then we map some values under the `saml:` key:
- `IDP_URL`
- `SP_HOST`
- `GLUU_SAML_PROVIDER`
- `NAME_ID_FORMAT`
- and `SAML_RBAC_ENABLED` 

But we _also_ turn ALL of the values under `saml:` into the ConfigMap Kaelan is talking about. That's done [like this](https://github.com/kubecost/cost-analyzer-helm-chart/blob/4a6b07dfb1679983b6dc655e8aac2fb74b735cec/cost-analyzer/templates/cost-analyzer-saml-config-map-template.yaml#L11): 
```yaml
saml.json: '{{ toJson .Values.saml }}'
```

That ConfigMap is only mounted if `.Values.saml.rbac.enabled=true`, as Kaelan has noted.

@AjayTripathy Here are some questions that I hope will bring some clarity:
1. Why do we serialize everything under `saml:` into the ConfigMap?
2. Why do we pass only some `saml:` values as env vars?
3. Why do we only mount the ConfigMap if `.Values.saml.rbac.enabled=true`?
4. Are there values under `saml:` that should always be in the ConfigMap (instead of env var)?
5. Are there values under `saml:` that should always be in an env var (instead of the ConfigMap)?




I think there's definitely a ticket worth opening here to standardize how SAML config gets into the application. I think our discussion here can inform some documentation and, eventually, that ticket.

---

Here's a more direct contribution to this conversation, based on my current understanding:

A brief tour into the code suggests that mounting the ConfigMap but keeping `.Values.saml.rbac.enabled=false` shouldn't have adverse effects, see this snippet for example (other code paths are similar):
https://github.com/kubecost/kubecost-cost-model/blob/2f61e609cfae098886be61208831ed3439b66187/pkg/cmd/costmodel/auth.go#L204

I don't think Kaelan is suggesting that we remove any environment variables.

I think that he has read the code, found that much of SAML config is being pulled from the ConfigMap, and decided to follow what he found and implement the new configuration in the ConfigMap instead of environment variables.

I don't think he's suggesting we introduce a new ConfigMap, either; he'd like to always _mount_ the already-existing ConfigMap to have access to these new config values. I think this is the crux: I don't quite understand why this isn't allowed. I hope answers to the above questions will get us all on the same page.

kaelanspatel: @michaelmdresser @AjayTripathy Apologies for the long discussion. After looking at the RBAC code, I think there is a possibility that unintended effects could occur if the configmap is mounted while RBAC is disabled. While I would like to address these in the interests of clarity, I understand that changes may impact RBAC functionality and so will leave that out of this PR.

To that end, I have implemented and tested the proposed env variables. If I could get another look, that would be great.

AjayTripathy (approved):

PR #1249, opened 2022-02-11
Add command and args fields for cost-model and network-costs pods

rdavaze-talend:
## What does this PR change?

- Add command and args fields for cost-model and network-costs pods

No

Users will be able to override default command and args of cost-model and network-costs pods if they want to.

N/A

## How was this PR tested?

- Manually using helm template command
- Upgrade of existing helm chart on a k8s cluster

Not sure what documentation can be updated to reflect this change.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -250,10 +250,6 @@ spec:
         - image: gcr.io/kubecost1/cost-model:prod-{{ $.Chart.AppVersion }}
         {{ end }}
           name: cost-model
-        {{- if .Values.kubecostModel.command }}
-          command:
-          {{- toYaml .Values.kubecostModel.command | nindent 12 }}
-        {{- end }}
         {{- if .Values.kubecostModel.extraArgs }}
           args:
           {{- toYaml .Values.kubecostModel.extraArgs | nindent 12 }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -250,6 +250,14 @@ spec:
         - image: gcr.io/kubecost1/cost-model:prod-{{ $.Chart.AppVersion }}
         {{ end }}
           name: cost-model
+        {{- if .Values.kubecostModel.command }}
+          command:
+          {{- toYaml .Values.kubecostModel.command | nindent 12 }}
+        {{- end }}
+        {{- if .Values.kubecostModel.args }}
+          args:
+          {{- toYaml .Values.kubecostModel.args | nindent 12 }}
+        {{- end }}
         {{- if .Values.kubecostModel.imagePullPolicy }}
           imagePullPolicy: {{ .Values.kubecostModel.imagePullPolicy }}
         {{- else }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -254,9 +254,9 @@ spec:
           command:
           {{- toYaml .Values.kubecostModel.command | nindent 12 }}
         {{- end }}
-        {{- if .Values.kubecostModel.args }}
+        {{- if .Values.kubecostModel.extraArgs }}
           args:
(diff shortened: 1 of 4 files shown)


AjayTripathy:
Hi Romain, thank you so much for this PR. Assigning for review!

mbolt35:
@rdavaze-talend Can you give me an example of what commands you would want to override or commands you'd be including?

rdavaze-talend:
@mbolt35 It will be used to reduce the log verbosity of the cost-model and network pods.
This will be done by setting `-v=1` in the `args` property (already tested by patching the deployment manually).

I have also added the command property just in case someone needs it in the future. For my use case I only need to set the `args` field.

mbolt35:
> @mbolt35 It will be used to reduce the log verbosity of the cost-model and network pods.
> 
> This will be done by setting `-v=1` in the `args` property (already tested by patching the deployment manually).
> 
> 
> 
> I have also added the command property just in case someone needs it in the future. For my use case I only need to set the `args` field.

Oh, interesting! I remember a while back someone tried setting the -v flag, but I thought our network-costs and cost-model code set that flag directly (overwriting the command line arg). Were you able to confirm this actually reduced the verbosity of the logs? Otherwise, this may require some changes to the respective code bases.

rdavaze-talend:
> Oh, interesting! I remember a while back someone tried setting the -v flag, but I thought our network-costs and cost-model code set that flag directly (overwriting the command line arg). Were you able to confirm this actually reduced the verbosity of the logs? Otherwise, this may require some changes to the respective code bases.

Yeah I can confirm that it does indeed reduce the number of logs. Setting the flag to **2** does not seem to change anything but when I set it to **1**, only **ERROR** logs are shown. This reduce the number of logs tremendously, especially for cost-model which prints a lot at the **INFO** level.

For what I understand the **-v** flag is handled by default by the klog library and is taken into account automatically by cost-model and network-costs.

mbolt35:
@rdavaze-talend Ok, so to provide some back story here, the cost-model actually expects a single command line argument denoting which "mode" to run in. The two valid arguments are `cost-model` and `agent`. The helm chart doesn't supply the argument for the cost-model container because it's also the default mode (if the `cost-model` arg is missing, we just prepend it in the arg list):
https://github.com/kubecost/cost-model/blob/76c9adc9030b830dec700d5d14b12f45ea0b492d/pkg/cmd/commands.go#L50-L61

We are also considering adding more to command line in the future which we may not want to expose via values. So, I think it's best to name `args` value in the `values.yaml` file to `extraArgs` or `additionalArgs` (I'll leave that to you, I'm ok with either). That way it's clear that the arguments are appended. This is similar to the way the thanos chart does it: 
https://github.com/kubecost/cost-analyzer-helm-chart/blob/b95268f11ea491838b4ea2f214690e7ca431c3d2/cost-analyzer/charts/thanos/templates/query-deployment.yaml#L97-L99

Network costs should get the same treatment (even though it doesn't currently accept any network-costs specific command line args). 

Lastly, there's another deployment that needs the same addition: 
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml

Specifically, there's already an `arg` being passed here: 
https://github.com/kubecost/cost-analyzer-helm-chart/blob/b95268f11ea491838b4ea2f214690e7ca431c3d2/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml#L173-L174

You can add the `extraArgs` value in this block: 
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L225-L252

Thanks again for the PR, are you ok with making these changes?

rdavaze-talend:
@mbolt35 I agree with your proposal, renaming this to `extraArgs` will be more self explanatory.

I made the changes as you requested and tested them, everything seems to be working.
I also removed the command field because like you explained, it might not be wise to expose this in the values file.

Does that seem ok to you now ?

mbolt35:
> @mbolt35 I agree with your proposal, renaming this to `extraArgs` will be more self explanatory.
> 
> I made the changes as you requested and tested them, everything seems to be working. I also removed the command field because like you explained, it might not be wise to expose this in the values file.
> 
> Does that seem ok to you now ?

Looks great! Thanks!

PR #1259, opened 2022-02-17
Give cluster-controller access to KC cluster ID

michaelmdresser:
## What does this PR change?

Adds an environment variable for cluster-controller to know the cluster ID of its local cluster.

Knowing the ID of the cluster it is running in helps cluster-controller
work with ETL-based data to avoid problematic write scenarios, like
resizing itself, thus aborting the resize partway.

Long-term, this probably shouldn't be necessary if we put sizing state
in proper K8s structures like a CRD.

N/A

N/A

N/A

## How was this PR tested?

See https://github.com/kubecost/cluster-controller/pull/2

N/A


mbolt35 (approved):
LGTM

Issue #986, opened 2021-07-19
Support for Workload Identities


voidshard:
**What problem are you trying to solve?**
When deploying on GCP a service account key is currently required by the Kubecost pod. This is somewhat cumbersome and ideally we would support Workload Identities. Currently without a service account key secret kubecost errors & dies (in my case, no configuration is mounted & only the PROMETHEUS_SERVER_ENDPOINT env var is set), even when a Workload Identity is configured granting all perms it actually requires. That is, I believe the helm chart includes a secret that users can use for evaluation (so in theory a user building with helm will always have a secret included), but having kubecost require this to be set when it's not strictly needed seems suboptimal.

**Describe the solution you'd like**
I'm not really a kubecost expert here or familiar with the long term game plan; but In the most straightforward approach we could have kubecost print a warning if a service account key isn't found / isn't configured, but attempt to continue anyhow. Presumably kubecost would print said warning them attempt to collect the GCP data it requires & succeed (to it's great surprise I guess).

**Describe alternatives you've considered**
We could also have some configuration around how kubecost should authenticate itself with GCP and act/error accordingly (we may do this in addition).

**How would users interact with this feature?**
Ideally a user would setup a WI for each GCP cluster they oversee & then add it to the kubecost ServiceAccount either via the UI or helm/kustomize. 

gz#573

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-243) by [Unito](https://www.unito.io)

AjayTripathy:
Hi @voidshard thank you for this feedback! We'll add this to our roadmap. This might be as simple as removing our check for a key to connect to billing data if WIs "just work" with the google cloud SDK.

dwbrown2:
Currently on backlog, marking as help wanted for now.

anasinnyk:
Hi there
what is this issue status? Does it have any progress?

AjayTripathy:
Thank you so much for the PR! Let's continue discussion there but that should help!

keithhand:
Do we have an update on adding support for this?

Adam-Stack-PM:
@keithhand, Not yet. Looks like the attached PR is close to being merged for v1.97. Then will need to review Sean's issue for an upcoming release.

BenCoughlan15:
Eagerly awaiting this. 🙂

BenCoughlan15:
Hey, is this scheduled at all?

[...]

PR #1260, opened 2022-02-17
fix: gcp workload identity

anasinnyk:
Signed-off-by: Andrii Nasinnyk <anasinnyk@macpaw.com>

## What does this PR change?

I just wrapped `GOOGLE_APPLICATION_CREDENTIALS` environment variable to if statement for fixing #986 

I think this changes cannot affect any users.

- #986

## How was this PR tested?

n/a

n/a

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -375,8 +375,10 @@ spec:
                   key: prometheus-server-endpoint
             - name: CLOUD_PROVIDER_API_KEY
               value: "AIzaSyDXQPG_MHUEy9neR7stolq6l0ujXmjJlvk" # The GCP Pricing API requires a key.
+            {{- if .Values.kubecostProductConfigs.gcpSecretName }}
             - name: GOOGLE_APPLICATION_CREDENTIALS
               value: /var/configs/key.json
+            {{- end }}
             - name: CONFIG_PATH
               value: /var/configs/
             - name: DB_PATH
@@ -813,8 +815,10 @@ spec:
             {{- end }}
             - name: KUBECOST_NAMESPACE
               value: {{ .Release.Namespace }}
+            {{- if .Values.kubecostProductConfigs.gcpSecretName }}
             - name: GOOGLE_APPLICATION_CREDENTIALS
               value: /var/configs/key.json
+            {{- end }}
             - name: KUBECOST_TOKEN
               valueFrom:
                 configMapKeyRef:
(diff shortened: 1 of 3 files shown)


AjayTripathy:
Thank you so much for this PR! Were you able to verify this works for workload identities?

fourstepper:
@AjayTripathy would you mind sharing a way of how to find out if this works? I am trying a similar setup right now, so I could theoretically confirm

AjayTripathy:
Sure thing @fourstepper ! What you'd want is to give the workload the same permissions as the service key (documented here: https://github.com/kubecost/docs/blob/main/gcp-out-of-cluster.md#step-2--visit-kubecost-setup-page-and-provide-configuration-info) and then check to see if we're able to connect to BigQuery and pull in Assets beyond Kubernetes assets from the cost and usage reports. You can see that on /assets.html . A screenshot of Assets.html should be sufficient for me to confirm that this is working and get this merged!

diogokiss:
Any updates about this PR? 👀

Adam-Stack-PM:
@diogokiss Thanks for the interest in this PR.  @anasinnyk, Are you still interested in moving this forward?  Is anything blocking you on the Kubecost side?

anasinnyk:
@Adam-Stack-PM I resolve a conflict.

Adam-Stack-PM:
@anasinnyk So cool! Thank you. We are in the middle of a release. A review will come in the next few days, and are targeting v1.97 for this.

BenCoughlan15:
Would be awesome to have this, looking forward to 1.97 :)

Adam-Stack-PM:
@AjayTripathy, Friendly nudge when you have time.

AjayTripathy:
Thanks. I am not sure this won't interfere with our other methods of authentication. To that end I'd like to add @Sean-Holcomb to review... @Sean-Holcomb basically it looks like we want to use the default workload permissions IF no authentication methods have been provided to Kubecost. Is this helm check sufficient for that? My gut says no, but wanted you to check if there are other conditionals to where we should be applying the file.

Sean-Holcomb:
There is a little more to this issue than what this PR does, there will need to be some updates in the closed source, however we may want to move to having this be the recommended config method as the key files are quite cumbersome. I would merge this PR anyway because what it does is correct.

Sean-Holcomb (approved):

Adam-Stack-PM:
@Sean-Holcomb, Thanks for the additional context here. Could you open a new issue for the additional changes we will need to make in the closed source?

Adam-Stack-PM:
Note to engineer on rotation please keep the need-follow-up label on this issue until I follow up.

AjayTripathy:
@biancaburtoiu has the corresponding PR in the community edition ; can this be merged and closed when that is?

Issue #1273, opened 2022-02-25
awsstore deployment doesnt work.


wesleyorama-grindr:
**Describe the bug**
"awsstore" in cost-analyzer helm chart values file doesnt represent all required fields.

**To Reproduce**
1. Attempt to deploy helm chart with awsstore / useAwsStore set to true
2. Note deployment doesn't complete / hangs or fails with `The Deployment "kubecost-cost-analyzer-awsstore" is invalid: spec.template.spec.containers[0].image: Required value` depending on your tooling. 

**Expected behavior**
the awsstore deployment deploys successfully. 

**Screenshots**

**Collect logs (please complete the following information):**

I have already resolved the problem and cannot get this Information without breaking it again. 

gz#1439

wesleyorama-grindr:
See PR above for the fix I implemented locally.

kbrwn:
@wesleyorama-grindr Thank you for this bug report and commit! We will review asap.

kbrwn:
Hello @wesleyorama-grindr I have updated the documentation to more accurately reflect the steps to complete an AWS Marketplace Install of kubecost: https://github.com/kubecost/docs/blob/main/aws-marketplace-install.md 

You can find this doc and many others on guide.kubecost.com as well: https://guide.kubecost.com/hc/en-us/articles/4407596808087

wesleyorama-grindr:
@kbrwn Thanks, I will close this PR and issue.

PR #1274, opened 2022-02-25
FIX: add required field to awsstore object in values.yaml (ISSUE: #1273)

wesleyorama-grindr:
## What does this PR change?

adds the required key:value imageNameAndVersion to the awsstore object in the values.yaml, changes the command that the awsstore deployment runs so that it actually gets the AWS Marketplace subscription.

no

fixes issue where the awsstore deploy doesnt deploy.

- fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1273

## How was this PR tested?
in production

no, no update necessary

### DIFF
--- a/cost-analyzer/templates/awsstore-deployment-template.yaml
+++ b/cost-analyzer/templates/awsstore-deployment-template.yaml
@@ -25,7 +25,7 @@ spec:
             - image: {{ .Values.awsstore.imageNameAndVersion }}
               name: awsstore
               # Just sleep forever
-              command: [ "sleep" ]
-              args: [ "infinity" ]
+              command: [ "/bin/sh" ]
+              args: [ "-c", "while true; do go/bin/app; sleep infinity; done" ]
 {{- end }}
 {{- end }}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -667,6 +667,7 @@ serviceAccount:
 awsstore:
   useAwsStore: false
   createServiceAccount: false
+  imageNameAndVersion: alpine:latest # An image must be defined for the awsstore deployment.
 
 # readonly: false # disable updates to kubecost from the frontend UI and via POST request
 
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -667,7 +667,7 @@ serviceAccount:
 awsstore:
   useAwsStore: false
   createServiceAccount: false
-  imageNameAndVersion: alpine:latest # An image must be defined for the awsstore deployment.
+  imageNameAndVersion: gcr.io/kubecost1/awsstore:latest
 
 # readonly: false # disable updates to kubecost from the frontend UI and via POST request
 


wesleyorama-grindr:
Hey I am looking into this more and I am pretty sure this is a bigger issue than just missing a required key:value. 

In order for the register-usage permission to be used (which is required by awsstore [per the docs](https://github.com/kubecost/docs/blob/master/aws-marketplace-install.md) you have to call aws register-usage --product-code <value> --public-key-version <value> and do something with the returned JWT token. This obviously doesn't happen when the deployment is just running a alpine image and [sleeping](https://github.com/kubecost/cost-analyzer-helm-chart/blob/82e547498cf10955e75280cea2f57bb64c948df9/cost-analyzer/templates/awsstore-deployment-template.yaml#L28) 

I assume there is supposed to be an actual image here that will get the JWT from `register-usage` and decode / pass it somewhere. 

Looking though the commits on the file [Here](https://github.com/kubecost/cost-analyzer-helm-chart/commit/55675133060fe9a8bbd4810380b235f64a2edf0b) is where that image was lost, and [here](https://github.com/kubecost/cost-analyzer-helm-chart/commit/52b0155ee3274cfd18dcf81acc70b5040e9dbc49) is where the sleep commands where added. 

I am going to try to use that image that was removed from the awsstore deployment yaml locally with the sleep commands and see if that makes any difference...

wesleyorama-grindr:
I went though the commit history and poked around in the gcr.io/kubecost1/awsstore:latest and was able to successfully redeem my AWS Marketplace subscription. I have updated the files on this PR to reflect what needs to be done for this to work.

cost-analyzer/templates/awsstore-deployment-template.yaml
```diff
@@ -25,7 +25,7 @@ spec:
             - image: {{ .Values.awsstore.imageNameAndVersion }}
               name: awsstore
               # Just sleep forever
-              command: [ "sleep" ]
-              args: [ "infinity" ]
+              command: [ "/bin/sh" ]
+              args: [ "-c", "while true; do go/bin/app; sleep infinity; done" ]
```
michaelmdresser, line 29: I wonder if we need to specify command+args at all here; I feel like this should be the container's responsibility unless we're doing something non-standard

michaelmdresser (commented):
@wesleyorama-grindr Thank you for the PR and clear comments about what you're experiencing! This is not an area I have a ton of experience in; tagging in @AjayTripathy who I think has better context here.

cost-analyzer/templates/awsstore-deployment-template.yaml
wesleyorama-grindr: I thought the same thing the only input I can give is that without specifying an arg here I noticed nothing in the pod log. Specifying this arg at least showed:
`I0228 18:05:56.885938      11 main.go:28] Read region from metadata us-east-2`
`I0228 18:05:56.886008      11 main.go:38] Using productcode 8c0{redacted}9c`

That could just be on my side though.

wesleyorama-grindr: That being said, it just hangs at those two lines, and the dashboard doesnt update to show that I am on the enterprise license. 

I know the pod has the proper IRSA permissions cause I can do the call by installing the aws-cli on the container and get back the JWT using the product code given to be from the go app's logs.

AjayTripathy: Hi @wesleyorama-grindr, sorry for the late response! Which dashboard doesn't update to show the enterprise license? Nothing in kubecost automatically checks to assert that we've connected to the aws marketplace. We should just be using the marketplace for billing and redemption. You would want to just pop in the enterprise key that your account executive contact provided on setup here: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L769

AjayTripathy: That is to say-- if you see those logs being produced and the awsstore pod being run forever, this _should_ be working.

wesleyorama-grindr: hey @AjayTripathy, the settings page of Kubecost still shows the non-enterprise version of the software running. 

I got here by following the instructions from the AWS Marketplace. I did end up using the ent key, but since there is documentation showing people to do this... Maybe removing that from the marketplace templates and instead just pointing to the docs on using the ent key provided in the onboarding email would be better. Then we could clean this code up. 

Going back to the logs, I would only get those two logs and no more. The pod would run forever, but it would run forever without the sleep. Again without being able to look at the actual files that make up the image I can only assume its hanging trying to work with the marketplace. Like I said I verified IRSA permissions from that container so I know that shouldn't be a problem. 

Ultimately I have resolved my issue of not being able to use my license by using the ent key provided to me. However this process is still broken. If y'all decide to keep this process in place, I am happy to help troubleshooting and continuing to work on it.

Let me know what y'all think.

wesleyorama-grindr: This is the the instructions provided on the marketplace entitlement:
image

Here is that link for convenience:
https://github.com/kubecost/docs/blob/master/aws-marketplace-install.md

AjayTripathy: Hey Wesley, I'm so sorry that the docs were out of date.  Your point about removing the marketplace template until the process is updated is noted, and thank you for the feedback. We're currently using the AWS marketplace just for payment processing-- fulfillment is handled by us and the enterprise key.  

I've reached out to the team to re-visit those docs and help you all get a more detailed onboarding meeting with our current process for enterprise onboarding to hopefully smooth this out for the next users.

wesleyorama-grindr:
Closing this PR as it is no longer needed.

PR #1282, opened 2022-02-28
Fix isGLUUProvider env var parsing

nikovacevic:
## What does this PR change?
- Fixes malformed parsing of `.Values.saml.isGLUUProvider` that caused failure in `helm upgrade` 

- See above.

- https://kubecost.zendesk.com/agent/tickets/1442

## How was this PR tested?
Manually. @nikovacevic and @keithhand were able to reproduce using user's instructions. @nikovacevic used the fix and was able to upgrade.


Sean-Holcomb (approved):
LGTM

PR #1288, opened 2022-03-02
Allow passing custom env vars to cost-analyzer containers

### LINKED ISSUE #1287: Unable to configure frontend nginx.conf
**Describe the bug**

I run kubecost in a big 96-core server, but the default configuration of the frontend nginx container is to set `worker_processes: auto`, which will launch ~96 processes at startup. This uses a lot of unnecessary RAM and CPU.

Looks like the nginx container does have a way to override that behavior by just setting `NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE=1`, but that doesn't work for two reasons:

1. First, we'd need the helm chart to support setting `kubecostFrontend.extraEnv`. I could provide that pull request.
2. Second and more importantly, the file permissions on /etc/nginx/nginx.conf block edits. The container runs as user 1001 but the nginx.conf file is owned by root. That means the scripts in `/docker-entrypoint.d/` which edit nginx.conf fail at startup:

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
30-tune-worker-processes.sh: error: can not modify /etc/nginx/nginx.conf (read-only file system?)
/docker-entrypoint.sh: Configuration complete; ready for start up
```

**To Reproduce**
1. Run `helm install <kubecost> --set kubecostFrontend.resources.limits.cpu=250m` on a multi-core server
2. Observe that even though the container is limited to 1/4 of a CPU, nginx still logs `[notice] 1#1: start worker process N` once per core of the whole server, and that setting `NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE=1` fails to fix it

**Expected behavior**
nginx.conf should be configurable via environment variables. It might even be a good idea to set `NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE=1` by default since I can't see a downside, but at least providing the option would be nice.

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
30-tune-worker-processes.sh: error: can not modify /etc/nginx/nginx.conf (read-only file system?)
/docker-entrypoint.sh: Configuration complete; ready for start up
2022/03/02 17:08:09 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2022/03/02 17:08:09 [notice] 1#1: using the "epoll" event method
2022/03/02 17:08:09 [notice] 1#1: nginx/1.21.6
2022/03/02 17:08:09 [notice] 1#1: built by gcc 10.3.1 20211027 (Alpine 10.3.1_git20211027)
2022/03/02 17:08:09 [notice] 1#1: OS: Linux 4.18.0-348.12.2.el8_5.x86_64
2022/03/02 17:08:09 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
2022/03/02 17:08:09 [notice] 1#1: start worker processes
2022/03/02 17:08:09 [notice] 1#1: start worker process 24
2022/03/02 17:08:09 [notice] 1#1: start worker process 25
2022/03/02 17:08:09 [notice] 1#1: start worker process 26
2022/03/02 17:08:09 [notice] 1#1: start worker process 27
2022/03/02 17:08:09 [notice] 1#1: start worker process 28
2022/03/02 17:08:09 [notice] 1#1: start worker process 29
2022/03/02 17:08:09 [notice] 1#1: start worker process 30
2022/03/02 17:08:09 [notice] 1#1: start worker process 31
2022/03/02 17:08:09 [notice] 1#1: start worker process 32
2022/03/02 17:08:09 [notice] 1#1: start worker process 33
2022/03/02 17:08:09 [notice] 1#1: start worker process 34
...
2022/03/02 17:08:09 [notice] 1#1: start worker process 113
```

gz#1454

mac-chaffee:
## What does this PR change?

All containers in the cost-analyzer pod now support passing extra environment variables.

No

No effect for now, but can support future customization like in #1287

- Doesn't fully resolve #1287 but helps it

## How was this PR tested?

`helm lint`, `helm template .`

Looks like many of the "blank by default" values are left undocumented, which might be fine in this situation.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -187,6 +187,9 @@ systemProxy:
 kubecostFrontend:
   image: "gcr.io/kubecost1/frontend"
   imagePullPolicy: Always
+  # extraEnv:
+  # - name: NGINX_ENTRYPOINT_WORKER_PROCESSES_AUTOTUNE
+  #   value: "1"
   resources:
     requests:
       cpu: "10m"
@@ -200,6 +203,9 @@ kubecost:
   # Setting this to true will mean all /api/ endpoints will be unavailable
   disableServer: false
   image: "gcr.io/kubecost1/server"
+  # extraEnv:
+  # - name: SOME_VARIABLE
+  #   value: "some_value"
   resources:
     requests:
       cpu: "100m"
@@ -208,15 +214,15 @@ kubecost:
     #  cpu: "100m"
     #  memory: "256Mi"
 
-# Kubecost Metrics deploys a separate pod which will emit kubernetes specific metrics required 
-# by the cost-model. This pod is designed to remain active and decoupled from the cost-model itself. 
-# However, disabling this service/pod deployment will flag the cost-model to emit the metrics instead. 
-kubecostMetrics: 
-  # emitPodAnnotations: false 
-  # emitNamespaceAnnotations: false 
+# Kubecost Metrics deploys a separate pod which will emit kubernetes specific metrics required
+# by the cost-model. This pod is designed to remain active and decoupled from the cost-model itself.
+# However, disabling this service/pod deployment will flag the cost-model to emit the metrics instead.
+kubecostMetrics:
+  # emitPodAnnotations: false
+  # emitNamespaceAnnotations: false
   # emitKsmV1Metrics: true # emit all KSM metrics in KSM v1.
   # emitKsmV1MetricsOnly: false # emit only the KSM metrics missing from KSM v2. Advanced users only.
-  
+
   # Optional
   # The metrics exporter is a separate deployment and service (for prometheus scrape auto-discovery)
   # which emits metrics cost-model relies on. Enabling this deployment also removes the KSM dependency
@@ -239,7 +245,7 @@ kubecostMetrics:
     #    effect: "NoSchedule|PreferNoSchedule|NoExecute(1.6 only)"
     affinity: {}
 
-    # Service Monitor for Kubecost Metrics 
+    # Service Monitor for Kubecost Metrics
     serviceMonitor:
       enabled: false
       additionalLabels: {}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
This looks good to me! @mac-chaffee would you mind providing a sample for how this might be used in conjunction with values.yaml and the right env variable to solve #1288 ? That way when we make the call about defaulting this or not, it's easy to just uncomment what you've added from values.yaml and shipit.

AjayTripathy (approved):
Approved, would just love to see a sample in values.yaml!

cost-analyzer/values.yaml
```diff
@@ -208,15 +214,15 @@ kubecost:
     #  cpu: "100m"
     #  memory: "256Mi"
 
-# Kubecost Metrics deploys a separate pod which will emit kubernetes specific metrics required 
-# by the cost-model. This pod is designed to remain active and decoupled from the cost-model itself. 
-# However, disabling this service/pod deployment will flag the cost-model to emit the metrics instead. 
-kubecostMetrics: 
-  # emitPodAnnotations: false 
-  # emitNamespaceAnnotations: false
```
mac-chaffee, line 216: My editor trims trailing white space unfortunately.

AjayTripathy:
Thank you @mac-chaffee !

PR #1255, opened 2022-02-17
Fixed cluster-ultil and label-cost-util in grafana

jessegoodier:
## What does this PR change?
Fixed cluster-ultil and label-cost-util dashboards in grafana

## Does this PR rely on any other PRs?

- no

## How does this PR impact users? (This is the kind of thing that goes in release notes!)
Fixed N/A values and errors on the Label costs & utilization and Cluster cost & utilization metrics dashboards.
The issue with the Label costs & utilization is also fixed in Grafana 8.3.4

## Links to Issues or ZD tickets this PR addresses or fixes

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1088 

## How was this PR tested?
redeployed helm chart with updated dashboards
cluster cost
label costs

## Have you made an update to documentation?
Not needed

PR #1289, opened 2022-03-02
[DRAFT] Revert "Fixed cluster-ultil and label-cost-util in grafana"

AjayTripathy:
Reverts kubecost/cost-analyzer-helm-chart#1255 to see if it fixes grafana.


AjayTripathy:
not the issue.

PR #1299, opened 2022-03-09
Fix network costs service duplicate labels

### LINKED ISSUE #1150: YAML Key-duplication

kuznero:
## What does this PR change?

Fixes #1150 - this PR simply removes duplicate from network costs service labels.

No impact

#1150 

## How was this PR tested?

The scope is very limited to exactly the place where it needed to be fixed. Diffed the result to ensure nothing else was changed.

No need

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml
@@ -9,7 +9,7 @@ metadata:
     prometheus.io/scrape: "{{ .Values.networkCosts.prometheusScrape }}"
     prometheus.io/port: {{ (quote .Values.networkCosts.port) | default (quote 3001) }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


mbolt35 (approved):
Thank you!

PR #1302, opened 2022-03-09
Save mount path of productKey secret into environment variable

### LINKED ISSUE #1021: Allow product key to be read from env or custom file path
Some k8s clusters use tooling to read secret data (password/auth etc) from vault. Currently the product key can only be read from a secret/configmap that is mounted at a specific file path. This filepath should be configurable or product key should be able to be read from an environment variable to support vault an other in-memory secret injection tools.

biancaburtoiu:
## What does this PR change?
This PR defines:
- a `mountPath` parameter and calls it to populate a new `PRODUCT_KEY_MOUNT_PATH` environment variable

This PR should only be merged **after** the read source for the `ProductKeySecret` is modified in [this file](https://github.com/kubecost/kubecost-cost-model/blob/f14c1ab3aa6bedfdd81f8dc1cf7908e9f26a0e58/pkg/product/productkey.go#L19).

Users can now customise the path at which the product key secret is mounted: the user can declare the path at which an external secret provisioner has mounted the secret, informing us of its location (mountPath)

- fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1021

## How was this PR tested?
`kubectl exec container-name -it sh` and verifying that the expected files exist:
- `printenv PRODUCT_KEY_MOUNT_PATH` returns the path defined by the user in values.yaml

Defined the new field in the README.

### DIFF
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -46,3 +46,4 @@ Parameter | Description | Default
 `serviceAccount.create` | Set this to `false` if you want to create the service account `kubecost-cost-analyzer` on your own | `true`
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
+`kubecostProductConfigs.productKey.mountPath` | Customize path at which to mount the product key secret | `"/var/configs/productkey"`
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -46,4 +46,4 @@ Parameter | Description | Default
 `serviceAccount.create` | Set this to `false` if you want to create the service account `kubecost-cost-analyzer` on your own | `true`
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
-`kubecostProductConfigs.productKey.mountPath` | Set the path at which the product key secret is mounted (eg. by a secrets provisioner) | `"/var/configs/productkey"`
+`kubecostProductConfigs.productKey.mountPath` | Set the path at which the product key secret is mounted (eg. by a secrets provisioner) | `"/var/configs/productkey/productkey.json"`
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -46,4 +46,4 @@ Parameter | Description | Default
 `serviceAccount.create` | Set this to `false` if you want to create the service account `kubecost-cost-analyzer` on your own | `true`
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
-`kubecostProductConfigs.productKey.mountPath` | Set the path at which the product key secret is mounted (eg. by a secrets provisioner) | `"/var/configs/productkey/productkey.json"`
+`kubecostProductConfigs.productKey.mountPath` | Use instead of `kubecostProductConfigs.productKey.secretname` to declare the path at which the product key file is mounted (eg. by a secrets provisioner) | `N/A`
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -287,7 +287,7 @@ spec:
             {{- if .Values.kubecostProductConfigs.productKey }}
             {{- if .Values.kubecostProductConfigs.productKey.secretname }}
             - name: productkey-secret
-              mountPath: /var/configs/productkey
+              mountPath: {{ .Values.kubecostProductConfigs.productKey.mountPath }}
```
AjayTripathy, line 290: I would add a default here in case .Values.kubecostProductConfigs.productKey.mountPath is "" or unset, as you did below.

dwbrown2:
Congrats on first PR so fast!!

dwbrown2:
I may test with different mount paths + keys just to make sure this is working as expected!

cost-analyzer/README.md
```diff
@@ -46,3 +46,5 @@ Parameter | Description | Default
 `serviceAccount.create` | Set this to `false` if you want to create the service account `kubecost-cost-analyzer` on your own | `true`
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
+`kubecostProductConfigs.productKey.volumeMountPath` | Customize the path at which the product key secret should be mounted (as a volume) | `"/var/configs/productkey"`
+`kubecostProductConfigs.productKey.envvarMountPath` | Customize the path at which the product key secret is mounted (as an environment variable) | `"/var/configs/productkey"`
```
biancaburtoiu, line 50: Open question - review phrasing for clarity!

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -398,6 +398,10 @@ spec:
             {{- else }}
               value: production
             {{- end }}
+            {{- if .Values.kubecostProductConfigs.productKey }}
+            - name: PRODUCT_KEY_MOUNT_PATH
+              value: {{ .Values.kubecostProductConfigs.productKey.envvarMountPath | default "/var/configs/productkey" }}
```
biancaburtoiu, line 403: Open question - thoughts on the clarity of the names `volumeMountPath` and `envvarMountPath`?

AjayTripathy:
Hey @biancaburtoiu I gave this some thought and I actually think the only thing we want to add is the envVarMountPath and call it just MountPath. I don't think it's necessary to give users the option of where in /var/configs to mount the secret. Either you're provisioning via another tool and placing it in a configurable location or just using our default flow, in which case MountPath should just be set to /var/configs. Hopefully that makes sense!

cost-analyzer/README.md
```diff
@@ -46,3 +46,4 @@ Parameter | Description | Default
 `serviceAccount.create` | Set this to `false` if you want to create the service account `kubecost-cost-analyzer` on your own | `true`
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
+`kubecostProductConfigs.productKey.mountPath` | Use instead of `kubecostProductConfigs.productKey.secretname` to declare the path at which the product key file is mounted (eg. by a secrets provisioner) | `N/A`
```
biancaburtoiu, line 49: Open question - I used `N/A` here given that this field is optional. If it is undefined, clients (eg. the `/getProductKey()` endpoint) will rely on reading at the fallback path `/var/configs/productkey/productkey.json`, but I thought that it wasn't necessary to share that information here

AjayTripathy: I think that's fine.

AjayTripathy (approved):

PR #1306, opened 2022-03-10
Allow setting customised fqdn values for NGINX config (kubecostFrontend)

### LINKED ISSUE #1271: Allow set fqdn in nginx config (kubecostFrontend)
Our cluster coredns do not support following domains
`app.namespace`
It can support only fqdn (app.namespace.svc.cluster.local) or just namespace level, eg `app`
For example i want to set grafana upstream host to `kubecost-grafana.kubecost.svc.cluster.local`, but currently I can't.
The same with cost-model and server, I can't set it to `localhost:9003` and `localhost:9001` accordingly.
Related code:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/aefe69f814ef722cffd02b34eacae2c090248277/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L48-L72

Add some new values to the values.yaml file to set this fqdn in cost-analyzer-frontend-config-map-template.yaml for:
* upstream api 
* upstream model
* upstream grafana
* upstream clustercontroller
Just use this values if they set (server value with port) instead trying to build them from different parts.

Right now I use **VERY PAINFUL** workaround by completely replacing configmap `nginx-conf` and reload `kubecost-cost-analyzer` pod.

Just set this value under `kubecostFrontend` in local helm values.yaml file.

gz#1437

biancaburtoiu:
## What does this PR change?
This PR defines several new values to be populated in values.yaml:
- `.kubecostFrontend.api.fqdn`
- `.kubecostFrontend.model.fqdn`
- `.clusterController.fqdn`
- `.global.grafana.fqdn`

If any of these three values are defined by the user, they will be applied in the `upstream api/model/clusterController/grafana` NGINX configuration, otherwise the existing defaults would be used.

No

Users can now set customised fqdn values in the NGINX config. 

- fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1271 

## How was this PR tested?
After defining the placeholders and populating them in `values.yaml`, I verified that the customised values appear in the `nginx-conf` map under the `upstream` definitions, as well as that if no `.fqdn` values exist, the defaults are used in creating the configmap.

Updated the list at the [cost-analyzer-helm-chart README](https://github.com/kubecost/cost-analyzer-helm-chart#kubecost-helm-chart) to include the newly-added values.

### DIFF
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -50,3 +50,4 @@ Parameter | Description | Default
 `kubecostFrontend.api.fqdn` | Customize the upstream api FQDN | `computed in terms of the service name and namespace`
 `kubecostFrontend.model.fqdn` | Customize the upstream model FQDN | `computed in terms of the service name and namespace`
 `clusterController.fqdn` | Customize the upstream cluster controller FQDN | `computed in terms of the service name and namespace`
+`global.grafana.fqdn` | Customize the upstream grafana FQDN | `computed in terms of the release name and namespace`
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -47,6 +47,6 @@ Parameter | Description | Default
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
 `kubecostProductConfigs.productKey.mountPath` | Use instead of `kubecostProductConfigs.productKey.secretname` to declare the path at which the product key file is mounted (eg. by a secrets provisioner) | `N/A`
-`kubecostFrontend.api.fqdn` | Customize the upstream api FQDN | `kubecost-api.kubecost.svc.cluster.local`
-`kubecostFrontend.model.fqdn` | Customize the upstream model FQDN | `kubecost-model.kubecost.svc.cluster.local`
-`clusterController.fqdn` | Customize the cluster controller FQDN | `kubecost-cluster-controller.kubecost.svc.cluster.local`
+`kubecostFrontend.api.fqdn` | Customize the upstream api FQDN | `kubecost-api.kubecost.svc.cluster.local:9001`
+`kubecostFrontend.model.fqdn` | Customize the upstream model FQDN | `kubecost-model.kubecost.svc.cluster.local:9003`
+`clusterController.fqdn` | Customize the upstream cluster controller FQDN | `kubecost-cluster-controller.kubecost.svc.cluster.local:9731`
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -47,6 +47,6 @@ Parameter | Description | Default
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
 `kubecostProductConfigs.productKey.mountPath` | Use instead of `kubecostProductConfigs.productKey.secretname` to declare the path at which the product key file is mounted (eg. by a secrets provisioner) | `N/A`
-`kubecostFrontend.api.fqdn` | Customize the upstream api FQDN | `kubecost-api.kubecost.svc.cluster.local:9001`
-`kubecostFrontend.model.fqdn` | Customize the upstream model FQDN | `kubecost-model.kubecost.svc.cluster.local:9003`
-`clusterController.fqdn` | Customize the upstream cluster controller FQDN | `kubecost-cluster-controller.kubecost.svc.cluster.local:9731`
+`kubecostFrontend.api.fqdn` | Customize the upstream api FQDN | `computed in terms of the service name and namespace`
+`kubecostFrontend.model.fqdn` | Customize the upstream model FQDN | `computed in terms of the service name and namespace`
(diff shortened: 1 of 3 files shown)


cost-analyzer/values.yaml
```diff
@@ -615,6 +619,7 @@ clusterController:
   enabled: false
   image: gcr.io/kubecost1/cluster-controller:v0.0.2
   imagePullPolicy: Always
+  fqdn: TODO-some-fqdn-example
```
biancaburtoiu, line 623: TODO: I'd like some help in giving some suitable examples here!

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -46,26 +46,44 @@ data:
...
     }
 {{- end }}
 {{- end }}
 
 {{- if .Values.global.grafana.proxy }}
     upstream grafana {
 {{- if .Values.global.grafana.enabled }}
-        server {{ .Release.Name }}-grafana.{{ .Release.Namespace }};
-{{ else }}
         server {{.Values.global.grafana.domainName}};
```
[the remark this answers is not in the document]
AjayTripathy: I think you want to do the same thing as above where we try to use the predefined value or .Values.global.grafana.domainName -- no change should be required here at all. It already supports a custom URL definition.

cost-analyzer/values.yaml
AjayTripathy: Sorry, didn't see which block this was under. on this particular line you'll want to add the default servicename for the cluster-controller-- just something like kubecost-cluster-controller.kubecost.svc.cluster.local.

biancaburtoiu: thanks! just to check - a port as well?

biancaburtoiu: Thanks! I've added them commented-out by default and updated the examples.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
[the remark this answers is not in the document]
biancaburtoiu: I've removed the modification from the `upstream grafana` config

AjayTripathy (approved):

AjayTripathy (approved):

cost-analyzer/README.md
```diff
@@ -46,3 +46,6 @@ Parameter | Description | Default
 `serviceAccount.create` | Set this to `false` if you want to create the service account `kubecost-cost-analyzer` on your own | `true`
 `tolerations` | node taints to tolerate | `[]`
 `affinity` | pod affinity | `{}`
+`kubecostFrontend.api.fqdn` | Customize the upstream api FQDN | `kubecost-api.kubecost.svc.cluster.local:9001`
+`kubecostFrontend.model.fqdn` | Customize the upstream model FQDN | `kubecost-model.kubecost.svc.cluster.local:9003`
+`clusterController.fqdn` | Customize the upstream cluster controller FQDN | `kubecost-cluster-controller.kubecost.svc.cluster.local:9731`
```
evgenii-denisov, line 51: Actually default values is different right now.
For example, as I see default value for `kubecostFrontend.api.fqdn` equal to empty string.
And in this case upstream host is `{{ $serviceName }}.{{ .Release.Namespace }}:9001;` and not a `kubecost-api.kubecost.svc.cluster.local:9001`

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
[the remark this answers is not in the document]
evgenii-denisov: Please also add `fqdn` value for grafana too, or all this PR will be useless, at least for me.
I need both `Values.global.grafana.enabled=true` and `Values.global.grafana.fqdn` or by default it should be `{{ .Release.Name }}-grafana.{{ .Release.Namespace }}.svc.cluster.local`

cost-analyzer/README.md
biancaburtoiu: Thanks for spotting that @evgenii-denisov! I've removed those examples as they were not appropriate.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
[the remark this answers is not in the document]
evgenii-denisov: I prefer to do it in more consistent way, like this (indentation just for readability):
```
{{- if .Values.global.grafana.proxy }}
    upstream grafana {
 {{- if .Values.global.grafana.enabled }}
  {{- if .Values.global.grafana.fqdn }}
        server {{ .Values.global.grafana.fqdn }};
  {{- else }}
        server {{ .Release.Name }}-grafana.{{ .Release.Namespace }};
  {{- end }}
 {{- else }}
        server {{.Values.global.grafana.domainName}};
{{- end }}
    }
{{- end }}
```

AjayTripathy:
This LGTM! Were you able to verify this works according to the same testing methods in the PR description?

AjayTripathy (approved):

PR #1324, opened 2022-03-23
GH actions Values readme generation

keithhand:
## What does this PR change?
Adds Github action to update the `cost-analyzer/README.md` with annotations pre-supplied in the `cost-analyzer/values.yaml`. 

Improved README.md consistency on the primary chart.

## How was this PR tested?
The action relies on the [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm) from bitnami-labs. I tested using this fork from my repository. Here is an example of a successful [action run](https://github.com/keithhand/cost-analyzer-helm-chart/runs/5653548925?check_suite_focus=true) and [generated commit](https://github.com/keithhand/cost-analyzer-helm-chart/pull/3/commits/e7f16261bca782eff0721f13db79e848e545ed5f). You can also test locally by copying the steps the action takes to install and generate the README.md.

```
## Checkout
cd cost-analyzer-helm-chart

## Install readme-generator-for-helm
git clone https://github.com/bitnami-labs/readme-generator-for-helm
npm install ./readme-generator-for-helm

## Update README.md
./readme-generator-for-helm/bin/index.js -v cost-analyzer/values.yaml -r cost-analyzer/README.md
```

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-77) by [Unito](https://www.unito.io)


srpomeroy:
Very cool!

Adam-Stack-PM:

chipzoller:
I wasn't aware this action existed, but after looking over it I like this change. The downside is it will require some extensive annotation of the values file so as to be scraped by this action, but I think that'd be a good exercise for us to clean and tighten up the chart.

@keithhand, would you mind resolving conflicts?

@jessegoodier, @thomasvn: What are your thoughts here? I'm on-board with performing the field annotations with some assistance from you both to ensure accuracy.

thomasvn:
@chipzoller In my experience, I do not often see users consulting `README.md` for default params. Instead, most often go to the `values.yaml`. I'm in favor of removing these values from the `README.md` is possible!

chipzoller:
That would also be my preference. But I think we need to add more comprehensive comments in the values file so it is more self-documented than it is currently.

keithhand:
Personally, I have the opposite opinion. I generally reference READMEs when possible. Doing some quick maths from artifacthub.io, it looks like ~60% of the top helm charts include values in their readmes (3/5, 6/10, 8/15, 12/20). Either way, I agree with Chip that our values file is likely underdocumented if we intend on using that as a reference point. There are a ton of values that don't have a description and many aren't included in the values file at all.

jessegoodier:
I like it, seems like many users expect this. I certainly look at the readme first in an repo.

Only question is how hard it will be to maintain.

srpomeroy:
> There are a ton of values that don't have a description and many aren't included in the values file at all.

x2. I've had to do lots of digging for additional values besides what's in the base values.yaml. I wonder if there is a linter to verify that the values.yaml is complete? Or is it even best practice to include all available values in the values.yaml?

Is there any other tool or process that could benefit from codifying the comments so they can be used to generate the readme? Docs site? In product?

chipzoller:
So what I'd recommend is kind of a hybrid approach. I've seen many other projects do this and it's helpful to users, especially newcomers so they don't have to read the entire values file, but also makes maintenance easier by not having multiple documents with the full values set which always diverge.

* Use a README file for chart usage and _general_ structure of the layout which directs users where to look for either common configuration changes (ex., Federated, platform specific like EKS, etc.) and what the large components are.
* Use the values file as the single source of truth for ALL the params which can be changed in the chart. This means docs.kubecost.com and every place else just directs to the README and the README defers to the values file.

@srpomeroy, to your point, we need to refactor this Helm chart which we want to do for 2.0. At that time, we can go through it with a fine-toothed comb and ensure the values file records all possible definable values to eliminate these things. We can also ensure that all sections and fields have comments, they're descriptive, and so function as in-line documentation to the chart.

chipzoller:
Closing as this is stale. We'll revisit this when we revamp the chart which will include breaking changes but better organization and self-documentation.

PR #1335, opened 2022-03-29
disable cloud usage if cloud assets was disabled

Sean-Holcomb:
## What does this PR change?
This PR makes it so that users who had cloudAssets disabled will seamlessly continue to have cloud usage disabled.

- 
- 

Users should notice no change

- 
- 

## How was this PR tested?
Tested with all variations of values for `.Values.kubecostModel.etlCloudAsset` and `.Values.kubecostModel.etlCloudUsage`. If either are disabled Cloud Usage was disabled.
Screen Shot 2022-03-29 at 12 42 54 PM


AjayTripathy:
Thanks! This will make the upgrade a lot easier.

AjayTripathy (approved):

PR #1338, opened 2022-03-30
Ajay tripathy fix ne

AjayTripathy:
## What does this PR change?
Fix issue where ne was invalidly typed. Previously got
```
UPGRADE FAILED: template: cost-analyzer/templates/cost-analyzer-deployment-template.yaml:533:19: executing "cost-analyzer/templates/cost-analyzer-deployment-template.yaml" at <eq .Values.kubecostModel.etlCloudAsset false>: error calling eq: invalid type for comparison
```

Basically the behaviour should be:
1. If the user is using the new flag, great, use that but don't autodefine it in values.yaml
2. if user was using the legacy flag, use the state of the legacy flag
3.Default to true.

## How was this PR tested?
just with helm upgrade, please test the other cases.


Sean-Holcomb (approved):
Lgtm

PR #1330, opened 2022-03-27
remove server block and associated values

AjayTripathy:
## What does this PR change?
Remove cost-analyzer-server which has been deprecated for several releases!

Also repairs the default grafana install.

## Does this PR rely on any other PRs?
A number of PRs deprecating cost-analyzer-server.

## How does this PR impact users? (This is the kind of thing that goes in release notes!)
Fewer containers to run, save memory

## Links to Issues or ZD tickets this PR addresses or fixes

- 
- 

## How was this PR tested?
atripathy@Ajays-MacBook-Pro builds % kubectl get pods -n kubecost
NAME                                          READY   STATUS     RESTARTS   AGE
kubecost-cost-analyzer-578f956bcd-pwcng       2/2     Running    0          18m

Plus integration testing to make sure all endpoints referencing this have been removed.

For grafana:
After restarting the grafana pod:
Screen Shot 2022-03-28 at 1 25 49 PM

## Have you made an update to documentation?
There are many references to cost-analyzer-server that should be removed scattered throughout the docs. Let's make a note and remove these as we find them.

PR #1339, opened 2022-03-31
remove deprecated server port

keithhand:
## What does this PR change?
This PR removes the [deprecated](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1330) 9001 port used for the cost-analyzer-server from the service template. This port will cause service failures on EKS because when you create a service with the type LoadBalancer, it appears to make a health check for `.spec.ports[0]`. Because this port returned an empty response, the LoadBalancer would remain stuck in an `OutOfService` state.

Fixed bug where cost-analyzer-service health checks could fail

## How was this PR tested?
I ran an upgrade against a cluster in EKS to verify that this would allow the LoadBalancer to enter an `InService` state and upgraded a cluster in GKE to confirm that this wouldn't cause issues in another environment.

https://github.com/kubecost/docs/pull/212

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-service-template.yaml
@@ -27,9 +27,6 @@ spec:
   type: ClusterIP
 {{- end }}
   ports:
-    - name: tcp-server
-      port: 9001
-      targetPort: 9001
     - name: tcp-model
       port: 9003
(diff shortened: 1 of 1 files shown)


AjayTripathy:
Great find!! Amazing that we actually caught this in an integration test. Can we also get this cherrypicked in v1.92?

AjayTripathy (approved):

PR #1344, opened 2022-04-01
Fix issue in the helm chart to run multiple kubecost

dschunack:
## What does this PR change?

We want to run multiple kubecost on the same K8S/EKS to test new Versions.
This was not possible with the existing helm chart.

I changed the following:

- Fixed some helm chart templates to run multiple kubecost for Testing
  - Release Name instead of ServiceAccount Name for Role, ClusterRole and Bindings.
  - The Name of the Kubecost-network PSP was changed to use variables like it is in the Kubecost PSP

It's not really practical to change for any new Kubecost on the same k8s the serviceAccount Name. Instead, I changed it to use the Chart.Name. It's also possible to use the "fullname", but this caused changes for existing deployments. 

- No

- No Breacking changes for existing Installations

- [#1108](https://github.com/kubecost/cost-model/issues/1108)

## How was this PR tested?
- It was tested on several EKS with new and existing kubecost deployments.

- No

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
@@ -3,13 +3,13 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: {{ template "cost-analyzer.serviceAccountName" . }}
+  name: {{ .Release.Name }}
   labels:
     {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
-  name: {{ template "cost-analyzer.serviceAccountName" . }}
+  name: {{ .Release.Name }}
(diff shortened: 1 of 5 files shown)


kirbsauce:
Thanks for the PR @dschunack !  We'll get a review started!

dwbrown2:
@kirbsauce @AjayTripathy would this make current release that is in flight?

kirbsauce:
@dwbrown2 I'll defer to @AjayTripathy and @teevans as to whether or not we should squeeze this in.  While probably fine, my .02 would be to err on the side of caution and wait for next release or interim patch.

AjayTripathy:
Let's get this reviewed and in the next release.

AjayTripathy:
@nealormsbee mind picking this one up? Should be quick and I think you were helping out the support rotaiton this upcoming week?

nealormsbee (approved):
This all looks good to me. I was able to install multiple releases of Kubecost with network pods and network PSP enabled, where previously I could not.

Same thing with reusing a service account across releases.

PR #1347, opened 2022-04-04
Fix duplicate app labels and expose Grafana PSP config

CameronHall:
### Duplicate app label
The Kubecost metric deployment template uses the `cost-analyzer.commonLabels` template that includes the label `app: cost-analyzer`. However, in the kubecost-metric-deployment-template we also specify:
```
app: {{ template "kubecost.kubeMetricsName" . }}
```

Subsequently leaving us with two app labels. This fix uses the `kubecost.chartLabels` template instead which is the exact same thing without the `app` label.

### Grafana PSP Config
Exposes the PSP configuration for the Grafana Helm chart. It's nice to have this visible if you're trying to get rid of those pesky deprecation warnings.

No

```
fix: Fixed duplicate `app` label on Kubecost Metric Exporter or Agent deployments.
docs: Exposed Grafana Pod Security Policy (PSP) configuration in chart config
```

N/A

## How was this PR tested?
Locally

This isn't necessary.

### DIFF
--- a/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml
+++ b/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml
@@ -6,7 +6,7 @@ kind: Deployment
 metadata:
   name: {{ template "kubecost.kubeMetricsName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+    {{ include "kubecost.chartLabels" . | nindent 4 }}
     app: {{ template "kubecost.kubeMetricsName" . }}
 {{- with .Values.kubecostMetrics.exporter.labels }}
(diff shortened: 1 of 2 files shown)


CameronHall:
It looks like someone has beaten me to fix the app labels 😄  Though I'd argue my solution is simpler.

kirbsauce:
Thanks @CameronHall !  We'll have a look!

AjayTripathy (approved):

AjayTripathy:
Looks like we went with 
```    {{ unset (include "cost-analyzer.commonLabels" . | fromYaml) "app" | toYaml | nindent 4 }}
```
in another PR to fix duplication, but happy to accept the second fix. Thanks so much for this, I'll take it from here!

PR #1370, opened 2022-04-14
Grafana Enabled Env Var

nealormsbee:
## What does this PR change?
Adds an env var to the cost model indicating whether Grafana is enabled.

* https://github.com/kubecost/cost-analyzer-frontend/pull/598
* https://github.com/kubecost/kubecost-cost-model/pull/727
* https://github.com/kubecost/docs/pull/220/files

It does not.

N/A

## How was this PR tested?

By Helm upgrading with the local chart, setting `globals.grafana.enabled=false`, and also observing that `grafana.enabled` continues to default to `true`.

I have not.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -336,6 +336,18 @@ spec:
             {{- end }}
             {{- end }}
           env:
+            {{- if .Values.global.grafana}}
+            {{- if .Values.global.grafana.enabled }}
+            - name: GRAFANA_ENABLED
+              value: "true"
+            {{- else }}
+            - name: GRAFANA_ENABLED
+              value: "false"
+            {{- end }}
+            {{- else }}
+            - name: GRAFANA_ENABLED
+              value: "false"
+            {{- end }}
             {{- if .Values.kubecostModel.extraEnv -}}
             {{ toYaml .Values.kubecostModel.extraEnv | nindent 12 }}
             {{- end }}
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -336,6 +336,18 @@ spec:
...
+            - name: GRAFANA_ENABLED
+              value: "true"
+            {{- else }}
+            - name: GRAFANA_ENABLED
+              value: "false"
+            {{- end }}
+            {{- else }}
+            - name: GRAFANA_ENABLED
+              value: "false"
+            {{- end }}
```
mbolt35, line 350: I'd suggest doing something like:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/64e022c33024a3944a568d6b29844033aca1e7f6/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L511-L513

```
{{- if .Values.global.grafana }}
- name: GRAFANA_ENABLED
   value: {{ (quote .Values.global.grafana.enabled) | default (quote false) }}
{{- end}}

And then in the code, when we check for the `env.IsGrafanaEnabled()` we should return `GetBool("GRAFANA_ENABLED", false)`

nealormsbee: Updated. Thanks for the tip! I'm not familiar with all the available things in templates like `quote` and `default`, but it definitely felt like this could be clearer.

nealormsbee:
Have now tested this out and the env var does propagate correctly.

nealormsbee:
Tested out with changes and the env var propagates as expected.

michaelmdresser (approved):

Issue #1358, opened 2022-04-08
Support Custom Google Analytics


kirbsauce:
We would like to be able to enable google analytics on our kubecost deployments. To do this, we need to be able to add a tag right after the head tag on each page of kubecost. Is adding tags to the webpage code something that we could get support for? I am specifically referencing the “Add the global site tag directly to your web pages” section in the “Set up data collection (for websites)” section on this google analytics help page https://support.google.com/analytics/answer/9304153?hl=en&ref_topic=9303319#zippy=%2Cadd-the-global-site-tag-directly-to-your-web-pages

(created from [Zendesk ticket #1633](https://kubecost.zendesk.com/agent/tickets/1633))
gz#1633

kirbsauce:
We actually already optionally support GA today. So I think we would just need to support passing a custom tracking ID via a helm flag. Would that work in this case? We can move this to a public issue if that's easier for user input.

@dwbrown2

dwbrown2:
@teevans and I to start reviewing this week! currently expected for next release, v1.93

kirbsauce:
FYI @AdamStack18

dwbrown2:
> Note: this will provide basic frontend analytics, e.g. page views, time on page, etc. 

@teevans let us know if we can get someone to add in a BE endpoint with a helm chart flag. That will hopefully keep you moving really fast on FE!

teevans:
Let me see what I can come up with tomorrow for this. This is a bigger lift because of the SPA disconnect. But, we're gonna knock it out.

PR #1373, opened 2022-04-15
Added in Customer Google Tag to values and environment variables

teevans:
## What does this PR change?
This add the GoogleCustomerTag value to the Values file in the helm chart under kubecostFrontend, and adds the GoogleCustomerTag as an environment variable to be returned by /getConfigs

This allows users to use their own Google Global Tag to track the utilization of their Kubecost frontend! 

- #1358 

## How was this PR tested?
Manually by deploying to my GKE cluster.

No - Will do prior to closing after we find out if I did this right.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -198,6 +198,8 @@ kubecostFrontend:
     #limits:
     #  cpu: "100m"
     #  memory: "256Mi"
+#  customerGoogleTag allows you to embed your Google Global Site Tag to track usage of Kubecost.
+#  customerGoogleTag: G-XXXXXXXXX
 #  api:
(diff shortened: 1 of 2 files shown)


biancaburtoiu (approved):

cost-analyzer/values.yaml
```diff
@@ -198,6 +198,8 @@ kubecostFrontend:
     #limits:
     #  cpu: "100m"
     #  memory: "256Mi"
+#  customerGoogleTag allows you to embed your Google Global Site Tag to track usage of Kubecost.
```
dwbrown2, line 201: @teevans sorry for catching this late, but what if we moved this under the reporting block? that's where all related analytics flags are. 

https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/values.yaml#L609

PR #1377, opened 2022-04-18
Don't use eq to default a possibly-nil value

michaelmdresser:
## What does this PR change?

`kubecostModel.etlCloudAsset` is nil by default. Before this change,
running `helm template` would give the following error:

`
Error: template: cost-analyzer/templates/cost-analyzer-deployment-template.yaml:533:19: executing "cost-analyzer/templates/cost-analyzer-deployment-template.yaml" at <eq .Values.kubecostModel.etlCloudAsset false>: error calling eq: incompatible types for comparison
`

After the change, the error is gone.

cc @Sean-Holcomb does this maintain the default behavior you expect for this env var?

N/A

N/A (I believe the current release doesn't have this issue, I assume this regressed in between the release and now.)

N/A

## How was this PR tested?

```
→ helm template ./cost-analyzer | grep -A 1 -i ETL_CLOUD
            - name : ETL_CLOUD_USAGE_ENABLED
              value: "false"
```

```
→ helm template ./cost-analyzer --set kubecostModel.etlCloudAsset=true | grep -A 1 -i ETL_CLOUD
            - name : ETL_CLOUD_USAGE_ENABLED
              value: "true"
```

N/A

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -530,11 +530,7 @@ spec:
               value: {{ (quote .Values.kubecostModel.etlToDisk) | default (quote true) }}
             {{- end }}
             - name : ETL_CLOUD_USAGE_ENABLED
-            {{- if eq .Values.kubecostModel.etlCloudAsset false }}
-              value: "false"
-            {{- else }}
-              value: {{ (quote .Values.kubecostModel.etlCloudUsage) | default (quote true) }}
-            {{- end }}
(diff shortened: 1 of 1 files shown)


michaelmdresser:
Actually, @Sean-Holcomb should this default to true instead of false? I may have misread the current logic.

michaelmdresser:
Closing. This was fixed originally in v1.92 with https://github.com/kubecost/cost-analyzer-helm-chart/pull/1338 and is being brought back to develop with https://github.com/kubecost/cost-analyzer-helm-chart/pull/1380

PR #1381, opened 2022-04-18
Changing version

nareshagrawal:
Changing to lastest version i.e. 1.92.0

## What does this PR change?
Changing to latest version 1.92.0 

- 
- 

- 
-

### DIFF
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "1.91.0"
+appVersion: "1.92.0"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: "1.91.0"
+version: "1.92.0"
 annotations:
   "artifacthub.io/links": |
     - name: Homepage


AjayTripathy:
Hey Naresh! Thank you for this PR. I'm happy to make this update but just as a note you should be deploying the latest code from master https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/cost-analyzer/Chart.yaml#L2
 or v1.92 https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.92/cost-analyzer/Chart.yaml#L2
 
 Where version numbers are managed.

AjayTripathy (approved):

nealormsbee:
We set versions in release branches and basically ignore what's in develop these days, right @AjayTripathy ? I have no issue with this PR, but I don't think the version being set here is actually used anywhere.

One proposal by @michaelmdresser  that I like is setting the version in `kubecost:develop` to `vDevelop` or similar, to be explicit that develop branch is not indicative of a versioned release.

michaelmdresser:
> We set versions in release branches and basically ignore what's in develop these days, right

Yes. Version strings are handled by release scripts these days. The version in develop is arbitrary and will always be overwritten when doing a proper release of the chart (and our other code).

> One proposal by @michaelmdresser that I like is setting the version in kubecost:develop to vDevelop or similar, to be explicit that develop branch is not indicative of a versioned release.

I'd very much like to do this. I think Helm might require SemVer chart versions, though.

PR #1385, opened 2022-04-25
Allow more Prometheus flexibility

rlees85:
## What does this PR change?
* Add setting `prometheusScrape` for Metrics Exporter (default true) to allow turning off the default set of Prometheus scrape annotations.
* Allow setting custom annotations for Metrics Exporter service.
* Allow setting custom annotations for Network Costs service.
* Create the Network Costs service always, but only add Prometheus annotations if `prometheusScrape` is set to true.
* Remove `kubecostMetrics.exporter.annotations` (not used anywhere).
* Adds `ports` declaration in Metrics Exporter deployment.

N/A 

The Network Costs server will always be created but without Prometheus scrape annotations (by default). There should be no change in behaviour.

Everything else should behave as before.

fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1395

## How was this PR tested?
- `helm template cost-analyzer --values ...`
- `kubectl apply --dry-run=server -f ...`
- `diff --color ... ...`

No, `README.md` does not contain a complete list of values anyway - infarct it seems miles off.

## Notes
This is a bit of an awkward one because normally Helm charts do not include Prometheus scrape annotations by default, to allow people to specify their own. However the Kubecost situation is different since it ships with Prometheus by default and is trying to be as light touch as possible (I guess).

To accommodate this I have set defaults that match the current behaviour, but advanced users who want to integrate their own monitoring CAN override the defaults should they wish. Example:

```
###############################################################################

  kubecostMetrics:
    exporter:
      enabled: true
      prometheusScrape: false # https://github.com/kubecost/cost-analyzer-helm-chart/pull/1385
      service: # https://github.com/kubecost/cost-analyzer-helm-chart/pull/1385
        annotations:
          prometheus.io/port: "9005"
          prometheus.io/service-scape-slow: "true"
          prometheus.io/kube-state-metrics: "true"

###############################################################################
```

### DIFF
--- a/cost-analyzer/templates/kubecost-metrics-service-template.yaml
+++ b/cost-analyzer/templates/kubecost-metrics-service-template.yaml
@@ -13,7 +13,7 @@ metadata:
 {{ toYaml .Values.kubecostMetrics.exporter.service.annotations | indent 4 }}
 {{- end }}
 {{- if .Values.kubecostMetrics.exporter.prometheusScrape }}
-    prometheus.io/scrape: "{{ .Values.kubecostMetrics.exporter.prometheusScrape }}"
+    prometheus.io/scrape: "true"
     prometheus.io/port: {{ quote .Values.kubecostMetrics.exporter.port | default (quote 9005) }}
 {{- end }}
 {{- end }}
--- a/cost-analyzer/templates/kubecost-metrics-service-template.yaml
+++ b/cost-analyzer/templates/kubecost-metrics-service-template.yaml
@@ -6,10 +6,11 @@ kind: Service
 metadata:
   name: {{ template "kubecost.kubeMetricsName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{- if .Values.kubecostMetrics.exporter.service.annotations }}
   annotations:
-    prometheus.io/scrape: "true"
-    prometheus.io/port: {{ (quote .Values.kubecostMetrics.exporter.port) | default (quote 9005) }}
+{{ toYaml .Values.kubecostMetrics.exporter.service.annotations | indent 4 }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


cost-analyzer/templates/kubecost-metrics-service-template.yaml
```diff
@@ -6,10 +6,17 @@ kind: Service
...
   name: {{ template "kubecost.kubeMetricsName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+{{- if (or .Values.kubecostMetrics.exporter.service.annotations .Values.kubecostMetrics.exporter.prometheusScrape) }}
   annotations:
+{{- if .Values.kubecostMetrics.exporter.service.annotations }}
+{{ toYaml .Values.kubecostMetrics.exporter.service.annotations | indent 4 }}
+{{- end }}
+{{- if .Values.kubecostMetrics.exporter.prometheusScrape }}
```
AjayTripathy, line 15: This feels like it should have a "kindis" check followed up by a default if not exists to ensure that existing users don't need to modify their values.yaml to make this work. I know that the default is set lower in values.yaml but not all users modify values before upgrading, in which case these labels will be dropped silently.

https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L537

AjayTripathy (changes requested):

AjayTripathy:
Hi @rlees85 , very supportive of this change and thank you so much! just one small piece of feedback for backwards compatibility.

nikovacevic (commented):
Seems fine to me. @AjayTripathy want to follow up?

AjayTripathy (approved):

AjayTripathy:
Thanks @rlees85 and sorry for the late response. Merging into nightly for further testing!

PR #1388, opened 2022-04-26
fix cloud usage defaulting

### LINKED ISSUE #1382: No default value for ETL_CLOUD_USAGE_ENABLED variable
**Describe the bug**
When no value gets provided to the values.yaml file, no value gets assigned to the `ETL_CLOUD_USAGE_ENABLED` environment variable in the `cost-analyzer-deployment-template.yaml`. Based on [this doc](https://github.com/kubecost/docs/blob/6290d5818c9a78f92d088fc6d18c06385f82c1e2/cloud-integration.md#cloud-integration-configurations), this value should get set to true.

**To Reproduce**
```cmd
helm template kubecost kubecost/cost-analyzer | grep ETL_CLOUD_USAGE_ENABLED -A1`
            - name : ETL_CLOUD_USAGE_ENABLED
            - name: CLOUD_ASSETS_EXCLUDE_PROVIDER_ID
```

**Expected behavior**
```cmd
helm template kubecost kubecost/cost-analyzer | grep ETL_CLOUD_USAGE_ENABLED -A1`
            - name : ETL_CLOUD_USAGE_ENABLED
              value: "true"
```

gz#1713

Sean-Holcomb:
## What does this PR change?
Fixes default value for ETL_CLOUD_USAGE_ENABLED environment variable. Previous defaults failed because they were depended on values.yaml values being of type boolean, ie having a value, which means that the default was never used. I have removed this type of defaulting and replaced it with a base case else clause.

- 
- 

Users will see ETL_CLOUD_USAGE_ENABLED=true in their env variables. No product impact.

- closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1382
- 

## How was this PR tested?
This PR was tested by using the helm chart with none of the relevant values set, and ensuring that a default value was set.
Screen Shot 2022-04-26 at 12 04 33 PM

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -535,9 +535,11 @@ spec:
             {{- end }}
             - name : ETL_CLOUD_USAGE_ENABLED
             {{- if kindIs "bool" .Values.kubecostModel.etlCloudUsage }}
-              value: {{ (quote .Values.kubecostModel.etlCloudUsage) | default (quote true) }}
+              value: {{ (quote .Values.kubecostModel.etlCloudUsage) }}
             {{- else if kindIs "bool" .Values.kubecostModel.etlCloudAsset }}
-              value: {{ (quote .Values.kubecostModel.etlCloudAsset) | default (quote true) }}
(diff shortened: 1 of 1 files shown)


keithhand (approved):

keithhand:
LGTM

PR #1394, opened 2022-04-27
Remove GOGC tuning from Cost-Model

mbolt35:
## What does this PR change?
* Removes GOGC=60 from cost model as well as legacy go flags for memory release.

* Should speed up cost-model performance on lower CPU core nodes.

## How was this PR tested?
* As part of a performance audit

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -416,10 +416,6 @@ spec:
             - name: REMOTE_WRITE_ENABLED
               value: "true"
             {{- end }}
-            - name: GOGC
-              value: "60"
-            - name: GODEBUG
-              value: "madvdontneed=1"
             {{- if .Values.global.thanos.queryServiceBasicAuthSecretName}}
             - name: MC_BASIC_AUTH_USERNAME
               valueFrom:


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -416,10 +416,6 @@ spec:
             - name: REMOTE_WRITE_ENABLED
               value: "true"
             {{- end }}
-            - name: GOGC
-              value: "60"
-            - name: GODEBUG
-              value: "madvdontneed=1"
```
michaelmdresser, line 422: Are we removing this because its now default behavior in Go? Just want to make sure.

michaelmdresser (approved):

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
mbolt35: GODEBUG, yes. GOGC is 100 by default.

michaelmdresser:
Ref for GOGC change: https://docs.google.com/document/d/1QFnlOhQIBe7hAkNCfNISw5tc7sLv7t6OsKrsrkxFRco/edit#bookmark=id.dcbwcxp40zu2

PR #1402, opened 2022-05-03
White List LoadBalancer Cost

mbolt35:
## What does this PR change?
* Add kubecost_load_balancer_cost to Metric White List

* Fix for missing Load Balancers on hosted deployments which use a service discovery for `kubecost-agent`


dwbrown2:
Just curious, would this metric map to this block? https://github.com/kubecost/cost-analyzer-helm-chart/blob/3a429cbda4ffb8ca2681bdcc5f275d914506ac35/cost-analyzer/values.yaml#L704-L714

If so, worth adding a default reference? We've had that as a feature request.

Apologies if this is different.

AjayTripathy:
> Just curious, would this metric map to this block?
> 
> https://github.com/kubecost/cost-analyzer-helm-chart/blob/3a429cbda4ffb8ca2681bdcc5f275d914506ac35/cost-analyzer/values.yaml#L704-L714
> 
> If so, worth adding a default reference? We've had that as a feature request.
> 
> Apologies if this is different.

It sounds like a good feature request but it isn't configurable here without an additional corresponding codepath to set the pricing in the metrics emitter. Can you file that separately?

AjayTripathy (approved):

dwbrown2:
@mbolt35 good to merge? 

cc @AdamStack18

PR #1409, opened 2022-05-06
Add logging to readme

dramich:
## What does this PR change?
Add logging to readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,3 +55,17 @@ Parameter | Description | Default
 `affinity` | pod affinity | `{}`
 `extraVolumes` | A list of volumes to be added to the pod | `[]`|
 `extraVolumeMounts` | A list of volume mounts to be added to the pod | `[]`
+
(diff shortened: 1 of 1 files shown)


michaelmdresser (approved):

dwbrown2:
Like the output but this doesn't feel like quite the right location for this...

dwbrown2:
Maybe below the config options table?

dramich:
Generally, the explanations and info are above the config options so a user doesn't have to scroll forever to see things, but I can move it

dwbrown2:
Open to other locations, but this is the primary open README for installing Kubecost. Logging detail that close to the install steps feels like we run the risk of distracting / overwhelming a first time user. Maybe that's not the right way of phrasing... let me know if I can articulate better.

dramich:
Makes sense, moved it to the bottom

PR #1415, opened 2022-05-09
Update master to release v1.93.0-rc.1

nealormsbee:
Update master to release v1.93.0-rc.1


nealormsbee:
Something seems really screwy here. Why does GH think that we're coming from version 1.91.0? And there are conflicts? `master` should have `1.93.0-rc.0` for its version after last week

AjayTripathy:
Hm. I agree, let me resolve the conflicts to be correct for now to get the release out then I will take a look.

PR #1418, opened 2022-05-10
add detail to helm install notes

jessegoodier:
## What does this PR change?

Adds additional detail to helm install notes.

No

Helm installation now sets expectations for how long an initial deployment of Kubecost will take.

N/A

## How was this PR tested?

local helm chart

N/A

### DIFF
--- a/cost-analyzer/templates/NOTES.txt
+++ b/cost-analyzer/templates/NOTES.txt
@@ -4,16 +4,16 @@
 {{- $servicePort := .Values.service.port | default 9090 -}}
 Kubecost has been successfully installed.
 
-Please allow 5-10 minutes for Kubecost to gather metrics and report accurate health status in the UI and data to appear.
+Please allow 5-10 minutes for Kubecost to gather metrics.
 
 If you have configured cloud-integrations, it can take up to 48 hours for cost reconciliation to occur.
 
-When using Durable storage (Enterprise Edition) please allow up to 4 hours for data to be collected and the UI to be healthy.
+When using Durable storage (Enterprise Edition), please allow up to 4 hours for data to be collected and the UI to be healthy.
 
 When pods are Ready, you can enable port-forwarding with the following command:
 
-    kubectl port-forward --namespace kubecost deployment/kubecost-cost-analyzer 9090
+    kubectl port-forward --namespace {{ .Release.Namespace }} deployment/{{ template "cost-analyzer.fullname" . }} {{ $servicePort }}
 
-Next, navigate to http://localhost:9090 in a web browser.
+Next, navigate to http://localhost:{{ $servicePort }} in a web browser.
 
 Having installation issues? View our Troubleshooting Guide at http://docs.kubecost.com/troubleshoot-install
--- a/cost-analyzer/templates/NOTES.txt
+++ b/cost-analyzer/templates/NOTES.txt
@@ -2,10 +2,18 @@
 
 --------------------------------------------------
 {{- $servicePort := .Values.service.port | default 9090 -}}
-Kubecost has been successfully installed. When pods are Ready, you can enable port-forwarding with the following command:
-    
-    kubectl port-forward --namespace {{ .Release.Namespace }} deployment/{{ template "cost-analyzer.fullname" . }} {{ $servicePort }}
-    
-Next, navigate to http://localhost:{{ $servicePort }} in a web browser.
+Kubecost has been successfully installed.
+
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/NOTES.txt
```diff
@@ -2,10 +2,18 @@
...
+
+When using Durable storage (Enterprise Edition) please allow up to 4 hours for data to be collected and the UI to be healthy.
+
+When pods are Ready, you can enable port-forwarding with the following command:
+
+    kubectl port-forward --namespace kubecost deployment/kubecost-cost-analyzer 9090
+
+Next, navigate to http://localhost:9090 in a web browser.
 
 Having installation issues? View our Troubleshooting Guide at http://docs.kubecost.com/troubleshoot-install
```
srpomeroy, line 19: We might consider updating this to point to guide.kubecost.com. It's not mandatory as we have redirects in place.
i.e. https://guide.kubecost.com/hc/en-us/articles/4407601830679-Troubleshoot-Install

cost-analyzer/templates/NOTES.txt
```diff
@@ -2,10 +2,18 @@
...
+
+Please allow 5-10 minutes for Kubecost to gather metrics and report accurate health status in the UI and data to appear.
+
+If you have configured cloud-integrations, it can take up to 48 hours for cost reconciliation to occur.
+
+When using Durable storage (Enterprise Edition) please allow up to 4 hours for data to be collected and the UI to be healthy.
+
+When pods are Ready, you can enable port-forwarding with the following command:
+
+    kubectl port-forward --namespace kubecost deployment/kubecost-cost-analyzer 9090
```
michaelmdresser, line 15: I think you should keep the templating `{{ .Release.Namespace }}` etc. so this is correct if users install in a non-standard namespace.

srpomeroy, line 15: We should add back the templated objects here.

cost-analyzer/templates/NOTES.txt
```diff
@@ -2,10 +2,18 @@
...
+
+Next, navigate to http://localhost:9090 in a web browser.
```
srpomeroy, line 17: We should add back the templated objects here.

srpomeroy (changes requested):
Looks like the templated objects were removed from your copy.

cost-analyzer/templates/NOTES.txt
```diff
@@ -2,10 +2,18 @@
...
 --------------------------------------------------
 {{- $servicePort := .Values.service.port | default 9090 -}}
-Kubecost has been successfully installed. When pods are Ready, you can enable port-forwarding with the following command:
-    
-    kubectl port-forward --namespace {{ .Release.Namespace }} deployment/{{ template "cost-analyzer.fullname" . }} {{ $servicePort }}
-    
-Next, navigate to http://localhost:{{ $servicePort }} in a web browser.
+Kubecost has been successfully installed.
+
+Please allow 5-10 minutes for Kubecost to gather metrics and report accurate health status in the UI and data to appear.
```
dwbrown2, line 7: Like the of this PR, but it feel like a lot of text to me. Would love to see if we can potentially be more concise. For example, it feels like this sentence could be shortened.

cost-analyzer/templates/NOTES.txt
```diff
@@ -2,10 +2,18 @@
...
+
+If you have configured cloud-integrations, it can take up to 48 hours for cost reconciliation to occur.
```
dwbrown2, line 9: Is this the best place to share this information? Feels like docs/UI where users are adding a cloud integration would be best.

cost-analyzer/templates/NOTES.txt
dwbrown2: Any benefit to point to guide.kubecost? These urls are far less flexible IMO in that they have the article ID hardcoded in them.

dwbrown2 (approved):

cost-analyzer/templates/NOTES.txt
jessegoodier: I totally missed that, thanks!

cost-analyzer/templates/NOTES.txt
jessegoodier: Agree with web, especially if we ever migrate from zd.

teevans (approved):
LGTM.

srpomeroy (approved):

PR #1442, opened 2022-05-27
Update pod utilization json

zdenekvicar:
## What does this PR change?
- changes in query expressions
  - change `namespace="$namespace"` -> `namespace=~"$namespace"` to allow regex filtering for namespaces
  - change `namespace="$pod"` -> `namespace=~"$pod"` to allow regex filtering for pods
  - add `cluster_id=~\"$cluster\` to allow selecting specific cluster
- changes in variables
  - changed `pod` variable from query to text_field to allow using of regex -> example `cost-analyzer.*` with `cluster` variable set to `All` will allow me to see utilization of cost-analyzer across all clusters in one graph 
    - original query version did not refresh at all when `cluster` variable changed, so it was not useful at all
  - changed `namespace` variable from query to text_field
    - original query version did not refresh at all when `cluster` variable changed, so it was not useful at all
  - changed `cluster` variable query to list `kubecost_cluster_info` metric
    - original query was not able to return all clusters, because this metric has current data only from primary cluster, secondaries are delayed
    - by using `kubecost_cluster_info` I was able to get list of all available clusters immediately, which allows me to use this variable in all graphs in this dashboard
- changes in graph labels
  - added `{{ cluster_id }}` to every graph to show what cluster the given graph is from
    - original version of dashboard does not show this, so when entering grafana from Kubecost UI, I cannot tell what cluster graphs I am looking at, just form the dashboard itself

As also told to @jcharcalla, I am more than happy to jump on a call to discuss my changes & provide live demo of how it looks & works for us.
   
- not that I know of

My issue with previous version of dashboard:

I am able to:
- find resource recommendations for clusterA for cost-model -> open grafana with correct graph for clusterA & cost-model container
- find resource recommendations for clusterB for cost-model -> open grafana with correct graph for clusterB & cost-model container

What I am not able to do though:
- once I am in grafana and watching for cost-model on clusterA, I cannot switch to see clusterB right from grafana UI ... I have to go back to KubecostUI and open specific graph from there

Currently proposed version of dashboard allows this.

- Kubecost ticket #1872, working on this with @jcharcalla 

## How was this PR tested?
yes, on our instance of Kubecost

no, I did not


AdamStack18:
@zdenekvicar Thanks for this contribution. We will get this reviewed by tomorrow afternoon.

cost-analyzer/pod-utilization.json
```diff
@@ -178,14 +178,14 @@
       ],
       "targets": [
         {
-          "expr": "sum(\n  avg_over_time(container_memory_allocation_bytes{namespace=\"$namespace\", pod=\"$pod\", container!=\"POD\"}[$__range])\n) by (container,node)",
+          "expr": "sum(\n  avg_over_time(container_memory_allocation_bytes{namespace=~\"$namespace\", pod=~\"$pod\", container!=\"POD\", cluster_id=~\"$cluster\"}[$__range])\n) by (container,node, cluster_id)",
```
nikovacevic, line 181: `cluster_id` is not always used to identify clusters -- that is a configurable field

nikovacevic (commented):
Thanks for the contribution @zdenekvicar. Glad to hear you were able to work with @jcharcalla to get what you need out of this dashboard.

The way that `cluster_id` is used here will not work universally for federated Kubecost users, as that label is configurable. As such, and as I'm not certain that this is in alignment with our goals for Kubecost-published Grafana dashboards, I don't think I can approve this. But tagging in @AjayTripathy to see if he can shed more light.

cost-analyzer/pod-utilization.json
AjayTripathy: Thanks for calling that out... @zdenekvicar can we look to make that "cluster_id" configurable? maybe sub in $clusterLabel or the like?

AjayTripathy: After that I'd be supportive of approving and merging.

zdenekvicar: I am all in to change that so it can use whatever label is selected, but I am not sure if I know how :/ so any help is definitely welcome here (maybe I will figure this out later).

I see the label is controlled in here `prometheus.server.global.external_labels` in values file, but I am not really sure how to propagate whatever cluster-label is defined here, to the json dashboard source

AjayTripathy: I think it's ok to allow the user to specify and not necessarily propagate? Does that make this easier?

zdenekvicar:
I am sorry, but I am currently totally out of time to work on this ... I will close this PR for now.

PR #1449, opened 2022-06-01
Update helm chart to default to ETL file store

mbolt35:
## What does this PR change?
* Enables ETL FileStore (Virtual memory solution backed by disk) by default. 

* Users which have enabled a disk based backup for MemoryStore will seamlessly change to FileStore (unless it is explicitly disabled). 

## How was this PR tested?
```
helm template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml  > k.yaml
```

```
...
            - name: ETL_DAILY_STORE_DURATION_DAYS
              value: "91"
            - name: ETL_HOURLY_STORE_DURATION_HOURS
              value: "49"
            - name: ETL_FILE_STORE_ENABLED
              value: "true"
            - name: ETL_ASSET_RECONCILIATION_ENABLED
              value: "true"
            - name: ETL_USE_UNBLENDED_COST
              value: "false"
...
```

In the works

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -560,7 +560,7 @@ spec:
             - name: ETL_HOURLY_STORE_DURATION_HOURS
               value: {{ (quote .Values.kubecostModel.etlHourlyStoreDurationHours) | default (quote 49) }}
             - name: ETL_FILE_STORE_ENABLED
-              value: {{ (quote .Values.kubecostModel.etlFileStoreEnabled) | default (quote false) }}
+              value: {{ (quote .Values.kubecostModel.etlFileStoreEnabled) | default (quote true) }}
             - name: ETL_ASSET_RECONCILIATION_ENABLED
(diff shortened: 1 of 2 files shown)


dwbrown2:
Should we add to default values.yaml?

Any risks that we should keep in mind with this?

mbolt35:
> Should we add to default values.yaml?

We could add the default to `values.yaml`

> Any risks that we should keep in mind with this?

There shouldn't be any risk in this change, as the the memory store + file backup are 100% compatible/interchangeable with the filestore implementation. Additionally, MemoryStore + Bucket backup is compatible with FileStore + Bucket backup. 

The _only_ thing that I'm unsure of is the use of the node storage for mmapping files back into virtual memory. While we will cull the local file system files to the defined ETL coverage, it is possible that we could run into problems where we're unable to store an ETL file on a full disk. However, this scenario would only occur _if_ the users was on a pure memory store option with no backing disk storage. In which case, we case, disabling the file store will put them right back to where they were.

AjayTripathy (approved):

Issue #1322, opened 2022-03-22
Key duplication in kubecost-metrics-deployment-template.yaml


mortenmj:
**Describe the bug**

Similar to #1150 there is a key duplication where the app label applied from kubeMetricsName conflicts with commonLabels.
This causes kustomize to fail.

**To Reproduce**

Run yamllint against helm manifests of templates/kubecost-metrics-deployment-template.yaml
grep key-duplication

gz#1548

jcharcalla:
@mortenmj Can you provide details on what version you are using  and the error from kustomize?

I was unable to reproduce this using with v1.91.2 and `yamllint 1.26.3`. 
```$ yamllint cost-analyzer-helm-chart-1.91.2/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml | grep key-duplication```

It may also be helpful if you can provide a copy of the `kubecost-metrics-deployment-template.yaml` file you are using.

jcharcalla:
Thank you for your submission, it has been merged and should be included in an upcoming release.

PR #1457, opened 2022-06-03
Remove duplicate label configurations on service monitor

pcfens:
## What does this PR change?

It removes duplicate labels that were being attached to the ServiceMonitor object. It's a similar issue to #1322, but in a different location. The removed set of labels aren't used anywhere other than this one template that I removed them from either.

Tools such as the flux helm controller can install the chart now when `serviceMonitor.enabled` is set to true

## How was this PR tested?

The generated yaml was applied to a cluster and there were a reduction in errors from yamllint.

No. None was necessary.


AdamStack18:
@pcfens, Thanks for this contribution will get this PR reviewed next week.

AjayTripathy (approved):

AjayTripathy:
Thanks @pcfens !!

AjayTripathy:
Whoops, missed that. Ran workflow, merging now.

PR #1489, opened 2022-06-17
Master

AjayTripathy:
Merge master into develop to catch up versions


michaelmdresser:
I'm not sure that we should be handling this in this way. Perhaps we could lock `develop` to a version that expresses that its develop, and then actual releases override the version? I think Dan's PR adding version+commit information is a good example: https://github.com/kubecost/kubecost-cost-model/pull/835 -- defaulting to `HEAD` (for commit) and `dev` (for version).

michaelmdresser:
Going to close for the time being. Feel free to reopen if you need.

PR #1505, opened 2022-06-28
Drop the default clusterID and force adding one

### LINKED ISSUE #1504: Cluster-controller chart does not repsect clusterIDConfigmap
In other locations we have an if check around using clusterID, currently that does not happen and cluster-controller assumes the value exists. If the value is not provided then an empty envvar will get set in the container. 

https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/kubecost-cluster-controller-template.yaml#L210

vs. 

https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L664-L674

dramich:
## What does this PR change?
Currently the helm chart always defaults to adding a cluster_id of `cluster-one` which some users don't realize and end up with many clusters with the same ID. This can cause issues when attempting to do federated things (future etl work). 

Drop the default cluster ID, update the logic to allow the value to be missing (if this value was missing previously, helm would throw a nil pointer) and add a `required` call to ensure either `cluster_id` or `clusterIDConfigmap` is set.

If neither value is set a user will see an error:
```
Error: execution error at (cost-analyzer/templates/kubecost-cluster-controller-template.yaml:216:23): A cluster_id OR clusterIDConfigmap is required!
```

Setting either cluster_id OR clusterIDConfigmap is now required when using the helm chart. There is no default value.

## How was this PR tested?
Running `helm template` a lot. 

This would also fix https://github.com/kubecost/cost-analyzer-helm-chart/issues/1504

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Run helm template
         working-directory: ./cost-analyzer
         # Template the chart out for kubeval usage and also validates that it can 'build'
-        run: helm template . --set global.prometheus.enabled=false --set global.grafana.enabled=false > full.yaml
+        run: helm template . --set global.prometheus.enabled=false --set global.grafana.enabled=false --set prometheus.server.global.external_labels.cluster_id=cluster-1 > full.yaml
 
       - name: Kubeval
         uses: instrumenta/kubeval-action@master
@@ -57,7 +57,7 @@ jobs:
         working-directory: ./cost-analyzer
         run: |
           kubectl create ns kubecost
-          helm install kubecost . --namespace kubecost --set global.prometheus.enabled=false --set global.grafana.enabled=false
+          helm install kubecost . --namespace kubecost --set global.prometheus.enabled=false --set global.grafana.enabled=false --set prometheus.server.global.external_labels.cluster_id=cluster-1
 
       - name: Cleanup k3s
         if: always()
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -661,15 +661,14 @@ spec:
             {{- end }}
             {{- end }}
             {{- end }}
-            {{- if and (.Values.prometheus.server.global.external_labels.cluster_id) (not .Values.prometheus.server.clusterIDConfigmap) }}
+            {{- if and (hasKey .Values.prometheus.server.global.external_labels "cluster_id") (not .Values.prometheus.server.clusterIDConfigmap) }}
             - name: CLUSTER_ID
               value: {{ .Values.prometheus.server.global.external_labels.cluster_id }}
-            {{- end }}
-            {{- if .Values.prometheus.server.clusterIDConfigmap }}
+            {{- else }}
             - name: CLUSTER_ID
               valueFrom:
                 configMapKeyRef:
-                  name: {{ .Values.prometheus.server.clusterIDConfigmap }}
+                  name: {{ required "A cluster_id OR clusterIDConfigmap is required!" .Values.prometheus.server.clusterIDConfigmap }}
                   key: CLUSTER_ID
             {{- end }}
             {{- if .Values.remoteWrite.postgres.installLocal }}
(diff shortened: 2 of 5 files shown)


jessegoodier:
This is awesome.

Is it possible to use this same value for kubecostProductConfigs.clusterName to force them to be the same?

Or is that a bad idea?

dramich:
> This is awesome.
> 
> Is it possible to use this same value for kubecostProductConfigs.clusterName to force them to be the same?
> 
> Or is that a bad idea?

It could be used to populate the clusterName if one isn't provided by the user but those two things are different and not required to be the same. clusterID should only be set once otherwise it can/will cause issues on rollup but clusterName is a display name and can technically be changed at will. That change is outside the scope of this pr but I am sure we would accept a pr for it 😄

mbolt35:
ClusterName will automatically default to the `cluster_id` if a cluster name doesn't exist (at least within the product itself).

mbolt35:
@AjayTripathy I am fully supportive of requiring the user to input a cluster_id here, but I did want to loop you in. Existing customers running `helm upgrade` will just continue to use the old value, but new customers will be required to set this value during install (@dramich Correct me if that is wrong).

jessegoodier:
If they were forced to be the same, we could eventually remove clusterName altogether?

It would simplify the UI and the confusion caused. It is messed up on SaaS (with red arrow) and regular (clustername/cluster_id).
image
image

Adam-Stack-PM:
@alexkubecost Very cool seeing a PR by you.  Nice work.

mbolt35:
> If they were forced to be the same, we could eventually remove clusterName altogether?
> 
> It would simplify the UI and the confusion caused. It is messed up on SaaS (with red arrow) and regular (clustername/cluster_id). image image

@jessegoodier Right now, this isn't really an option. They will continue to be different values, one of which is modifiable (cluster name), the other is not (cluster id). 

> It would simplify the UI and the confusion caused. It is messed up on SaaS (with red arrow) and regular (clustername/cluster_id).

I believe it's possible that some users do not have human readable cluster_ids (especially if they were using Thanos/Cortex prior to installing Kubecost), in which case, the cluster name offers a significant UI advantage for identifying the cluster. I'm also not spotting what's "messed up" on the hosted? Is it the display of both the cluster id and name? Seems like a pretty reasonable request for the frontend team to only display one if they're identical. 

I just want to be very clear that what @dramich has implemented is very relevant to the technical implementation, and isn't simply a UX/UI "nice to have." The issues you've added to this PR are not relevant to this change, and should probably be a separate issue.

jessegoodier:
Agreed, different issue. @mbolt35 
This is a great PR and I apologize for the side track.

Would love your thoughts on the FE change, will tag you when I create the issue.

mbolt35 (approved):
Approving this. Thanks @dramich <3

AjayTripathy:
Just double checking @mbolt35 -- no need to update any of our install instructions right? Our standard helm install continues to work?

(helm install kubecost kubecost/cost-analyzer --namespace kubecost --set kubecostToken="")

AjayTripathy:
Hey @mbolt35 this is making our integration tests fail which leads me to believe this _will_ break the standard install. Do we need to walk this back or update our install instructions? What do you suggest?

Run helm upgrade kubecost -n kubecost \
Error: UPGRADE FAILED: execution error at (cost-analyzer/templates/cost-analyzer-deployment-template.yaml:701:27): A cluster_id OR clusterIDConfigmap is required!

dwbrown2:
We have many other companies referencing our install steps. Most of them will not update, or at least not in a timely manor. I would try to steer us away from anything breaking this unless we view it as critical.

mbolt35:
@mbolt35 @AjayTripathy I have no problem if you guys want to roll this back. My argument against this is: when it comes to conflicting cluster identifiers, this problem is MUCH more difficult to identify when it's wrong. Whereas, the messaging here is simple and clear: "hey, this upgrade failed because you didn't have a required field that we expect" -- It's clear and it saves them in the future of running into a complete disaster with conflicting cluster ids in the future. 

I do think we should likely update install instructions, or release notes/migrations. Let me know if you want to roll it back and I'll gladly do so as long as it's clear what the risk is.

AjayTripathy:
@mbolt35 sounds like we need to roll this back since there's a change to the default instructions. I'm not opposed at all to changing the default instructions but we would need sign off from other folks as well. Let's roll back before 1.97 and look to add this and update install instructions in 1.98?

PR #1509, opened 2022-06-30
Add OIDC config params in Values

kaelanspatel:
## What does this PR change?
Adds config values for OIDC, as well as the template(s) to handle them. Creates a secret that stores the OIDC client secret key, as well as a configmap used for the OIDC integration.

Further information on testing/behavior in linked KCCM PR

- 
- 

- 
-

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
@@ -1,0 +1,21 @@
+{{- if .Values.oidc }}
+{{- if .Values.oidc.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "cost-analyzer.fullname" . }}-oidc
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+data:
+{{- $root := . }}
+    oidc.json: |-
+      {
+        "enabled" : {{ .Values.oidc.enabled }},
+        "clientID" : "{{ .Values.oidc.clientID }}",
+        "secretName" : "{{ .Values.oidc.secretName }}",
+        "authURL" : "{{ .Values.oidc.authURL }}",
+        "loginRedirectURL" : "{{ .Values.oidc.loginRedirectURL }}",
+        "discoveryURL" : "{{ .Values.oidc.discoveryURL }}"
+      }
+{{- end -}}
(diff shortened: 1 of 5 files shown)


kaelanspatel:
P.S. I know there are keys in the commit history, these have been revoked and are solely used for test accounts with no existing connections.

cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
```diff
@@ -0,0 +1,13 @@
...
+{{- if .Values.oidc.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "cost-analyzer.fullname" . }}-oidc
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+data:
+{{- $root := . }}
+    oidc.json: '{{ toJson .Values.oidc }}'
```
dramich, line 11: Won't this put the `clientSecret` into the JSON?

cost-analyzer/templates/kubecost-oidc-secret-template.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.oidc }}
+{{- if .Values.oidc.secretName }}
```
dramich, line 2: Instead of `if` checks here and silently failing to create the secret, we can use  the `required` function on fields that are required when OIDC is enabled https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function

dramich:
> P.S. I know there are keys in the commit history, these have been revoked and are solely used for test accounts with no existing connections.

Squash your commits and force push.

cost-analyzer/templates/kubecost-oidc-secret-template.yaml
kaelanspatel: That's cool, didn't know you could do that.

In this case though I think it's fine, technically you can create the secret yourself, so if there's an issue there it'll fail nonsilently elsewhere, either within the cost-model or in the volume mount.

cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
kaelanspatel: Yep... fixing to only contain the relevant fields.

dramich (approved):

PR #1565, opened 2022-07-19
Update default datasource filename to match hardcoded filename value

gracedo:
## What does this PR change?
I saw grafana datasources values was modified in this PR https://github.com/kubecost/cost-analyzer-helm-chart/pull/1330 but this can break installation if `.Values.grafana.sidecar.datasources` values are also set, due to https://github.com/kubecost/cost-analyzer-helm-chart/blob/bc2321c725cc4b49fca229ae8d121080900777f7/cost-analyzer/templates/grafana-datasource-template.yaml#L24 where filename is defaulted to `datasource.yaml` instead of `datasources.yaml`. User could also fix by setting https://github.com/kubecost/cost-analyzer-helm-chart/blob/ebe7e088debecd23f90e6dd75b425828901a246c/cost-analyzer/values.yaml#L691 to `dataSourceFilename: datasources.yaml`

- 
- 

- 
-


gracedo:
Hmm actually, that wasn't the issue (although this change did get the installation working but was not really a fix), it was that my original values before upgrading to the version with the datasource changes, was setting `.grafana.sidecar.datasources.defaultDatasourceEnabled` to `true` which conflicted with the new datasource entries which was also setting a default datasource.
https://github.com/kubecost/cost-analyzer-helm-chart/blob/ebe7e088debecd23f90e6dd75b425828901a246c/cost-analyzer/templates/grafana-datasource-template.yaml#L30-L34 Setting this to `false` fixed the errors
```Failed to start grafana. error: Datasource provisioning error: datasource.yaml config is invalid. Only one datasource per organization can be marked as default```

AjayTripathy:
@gracedo thanks so much for the PR and investigation. Let me know if there's anything else I can help with here.

PR #1568, opened 2022-07-20
Leader/Follower Enable

mbolt35:
## What does this PR change?
* Adds the ability to enable leader/follower election given the following requirements are met:
  * Replicas are set to a value greater than 1
  * ETL FileStore is Enabled (it is enabled by default)
  * ETL Bucket Backup is Configured

* Requires Leader/Follower PRs in opencost and closed source repository 

* Allows enterprise users to enable a leader/follower election for running kubecost in a more highly available mode which provides read-only ETL on followers and read/write ETL on the leader. Requirements:
  * Replicas are set to a value greater than 1
  * ETL FileStore is Enabled (it is enabled by default)
  * ETL Bucket Backup is Configured

## How was this PR tested?
If the `kubecostDeployment.leaderFollower.enabled=true` flag is set, the chart will _require_ that specific configuration options are set and include valid values. 

### All Requirements:
```bash
helm3 template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml \
	--set kubecostDeployment.leaderFollower.enabled=true \ 
	--set kubecostDeployment.replicas=5 \
	--set kubecostModel.etlBucketConfigSecret=kubecost-bucket-secret > temp.yaml
```

*Result*:
```
- name: LEADER_FOLLOWER_ENABLED
  value: "true"
```

### Replicas = 1, Expected Failure:
```bash
helm3 template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml \
	--set kubecostDeployment.leaderFollower.enabled=true \
	--set kubecostDeployment.replicas=1 \
	--set kubecostModel.etlBucketConfigSecret=kubecost-bucket-secret > temp.yaml
```
*Result*:
```
Error: template: cost-analyzer/templates/cost-analyzer-deployment-template.yaml:736:16: executing "cost-analyzer/templates/cost-analyzer-deployment-template.yaml" at <fail "Leader/Follower should be used with kubecostDeployment.replicas > 1">: error calling fail: Leader/Follower should be used with kubecostDeployment.replicas > 1
```

### No Bucket Backup, Expected Failure:
```bash
helm3 template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml \
	--set kubecostDeployment.leaderFollower.enabled=true \
	--set kubecostDeployment.replicas=5 > temp.yaml
```
*Result*:
```
Error: template: cost-analyzer/templates/cost-analyzer-deployment-template.yaml:731:16: executing "cost-analyzer/templates/cost-analyzer-deployment-template.yaml" at <fail "Leader/Follower requires kubecostModel.etlBucketConfigSecret be valid.">: error calling fail: Leader/Follower requires kubecostModel.etlBucketConfigSecret be valid.
```

###  Explicitly Disabling ETL File Store, Expected Failure:
```bash
helm3 template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml \
	--set kubecostModel.etlFileStoreEnabled=false \
	--set kubecostModel.etlBucketConfigSecret=kubecost-bucket-secret \
	--set kubecostDeployment.leaderFollower.enabled=true \
	--set kubecostDeployment.replicas=5 > temp.yaml
```
*Result*:
```
Error: template: cost-analyzer/templates/cost-analyzer-deployment-template.yaml:728:16: executing "cost-analyzer/templates/cost-analyzer-deployment-template.yaml" at <fail "Leader/Follower requires kubecostModel.etlFileStoreEnabled be true.">: error calling fail: Leader/Follower requires kubecostModel.etlFileStoreEnabled be true.
```

### ClusterRole Update for coordination K8s API:
```bash
helm3 template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml \
	--set kubecostDeployment.leaderFollower.enabled=true \ 
	--set kubecostDeployment.replicas=5 \
	--set kubecostModel.etlBucketConfigSecret=kubecost-bucket-secret > temp.yaml
```
*Result*:
```
- apiGroups:
  - coordination.k8s.io
resources:
  - leases
verbs:
  - '*'
```

**Disabled Leader/Follower, and Observed no coordination API group role added, no requirements applied.**

**Same tests were run with `values.yaml` updates only, and the same results were observed**


cost-analyzer/templates/_helpers.tpl
```diff
@@ -217,3 +217,33 @@ The implied use case is {{ template "cost-analyzer.filterEnabled" .Values }}
 {{- template "cost-analyzer.filter" (dict "v" . "r" $result) }}
 {{- $result | toJson | b64enc }}
 {{- end -}}
+
+{{/*
+This template runs the full check for leader/follower requirements in order to determine
+whether it should be configured. This template will return true if it's enabled and all 
+requirements are met. 
+*/}}
+{{- define "cost-analyzer.leaderFollowerEnabled" }}
```
mbolt35, line 226: Helm templating, hooray! 👎

cost-analyzer/templates/_helpers.tpl
```diff
@@ -217,3 +217,33 @@ The implied use case is {{ template "cost-analyzer.filterEnabled" .Values }}
...
+            {{- else }}
+                {{- "" }}
+            {{- end }}
+        {{- else }}
+            {{- "" }}
+        {{- end }}
+    {{- else }}
+        {{- "" }}
+    {{- end }}
+{{- end }}
```
mbolt35, line 249: Helm templating, hooray! 👎

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -713,10 +713,50 @@ spec:
...
+            {{- end }}
             - name: RELEASE_NAME
               value: {{ .Release.Name }}
             - name: KUBECOST_NAMESPACE
               value: {{ .Release.Namespace }}
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.name
```
mbolt35, line 759: Works with: https://github.com/opencost/opencost/pull/1294

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -713,10 +713,50 @@ spec:
...
+            */}}
+            {{- $result := include "cost-analyzer.leaderFollowerEnabled" . }}
+            {{- if not $result }}
+            {{- fail (quote $result) }}
+            {{- end }}
+            - name: LEADER_FOLLOWER_ENABLED
+              value: "true"
+            {{- end }}
+            {{- end }}
+            {{- end }}
```
LaikaN57, line 750: This seems redundant. Can we just add the failures to the defined template `cost-analyzer.leaderFollowerEnabled` and then just check the first value for "true" otherwise re-fail the error up?

(I am still a n00b at templating so sorry if my understanding it wrong.)

cost-analyzer/templates/_helpers.tpl
Sean-Holcomb: I started typing out something I though would look marginally nicer, and it did not look better at all.

Sean-Holcomb (approved):
Looks good, I would like to hear the answer to @LaikaN57 question too

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
mbolt35: We could go that route, but we would lose the details regarding what caused the failure. The only reason the `"cost-analyzer.leaderFollowerEnabled"` is useful is as an all-in-one check, but since we don't control ordering for how the templates are generated, we need at least one place that hard-fails with good error messaging, and the other checks can just use the template.

cost-analyzer/templates/_helpers.tpl
mbolt35: I would love to think that there is a better way to do this, but each incremental "improvement" I made to the above would break a use-case.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
mbolt35: You may be correct about the template -- just have the hard failures (rather than return a true/false value). And you're right about the redundancy here with respect to calling the template in the immediate checks... It's redundant granted that both the template and my check in the cost-analyzer-deployment are logically equivalent (which I was sort of using as a unit test). We could technically remove the template call, but it's not necessarily hurting anything by leaving it (in fact, it helps me sleep at night).

LaikaN57: ah, so `cost-analyzer.leaderFollowerEnabled` gets computed once? I thought we could just do something like the following to "re-raise".

```
{{- $result, $err := include "cost-analyzer.leaderFollowerEnabled" . }}
{{- if $err }}
{{- fail (quote $err) }}
```

Again, sorry for the n00b-ness on this, but it sounds like you have thought it through. Go forth!

mbolt35: > (I am still a n00b at templating so sorry if my understanding it wrong.)
The helm templating is truly the most counter-intuitive tech I have ever used. However, I think you're understanding correctly -- In general though, I have found that using the long-winded approach (which includes some possible redundancy) is good way to ensure consistent output across "random" template ordering. The biggest issue IMO is the yaml types (especially bool) converting into strings where `"false"` is `true` and `""` is `false`.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -55,6 +55,15 @@ rules:
...
       - watch
+{{- $isLeaderFollowerEnabled := include "cost-analyzer.leaderFollowerEnabled" . }}
+{{- if $isLeaderFollowerEnabled }}
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - '*'
+{{- end }}
```
michaelmdresser, line 66: Way late to the party here. Should we only have these permissions in the Kubecost install namespace, rather than the whole cluster? My skim of the documentation suggests that leases are namespaced.

PR #1587, opened 2022-07-25
Prevent duplicate "app" labels when enabling cluster controller

### LINKED ISSUE #1542: Enabled cluster controller generates duplicate app labels
**Describe the bug**
If I enable cluster controller, helm will render it's resources with two 'app' labels, which make fluxcd error out (like in issues https://github.com/kubecost/cost-analyzer-helm-chart/issues/1322 and https://github.com/kubecost/cost-analyzer-helm-chart/issues/1150)

**To Reproduce**
1. Run `helm template test kubecost/cost-analyzer --set clusterController.enabled=true`
2. Inspect output and see duplicate labels:
```
# Source: cost-analyzer/templates/kubecost-cluster-controller-template.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test-cluster-controller
  labels:
[...]
    app: cost-analyzer
    app: test-cluster-controller
```
4. See error in fluxcd:
```
 Message:               Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
  line 12: mapping key "app" already defined at line 11
```

**Expected behavior**
All labels should be unique

chart version 1.94.3

gz#2158<br><br><i>(related to <a href='https://kubecost.zendesk.com/agent/tickets/2158'>Zendesk ticket #2158</a>)</i>

nealormsbee:
## What does this PR change?

Stop using `commonLabels` in some parts of the `clusterController` template that were leading to entries with duplicate `app: ` label metadata. Also ensure that `app: ` and `app.kubernetes.io/name: ` labels match for said entries.

Some definition in the template continue to use `cost-analyzer` for their app name, as they did before. Definitions which had conflicting labels `app: cost-analyzer` and `app: Release.name-cluster-controller` are consolidated on the latter name. I'm unsure whether all definitions in the file should be changed to consolidate on one name or another; please advise.

No.

- Fixes a bug where enabling `clusterController` produced a Helm template with non-unique label metadata

- Closes https://github.com/kubecost/cost-analyzer-helm-chart/issues/1542

## How was this PR tested?
By running

```
helm template test ./cost-analyzer --set clusterController.enabled=true > out.yaml
```

locally and observing that duplicate `app: ` labels no longer appear for clusterContorller items.

No

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -5,8 +5,11 @@ kind: ServiceAccount
 metadata:
   name: {{ template "kubecost.clusterControllerName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
     app: {{ template "kubecost.clusterControllerName" . }}
+    app.kubernetes.io/name: {{ template "kubecost.clusterControllerName" . }}
+    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
 ---
 #
 # NOTE: 
@@ -18,8 +21,11 @@ kind: ClusterRole
 metadata:
   name: {{ template "kubecost.clusterControllerName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
     app: {{ template "kubecost.clusterControllerName" . }}
+    app.kubernetes.io/name: {{ template "kubecost.clusterControllerName" . }}
+    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
 rules:
   - apiGroups:
       - kubecost.com
@@ -155,8 +161,11 @@ kind: ClusterRoleBinding
 metadata:
   name: {{ template "kubecost.clusterControllerName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
     app: {{ template "kubecost.clusterControllerName" . }}
+    app.kubernetes.io/name: {{ template "kubecost.clusterControllerName" . }}
+    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -5,11 +5,7 @@ kind: ServiceAccount
 metadata:
   name: {{ template "kubecost.clusterControllerName" . }}
   labels:
-    app: {{ template "kubecost.clusterControllerName" . }}
-    app.kubernetes.io/name: {{ template "kubecost.clusterControllerName" . }}
-    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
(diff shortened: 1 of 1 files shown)


LaikaN57 (changes requested):
## Primary Requests

Let's actually just replace `cost-analyzer.commonLabels` with `kubecost.chartLabels`. I do not see where we are manually setting any other similar labels directly instead of using one of these helpers.

### Alternative

Alternatively, let's update the `app` label in `cost-analyzer.commonLabels` I think and remove it here. This will keep everything in-sync.

### Secondary Requests

Additionally, we should fix the parts of this chart that use this method of overriding (scope creep, but I think still related enough):

```
{{ unset (include "cost-analyzer.commonLabels" . | fromYaml) "app" | toYaml | nindent 4 }}
```

In:

- cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml
- cost-analyzer/templates/kubecost-metrics-deployment-template.yaml

LaikaN57:
Also, please update PR title.

nealormsbee:
Hi @LaikaN57, thanks for the feedback. PR title is updated. Re: other suggestions:

>Let's actually just replace cost-analyzer.commonLabels with kubecost.chartLabels. I do not see where we are manually setting any other similar labels directly instead of using one of these helpers.

I believe that switching to `kubcost.chartLabels` lands us in a similar situation. `chartLabels` still sets, e.g., `app.kubernetes.io/name` to `cost-analyzer.name`. If I'm understanding the intention of the cluster-controller chart correctly, we want that label to have its own, distinct name. So we're back to needing to override a specific property of the helper.

I suppose one option is to make this change, and remove `app.kubernetes.io/name` from the `chartLabels` helper, which could be done because this helper isn't actually used anywhere currently. But it does seem like app name is a basic chart label that you'd want included in such a helper.

>Alternatively, let's update the app label in cost-analyzer.commonLabels I think and remove it here. This will keep everything in-sync.

Again, I think the core problem with this approach is that the cluster-controller app name is intended to be distinct from the primary chart's app name, so in that sense we don't *want* them to be synced. The commonLabels helper should have an app name label of `cost-analyzer.name`, and the cluster-controller should have an app name of `kubecost.clusterControllerName`.


I think visiting these concepts to clean up our helm chart templating is a reasonable ask, but should perhaps be its own issue. My intention with this PR is to unblock some folks who are currently totally unable to run Kubecost on Flux because of label collisions, rather than attempt to re-architect our labeling templates.

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -5,8 +5,11 @@ kind: ServiceAccount
 metadata:
   name: {{ template "kubecost.clusterControllerName" . }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
     app: {{ template "kubecost.clusterControllerName" . }}
+    app.kubernetes.io/name: {{ template "kubecost.clusterControllerName" . }}
+    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
```
michaelmdresser, line 12: For the sake of parity with the rest of this file, in this PR I suggest that we stick with just
```
{{ include "cost-analyzer.commonLabels" . | nindent 4 }}
```
(just deleting the specific `app` label, not replacing the common labels)

This is what the rest of this file uses (the Deployment [1], etc.) so we'll have a uniform experience, even if it isn't exactly correct.

I think it's worth changing these to be cluster controller-specific at some point, but we have quite a bit going on right now so I won't call it a priority.

[1] https://github.com/kubecost/cost-analyzer-helm-chart/blob/296ac4ef244168443e085c3e179511b8ed5409ea/cost-analyzer/templates/kubecost-cluster-controller-template.yaml#L170-L175

michaelmdresser (commented):
Hi @LaikaN57, we appreciate the suggestions. Unfortunately, they fall a bit outside the scope of this fix and the time we have allotted to it right now. We would welcome a discussion and/or contribution about overall labeling strategy in a separate issue or PR, respectively, if you'd like to contribute! Neal has already put forward some great ideas.

michaelmdresser (approved):

PR #1588, opened 2022-07-25
Needs-follow-up fix

nealormsbee:
## What does this PR change?
The GH action used to apply the `needs-follow-up` label to issues is applying the label to issues when Kubecost org members comment, as well as community members. We actually want to *remove* the label when members from within the org comment.

The issues stems from checking for an HTTP 204 response from the organization membership API. In testing, this was always returned for members. In practice, we have been receiving 3XX responses from the API (redirects), which has lead the Action to believe that members are non-members.

This changes the `grep` for `HTTP/2 204` to a `grep` for `HTTP/2 [235]`, which *checks for the non-presence of a 4XX response*.

No

It doesn't -- it just makes this issue housekeeping tool actually useful.

N/A

## How was this PR tested?
Locally with the `act` tool

No

### DIFF
--- a/.github/workflows/label-comments.yml
+++ b/.github/workflows/label-comments.yml
@@ -18,7 +18,7 @@ jobs:
       - name: "Check for non-4XX response"
         id: membership
         run: |
-          echo '${{ steps.response.outputs.MEMBER_RESPONSE }}' && echo "::set-output name=IS_MEMBER::$(grep -E 'HTTP/2 [235]' <<< '${{ steps.response.outputs.MEMBER_RESPONSE }}')"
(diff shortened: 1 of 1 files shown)


nealormsbee:
@Adam-Stack-PM what was the label we were going to use for non-release items? Definitely applies here.

Adam-Stack-PM:
> @Adam-Stack-PM what was the label we were going to use for non-release items? Definitely applies here.

Lets go with "outside-release"

nealormsbee:
Would appreciate a review here when we're over the release hump today. the `needs-follow-up` label won't actually be effective until then (current implementation is labeling *everything* with needs-follow-up)

michaelmdresser (approved):

PR #1650, opened 2022-08-21
Add value option - 'affinity' for 'prometheus/node-exporter-daemonset'

kc-adawson:
## What does this PR change?

Adds option to set "Affinity" for NodeExporter DaemonSet via the values file. 

.Values.nodeExporter.affinity | indent 8

No

Allows users to set the "affinity" for the daemonset via the values file.

## How was this PR tested?

Bolt tested the generation of templates with this code and it properly added affinity to the daemonset.

Documentation is here - We tell the user to taint the nodes, but this would allow users to set this affinity using the Kubecost values. https://guide.kubecost.com/hc/en-us/articles/6152374933655-Windows-Node-Support

### DIFF
--- a/cost-analyzer/charts/prometheus/templates/node-exporter-daemonset.yaml
+++ b/cost-analyzer/charts/prometheus/templates/node-exporter-daemonset.yaml
@@ -30,6 +30,10 @@ spec:
 {{ toYaml .Values.nodeExporter.pod.labels | indent 8 }}
 {{- end }}
     spec:
+{{- if .Values.nodeExporter.affinity }}
+      affinity:
+{{ toYaml .Values.nodeExporter.affinity | indent 8 }}
+{{- end }}
       serviceAccountName: {{ template "prometheus.serviceAccountName.nodeExporter" . }}
 {{- if .Values.nodeExporter.priorityClassName }}
       priorityClassName: "{{ .Values.nodeExporter.priorityClassName }}"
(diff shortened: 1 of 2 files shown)


michaelmdresser:
@kc-adawson why is this targeting `releasebot/master-release-v1.96.0` as the base branch? 

By default, PRs should be targeted at `develop` (or in other repos, whatever the default branch is). If you think a PR is worthy of a patch to a release, consult with the team to make sure it is, then follow this process: https://github.com/kubecost/release-scripts#if-you-need-to-patch-the-release-branch

mbolt35:
@kc-adawson Yeah, this changeset is really hard to read -- I would suggest closing and reopening with a PR against develop

kc-adawson:
> @kc-adawson Yeah, this changeset is really hard to read -- I would suggest closing and reopening with a PR against develop

Hey Bolt, Michael is meeting with Jesse and I today to talk through the proper way to do this for SEs. 

If you notice, I actually switched it to develop which is the reason it is so messy, since develop is based off an older version it seems.. If I switch it back to v.1.96.0 its clean with only the 2 changes I made.

mbolt35:
@kc-adawson Yeah, the way you can "reset it" is by closing the PR and just re-opening it targetting the same branch. The "PR History" is tied to the github not git (I'm nearly certain)

michaelmdresser:
To close the loop here: `releasebot/master-release-v1.96.0` ended up in an exceptionaly unusual state because `master` was merged into it before it was merged into `master`. See commit `a2cdac6` in https://github.com/kubecost/cost-analyzer-helm-chart/pull/1632.

Andrew is going to open a fresh PR branched off of `develop` and then we'll be out of trouble!

PR #1652, opened 2022-08-22
Add ability to add 'affinity' value to 'prometheus/node-exporter'

kc-adawson:
## What does this PR change?

Adds ability to set affinity and nodeSelectors to run only on Linux nodes via values file.

No.

Adds ability to set affinity and nodeSelectors to run only on Linux nodes.

-

## How was this PR tested?

Tested by Jesse on EKS with windows nodes. 

Will be added to this page:
https://guide.kubecost.com/hc/en-us/articles/6152374933655-Windows-Node-Support

### DIFF
--- a/cost-analyzer/charts/prometheus/templates/node-exporter-daemonset.yaml
+++ b/cost-analyzer/charts/prometheus/templates/node-exporter-daemonset.yaml
@@ -30,6 +30,10 @@ spec:
 {{ toYaml .Values.nodeExporter.pod.labels | indent 8 }}
 {{- end }}
     spec:
+{{- if .Values.nodeExporter.affinity }}
+      affinity:
+{{ toYaml .Values.nodeExporter.affinity | indent 8 }}
+{{- end }}
       serviceAccountName: {{ template "prometheus.serviceAccountName.nodeExporter" . }}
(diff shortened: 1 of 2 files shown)


mbolt35 (approved):
This looks great. I'm mostly unsure about the "multiple values.yaml" file approach we're taking - it's seems like a very good way to separate different groups of installs, which I think is awesome. Is there a "better" way to do it? Is this something that's intuitive for customers? All signs point to yes, just wanted to check :)

kc-adawson:
Our thought was being able to give the customer a base install command that covers it, like: 
`helm upgrade kubecost "kubecost/cost-analyzer" --namespace kubecost --install -f https://raw.githubusercontent.com/kubecost/cost-analyzer-helm-chart/adawson-value-update/cost-analyzer/values-windows-node-affinity.yaml -f ./values-amazon-primary.yaml

The one issue i've found here is that clients will try to upgrade in the future using only their` -f ./values-amazon-primary.yaml` file, and it will overwrite the winodws-node-affinity.

What about writing into the helm template values something like `linuxOnly: true`, and if that is marked true then the affinity and nodeSelectors are automatically added properly? That may be a more graceful way than requiring the -f helm files.

PR #1654, opened 2022-08-23
prometheus considerations + link KSM values

jessegoodier:
## What does this PR change?

Add link to documentation around disabling Prometheus.
Add section for disabling KSM

- NA

- Reduces the support issues from users implementing kubecost on existing Prometheus
- Multiple KSM pods are inefficient

NA

## How was this PR tested?

```
helm template kubecost \
    --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
    --namespace kubecost --create-namespace \
    --set prometheus.nodeExporter.enabled=false \
    --set prometheus.serviceAccounts.nodeExporter.create=false \
    --set prometheus.kubeStateMetrics.enabled=false|ag 'statemetric|state-metric'
```
image

yes: https://github.com/kubecost/docs/blob/main/custom-prom.md

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1,7 +1,7 @@
 global:
   # zone: cluster.local (use only if your DNS server doesn't live in the same zone as kubecost)
   prometheus:
-    enabled: true # If false, Prometheus will not be installed -- only actively supported on paid Kubecost plans
+    enabled: true # If false, Prometheus will not be installed -- please read this before disabling: https://github.com/kubecost/docs/blob/main/custom-prom.md
     fqdn: http://cost-analyzer-prometheus-server.default.svc #example address of a prometheus to connect to. Include protocol (http:// or https://) Ignored if enabled: true
     # insecureSkipVerify : false # If true, kubecost will not check the TLS cert of prometheus
     # queryServiceBasicAuthSecretName: dbsecret # kubectl create secret generic dbsecret -n kubecost --from-file=USERNAME --from-file=PASSWORD
@@ -184,7 +184,7 @@ oidc:
   clientID: "" # application/client client_id paramter obtained from provider, used to make requests to server
   clientSecret: "" # application/client client_secret paramter obtained from provider, used to make requests to server
   secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
-  authURL: "https://my.auth.server/authorize" # endpoint for login to auth server 
+  authURL: "https://my.auth.server/authorize" # endpoint for login to auth server
   loginRedirectURL: "http://my.kubecost.url/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
   discoveryURL: "https://my.auth.server/.well-known/openid-configuration" # url for OIDC endpoint discovery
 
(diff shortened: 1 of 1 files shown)


linhlam-kc (approved):
LGTM

srpomeroy:
The root readme.md and the cost-analyzer/readme.md both reference `prometheus.kube-state-metrics.disabled`. We should consider either updating the readmes or switching to that value if it makes more sense.

jessegoodier:
I'm not merging until Bolt gives his ok. 

It is odd. Prometheus has updated to `condition: kubeStateMetrics.enabled`
(not that this matters): https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/Chart.yaml#L29)

From what we have: 
`  - condition: kube-state-metrics.disabled`


but disabled doesn't work:
```
helm template kubecost \
    --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
    --namespace kubecost --create-namespace \
    --set prometheus.nodeExporter.enabled=false \
    --set prometheus.serviceAccounts.nodeExporter.create=false \
    --set prometheus.kubeStateMetrics.disabled=true \
    | grep -iE 'kube-state-metrics/templates/deployment' -A5
```
output:
```# Source: cost-analyzer/charts/prometheus/charts/kube-state-metrics/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kubecost-kube-state-metrics
...
```

whereas enabled=false works as captured in initial testing section does not deploy KSM.

srpomeroy:
For `disabled` it should be `--set prometheus.kube-state-metrics.disabled=true` note the hyphens.
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/charts/prometheus/charts/kube-state-metrics/templates/deployment.yaml#L1

jessegoodier:
> For `disabled` it should be `--set prometheus.kube-state-metrics.disabled=true` note the hyphens. https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/charts/prometheus/charts/kube-state-metrics/templates/deployment.yaml#L1

sure enough:
```
helm template kubecost \
    --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
    --namespace kubecost --create-namespace \
    --set prometheus.kube-state-metrics.disabled=true 
```
also disables it.

So, which is better @mbolt35 ?
--set prometheus.kubeStateMetrics.enabled=false
or
--set prometheus.kube-state-metrics.disabled=true 

or just do both?

mbolt35:
@jessegoodier I'll look at this today at some point. A few things to note before I dive into the specifics: we don't require the KSM deployment anymore, as our cost-model pod emits all of the KSM metrics we use. (Although deploying KSM is fine because we handle duplicate metrics cases anyways).

jessegoodier:
maybe we should just set KSM to disabled here by default?

jessegoodier:
updated per @mbolt35's comment on KSM on other PR regarding both KSM enable/disable. 
also added comment for additional details.

I believe this is good to go. Please reapprove if you agree.

mbolt35 (approved):
Love the added troubleshoot links! Also, yes, I think at some point we will probably want to unify the whole `kubeStateMetrics` and `kube-state-metrics` duality, but the comment and explicit defaults capture it well. Thanks Jesse!

PR #1658, opened 2022-08-24
Update securityContext configs to support Openshift cluster.

linhlam-kc:
## What does this PR change?
This PR is to update cost-analyzer helmchart to support Openshift. These are required changes:

* Update Security context in Values.yaml of following sub charts to allow Openshift assigns random UIDs:
  * [Prometheus](https://github.com/kubecost/openshift-helm-chart/blob/7ad4e779e621b7daba96fe684c5bfff33fe0429a/cost-analyzer/charts/prometheus/values.yaml)
  * [Thanos](https://github.com/kubecost/openshift-helm-chart/blob/linh-cost-analyzer-195/cost-analyzer/charts/thanos/values.yaml)
* Update [cost-analyzer-deployment-template](https://github.com/kubecost/openshift-helm-chart/blob/linh-cost-analyzer-195/cost-analyzer/templates/cost-analyzer-deployment-template.yaml) to use random UID when Kubecost is installed on Openshift
* Add additional specific Values file for Openshift name `values-openshift.yaml`

N/A

* This PR allows customer to install Kubecost on Openshift easily using our main Helm chart repository. Currently, customer need to clone this Helm chart for Openshift installation https://github.com/kubecost/openshift-helm-chart

N/A

## How was this PR tested?
* It was tested on Openshift 4.10.25 following instructions at https://github.com/kubecost/openshift-helm-chart/blob/main/README.md

* Yes and it need to update Helm chart location once this PR is merged.

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-162) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/values-openshift.yaml
+++ b/cost-analyzer/values-openshift.yaml
@@ -1,0 +1,36 @@
+kubecostProductConfigs:
+  clusterName: CLUSTER_NAME
+  projectID: CLUSTER_NAME
+
+prometheus:
+  nodeExporter:
+    enabled: false
+  serviceAccounts:
+    nodeExporter:
+      create: false
+  kube-state-metrics:
+    disabled: true
+  server:
+    global:
+      external_labels:
+        cluster_id: CLUSTER_NAME
+
+networkCosts:
+  enabled: false
+global:
+  grafana:
+    enabled: false
+    proxy: false
+
+
+kubecostFrontend:
+  fullImageName: "linhlamkc/cost-analyzer-frontend:prod-1.96.0"
+openshiftDeployment: true
+
+# Define persistence volume for cost-analyzer
+persistentVolume:
+  size: 32Gi
+  dbSize: 32.0Gi
+  enabled: true # Note that setting this to false means configurations will be wiped out on pod restart.
+  # storageClass: "-" #
+  # existingClaim: kubecost-cost-analyzer # a claim in the same namespace as kubecost
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -276,11 +276,11 @@ alertmanager:
 
   ## Security context to be added to alertmanager pods
   ##
-  securityContext:
-    runAsUser: 1001
-    runAsNonRoot: true
-    runAsGroup: 1001
-    fsGroup: 1001
+  securityContext: {}
+    # runAsUser: 1001
+    # runAsNonRoot: true
+    # runAsGroup: 1001
+    # fsGroup: 1001
 
   service:
     annotations: {}
@@ -858,11 +858,11 @@ server:
 
   ## Security context to be added to server pods
   ##
-  securityContext:
-    runAsUser: 1001
-    runAsNonRoot: true
-    runAsGroup: 1001
-    fsGroup: 1001
+  securityContext: {}
+    # runAsUser: 1001
+    # runAsNonRoot: true
+    # runAsGroup: 1001
+    # fsGroup: 1001
 
   service:
     annotations: {}
@@ -1018,9 +1018,9 @@ pushgateway:
 
   ## Security context to be added to push-gateway pods
   ##
-  securityContext:
-    runAsUser: 1001
-    runAsNonRoot: true
+  securityContext: {}
+    # runAsUser: 1001
+    # runAsNonRoot: true
 
   service:
     annotations:
--- a/cost-analyzer/charts/thanos/values.yaml
+++ b/cost-analyzer/charts/thanos/values.yaml
@@ -122,10 +122,10 @@ store:
       #    hosts:
       #      - chart-example.local
   # Optional securityContext
-  securityContext:
-    fsGroup: 1001
-    runAsNonRoot: true
-    runAsUser: 1001
+  securityContext: {}
+    # fsGroup: 1001
+    # runAsNonRoot: true
+    # runAsUser: 1001
 
   resources: {}
   #  limits:
@@ -294,10 +294,10 @@ queryFrontend:
       labels: {}
 
   # Optional securityContext
-  securityContext:
-    fsGroup: 1001
-    runAsNonRoot: true
-    runAsUser: 1001
+  securityContext: {}
+    # fsGroup: 1001
+    # runAsNonRoot: true
+    # runAsUser: 1001
 
   resources: {}
   #  limits:
@@ -460,10 +460,10 @@ query:
       labels: {}
 
   # Optional securityContext
-  securityContext:
-    fsGroup: 1001
-    runAsNonRoot: true
-    runAsUser: 1001
+  securityContext: {}
+    # fsGroup: 1001
+    # runAsNonRoot: true
+    # runAsUser: 1001
 
   resources: {}
   #  limits:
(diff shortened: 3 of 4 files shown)


mbolt35:
@linhlam-kc This looks good... Since we're removing the defaults from the `prometheus/values.yaml` and `thanos/values.yaml`, can we add those defaults back in `cost-analyzer/values.yaml` and `cost-analyzer/values-thanos.yaml`? I think we want to make sure that any existing customers that run upgrades don't run into conflicts.

linhlam-kc:
> @linhlam-kc This looks good... Since we're removing the defaults from the `prometheus/values.yaml` and `thanos/values.yaml`, can we add those defaults back in `cost-analyzer/values.yaml` and `cost-analyzer/values-thanos.yaml`? I think we want to make sure that any existing customers that run upgrades don't run into conflicts.

@mbolt35 Thank you so much for helping to review. I have tested to add those defaults back to `values.yaml` and `values-thanos.yaml` . However, doing that prevents `values-openshift.yaml` overwrite `securitycontext` with `null` value similarly when we have those defaults in `prometheus/values.yaml` and `thanos/values.yaml`. From my testing, the configuration for Openshift is also working for Amazon EKS. How about we add those defaults back to `values.yaml` and `values-thanos.yaml` but commend them out, so customer can also enable them easily if they want to. For example:
```
  alertmanager:
    enabled: false
    persistentVolume:
      enabled: true
    # securityContext:
    #   runAsUser: 1001
    #   runAsNonRoot: true
    #   runAsGroup: 1001
    #   fsGroup: 1001
  nodeExporter:
    enabled: true
  pushgateway:
    enabled: false
    persistentVolume:
      enabled: true
    # securityContext:
    #   runAsUser: 1001
    #   runAsNonRoot: true
```

dwbrown2:
@mbolt35 @linhlam-kc what are suggested next steps on this one? do we want this as part of 1.97?

cost-analyzer/values-openshift.yaml
```diff
@@ -0,0 +1,36 @@
...
+  clusterName: CLUSTER_NAME
+  projectID: CLUSTER_NAME
+
+prometheus:
+  nodeExporter:
+    enabled: false
+  serviceAccounts:
+    nodeExporter:
+      create: false
+  kube-state-metrics:
```
jessegoodier, line 11: Bolt suggested doing both on the other PR:
  kubeStateMetrics:
    enabled: false
  kube-state-metrics:
    disabled: true

mbolt35:
> @mbolt35 Thank you so much for helping to review. I have tested to add those defaults back to values.yaml and values-thanos.yaml . However, doing that prevents values-openshift.yaml overwrite securitycontext with null value. It is similar to when we have those defaults in prometheus/values.yaml and thanos/values.yaml. From my testing, the configuration for Openshift is also working for Amazon EKS. How about we add those defaults back to values.yaml and values-thanos.yaml but commend them out, so customer can also enable them easily if they want to. For example:
```
  alertmanager:
    enabled: false
    persistentVolume:
      enabled: true
    # securityContext:
    #   runAsUser: 1001
    #   runAsNonRoot: true
    #   runAsGroup: 1001
    #   fsGroup: 1001
  nodeExporter:
    enabled: true
  pushgateway:
    enabled: false
    persistentVolume:
      enabled: true
    # securityContext:
    #   runAsUser: 1001
    #   runAsNonRoot: true
```

@linhlam-kc  We have to expect our users won't change anything, which will be a concern. I'm not sure why the overwrites aren't working.

If I run my helm command with `-f <values1.yaml> -f <values2.yaml>` then each values file is applied in the order they're written in the command. ie: If `values2.yaml` in the above example overwrites values from `values1.yaml` then it will have the `values2.yaml` values. 

Am I missing something?

mbolt35:
> @mbolt35 @linhlam-kc what are suggested next steps on this one? do we want this as part of 1.97?

@dwbrown2 I'll keep an eye on this ticket - I assume this is something that we can get in the release as long as we can button up any of the remaining discrepancies. I just want to ensure that we want break existing users attempting to helm upgrade by swapping out default values.

linhlam-kc:
@mbolt35 Sorry for the late reply here, thank you for clarifying about your concern. I agree that it could have un-expected side effect given that there is a change that customer does not allow 'null' securityContext. Regarding to the issue, it is related to this Helm client bug: https://github.com/helm/helm/issues/5184

The issue is that if the default values (value file 1) are null, it could be overwritten successfully but it doesn't work if the default values (value file 1) arenot null that need to be replaced by null values. To understand the bug easier, please see the following examples below

For example: 

**Scenario 1**: 
- The default value in `values.yaml` and `values-thanos.yaml` is null: `securityContext: {}`
- The value in `values-openshift.yaml` is:
```yaml
    securityContext:
       fsGroup: 1001
       runAsNonRoot: true
       runAsUser: 1001
```
- I run the following Helm command to generate flat manifests:
```bash
helm template kubecost4 . \
--namespace kubecost4 \
--create-namespace \
-f ./values-thanos.yaml \
-f ./values-openshift.yaml
```
- Result: `securityContext` is overwritten successfully with the values from `values-openshift.yaml` 

**Scenario 2**: 
- The default value in `values.yaml` and `values-thanos.yaml` is not null: 
```yaml
    securityContext:
       fsGroup: 1001
       runAsNonRoot: true
       runAsUser: 1001
```

- The value in `values-openshift.yaml` is null: `securityContext: {}`

- I run the following Helm command to generate flat manifests:
```bash
helm template kubecost4 . \
--namespace kubecost4 \
--create-namespace \
-f ./values-thanos.yaml \
-f ./values-openshift.yaml
```
- Result: securityContext is not overwritten with the default values from `values.yaml` and `values-thanos.yaml`

mbolt35:
@linhlam-kc Looking through this PR, and based on the helm bug description, it's still not adding up: you're changing all non-null values to null value defaults. If you're overwriting in the openshift values with a _non-null_ security context, then why are you changing the existing defaults? 

From our discussions, you're not going to be setting openshift's security context to null, in which case, we would _never_ run into **Scenario 2**, so we shouldn't have to change the existing defaults.

linhlam-kc:
> @linhlam-kc Looking through this PR, and based on the helm bug description, it's still not adding up: you're changing all non-null values to null value defaults. If you're overwriting in the openshift values with a _non-null_ security context, then why are you changing the existing defaults?
> 
> From our discussions, you're not going to be setting openshift's security context to null, in which case, we would _never_ run into **Scenario 2**, so we shouldn't have to change the existing defaults.

@mbolt35 My previous comment [here](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1658#issuecomment-1238755222) is actually only for clarifying the Helm client bug which describe why adding non-null_securityContext back in `values.yaml` and `values-thanos.yaml` is not feasible after setting them to null_SecurityContext in `Prometheus's chart values.yaml` and `Thanos's chart values.yaml`. The real changes for OpenShift is to have all required SecurityContext to be null as in my commit, so OpenShift can assign random `userid` and `groupid` when Kubecost is installed.

mbolt35:
> The real changes for OpenShift is to have all required SecurityContext to be null as in my commit

Ok, this is what I missed. Thanks for the explanation. 

Given that this is a helm issue we can't avoid, I need to think about a better way to do this, or look into a different approach. It doesn't give me much hope that helm hasn't addressed this either.

chipzoller:
This will be resolved shortly by introducing a `platforms` key with a switch for OpenShift (and others) which will result in the appropriate securityContext to be set.

PR #1661, opened 2022-08-25
feat: respect the fullnameOverride value for resource names

diogokiss:
Currently, not all resources respect the fullnameOverride when set in the
Values. For example, the cost-analyzer service name.

This commit extends the pattern where the resource names are defined based on
`.Values.fullnameOverride` in case that is set.

## What does this PR change?
Some resource names when `.Values.fullnameOverride` is set.

No.

If `.Values.fullnameOverride` is not set, there is no change in the behavior.
If `.Values.fullnameOverride` is set, some resources will have their name changed.

None.

## How was this PR tested?
Yes, it's been tested in my Testing cluster.

No.

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-164) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -63,25 +63,25 @@ Create the fully qualified name for Prometheus alertmanager service.
 {{- end -}}
 
 {{- define "cost-analyzer.serviceName" -}}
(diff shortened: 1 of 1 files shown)


chipzoller:
Conflicts to resolve here.

PR #1685, opened 2022-09-08
[DRAFT] thomasn extra thanos pod

thomasvn:
DO NOT MERGE. This is a test branch.

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-175) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/charts/thanos/templates/query-deployment.yaml
+++ b/cost-analyzer/charts/thanos/templates/query-deployment.yaml
@@ -78,7 +78,9 @@ spec:
         {{- end }}
         {{- if .Values.query.storeDNSDiscovery }}
         - "--store=dnssrv+_grpc._tcp.{{ include "thanos.componentname" (list $ "store") }}-grpc.{{ .Release.Namespace }}.svc"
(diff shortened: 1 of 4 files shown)


thomasvn:
I've updated the PR to create multiple Thanos deployments `-test1`, `-test2`, `-test3`, in the `store-deployment2.yaml` file. 

Each of these deployments have differing arguments for `--min-time=` and `--max-time=` so that each deployment is responsible for metrics from different time ranges. This should effectively horizontally scale your Thanos solution.

I additionally made changes to `store-service2.yaml` and `query-deployment.yaml` to reflect the multiple new deployments that were created.

thomasvn:
Previously, all three Deployments were using a single PVC.

Now each Deployment gets its own PVC which are named `-test1`, `-test2`, `-test3` corresponding to their deployment. Changes made in the `store-pvc2.yaml` and `store-deployment2.yaml` file.

thomasvn:
Successfully tested changes on a smaller dataset:

- `--min-time=-3d` and `--max-time=-2d`
- `--min-time=-2d` and `--max-time=-1d`
- `--min-time=-1d`

PR #1703, opened 2022-09-19
NGINX changes to open node port and container port to serve grpc request

avrodrigues5:
## What does this PR change?
Opens Node port 50051 to serve GRPC request and port forwards grpc request to cost-model container port running grpc server on 50052 

- Will open a KCM PR soon!
- 

Hopefully none , i dont think changes to this would impact regular http calls.

- 
- 

## How was this PR tested?
Applying changes to helm deployment and running GRPC calls along with regular HTTP calls.

Don't know where to update the documentation for this.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -303,6 +303,10 @@ spec:
         - image: gcr.io/kubecost1/cost-model:prod-{{ $.Chart.AppVersion }}
         {{ end }}
           name: cost-model
+          ports:
+          - containerPort: 50052
+            name: grpcserver
+            protocol: TCP
         {{- if .Values.kubecostModel.extraArgs }}
           args:
           {{- toYaml .Values.kubecostModel.extraArgs | nindent 12 }}
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -196,7 +196,7 @@ data:
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
 
-        location /pkg.service.allocation.v1/ {
+        location /services/ {
             grpc_pass grpc://grpcserver;
         }
 
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -62,6 +62,17 @@ data:
 {{- end }}
     }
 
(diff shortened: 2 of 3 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -303,6 +303,10 @@ spec:
         - image: gcr.io/kubecost1/cost-model:prod-{{ $.Chart.AppVersion }}
         {{ end }}
           name: cost-model
+          ports:
+          - containerPort: 50052
+            name: grpcserver
```
sundeepghuman, line 308: Is this one port for all services? I.e. if we had an assets service would it also be on this port?

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -184,6 +196,10 @@ data:
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
 
+        location /pkg.service.allocation.v1/ {
```
sundeepghuman, line 199: same question whether all services will share 'grpcserver'

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
avrodrigues5: It will be upto you as to how you would want it because u can open another tcp port exclusive of this to listen to new request or have a new route with same grpc server. 

If we choose the latter, you would register it with the same tcp port
Like in here you would need another similar registering of service https://github.com/kubecost/kubecost-cost-model/pull/955/files#diff-c9127ce2d87ae5fe988d6b5f16b63f7ba022998d617dfbafd0381bfda6ceba41R2449|

Something along lines of https://stackoverflow.com/questions/52634217/access-multiple-grpc-services-over-the-same-connection

Once that's done you would need to have another route defined pointing to package of the new service so it would route the request appropriately .

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
mbolt35: I think all services should share grpc server, so we likely want to make the endpoint `/` if we can? @avrodrigues5 can we just make it:
```
location / {
...
}
```

?

difference.txt
```diff
@@ -0,0 +1,76 @@
+diff --git a/cost-analyzer/Chart.yaml b/cost-analyzer/Chart.yaml
```
mbolt35, line 1: This file was accidentally added?

avrodrigues5:
This PR might act as a template when GRPC server are created in future but for allocation/view API we ended up not creating a GRPC server so closing this PR

Issue #1605, opened 2022-08-03
Stuck with OIDC google auth (v1.98)


korjavin:
Kubecost installed with official helm chart.

Values.yaml contains:

```
oidc:
  enabled: true
  clientID: " #real client id that works for another project
  clientSecret: "" # a real secret
  secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored   
  authURL: "https://accounts.google.com/o/oauth2/v2/auth" # endpoint for login to auth server-
  loginRedirectURL: "https://mydomain.com/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
  discoveryURL: "https://accounts.google.com/.well-known/openid-configuration" # url for OIDC endpoint discovery
```

I checked that redirect URL included to “authorized” list on the Google site.

When I open kubecost front page I'm getting
```
Authorization Error
Error 400: invalid_request
Required parameter is missing: response_type
[Learn more](https://developers.google.com/identity/protocols/oauth2)
```

![image](https://user-images.githubusercontent.com/148765/182556862-539a3731-bd3d-440c-9c2b-3cc68fd4ba9a.png)

Is there anything I can do with this issue?

gz#2289<br><br><i>(related to <a href='https://kubecost.zendesk.com/agent/tickets/2289'>Zendesk ticket #2289</a>)</i>

Adam-Stack-PM:
@zorrobyte, Any update on this ticket? Is this a bug or support issue?

kaelanspatel:
Hi @korjavin 

This should be made clearer, so we'll work on documentation here, but the AuthURL parameter in the helm chart needs to be the full URL required for login to the OIDC provider _including_ various query params (these are different based on which provider you use, and so are left user-configurable).

To resolve this specific error you'd need to specify at least:
```
 authURL:  "https://accounts.google.com/o/oauth2/v2/auth?response_type=code&client_id=<your client ID>"
```
On a cursory glance at [the docs](https://developers.google.com/identity/protocols/oauth2/web-server#sample-oauth-2.0-server-response) it looks like there may be more params required, so it may do to add these as well.

korjavin:
It worked.

But how can I limit domain to mine.

biancaburtoiu:
Hello @korjavin! Thanks for letting us know the `authURL` worked once you added the additional params. 

Could you explain a bit more what you are trying to achieve by limiting the domain?

korjavin:
[...]

PR #1706, opened 2022-09-20
Set up OIDC hosted domain check

biancaburtoiu:
## What does this PR change?
Include domain in helm parameters for OIDC auth.

https://github.com/kubecost/kubecost-cost-model/pull/951

Users can now set a `.Values.oidc.hostedDomain` parameter to block access for users outside of the configured domain.

**Note** Removing a previously set helm domain value is done by setting `.Values.oidc.hostedDomain = ""`, as opposed to simply removing the `hostedDomain` field from the config. 

https://github.com/kubecost/cost-analyzer-helm-chart/issues/1605

## How was this PR tested?
Verify that new helm parameter is parsed and loaded correctly in the `OIDCConfigs` object to be used at authentication time.

Will add domain note to internal OIDC documentation

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -199,6 +199,7 @@ oidc:
   authURL: "https://my.auth.server/authorize" # endpoint for login to auth server
   loginRedirectURL: "http://my.kubecost.url/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
   discoveryURL: "https://my.auth.server/.well-known/openid-configuration" # url for OIDC endpoint discovery
+  hostedDomain: "example.com" # blocks access to the auth domain specified in the hd claim of the provider ID token
 
 # Adds an httpProxy as an environment variable. systemProxy.enabled must be `true`to have any effect.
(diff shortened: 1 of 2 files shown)


kaelanspatel (approved):

cost-analyzer/values.yaml
```diff
@@ -199,6 +199,7 @@ oidc:
   authURL: "https://my.auth.server/authorize" # endpoint for login to auth server
   loginRedirectURL: "http://my.kubecost.url/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
   discoveryURL: "https://my.auth.server/.well-known/openid-configuration" # url for OIDC endpoint discovery
+  hostedDomain: "domain" # blocks access to specified domain only
```
LaikaN57, line 202: Needs better default example. Is this talking about the auth domain or the application/service domain?

LaikaN57 (changes requested):

cost-analyzer/values.yaml
biancaburtoiu: thank you for your feedback, @LaikaN57! I've updated the description to capture more detail - if set, we expect the value of `hostedDomain` to match the `hd` claim part of the ID token supplied during the authentication process.

Issue #1709, opened 2022-09-21
create serviceMonitor for network-costs


jessegoodier:
**What problem are you trying to solve?**  

kubecost have a serviceMonitor for kubecost metrics, but not for the daemonset.

FYI: the scrapeConfig does work

**To Reproduce**  

there is no serviceMonitor in the repo for network-costs

**Expected behavior**  
create network cost service monitor similar to:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-servicemonitor-template.yaml

**Describe the solution you'd like**  

helm values

**Describe alternatives you've considered**  

https://github.com/kubecost/poc-common-configurations/blob/main/custom-tsdb/network-costs-serviceMonitor.yaml

**How would users interact with this feature?**
helm:

```
networkCosts:
  serviceMonitor: 
    enabled: true
```

mbolt35:
Hey @jessegoodier - I think this is because we don't want the functionality from a service monitor when it comes to network costs. A service in this case would only extract metrics from one of the N  daemonset network costs pods. We have a `PodMonitor` template though: 

https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-network-costs-podmonitor-template.yaml

I think a ServiceMonitor for network-costs will cause problems since it's basically load-balancing scrape requests. There _is_ a way that we have setup to create a service for network-costs, but it requires that the network-costs setup be removed from the values.yaml and a special configuration has to be used with Prometheus to ensure that it uses "endpoint" scraping (Service creates N endpoints, all endpoints are scraped). There may be some way to configure the endpoint scrapes AND a service monitor? In general though, we just want to ensure that prometheus is always scraping all of the network costs pods AND if someone is using prom operator, it's often just easier to use the pod monitor instead of a service monitor.

mbolt35:
Just so it's clear, there are basically two ways you can setup scraping for network-cost pods. The _default_ way in our current helm chart is with this prometheus "extra" config:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/9c93abb6ba432d95e84f787c3000c38c5b5414b8/cost-analyzer/values.yaml#L449-L456

(Note line 41 - `sd_role: pod`) indicating that it will match pods to scrape based on the name. There's no service to create, just create the daemonset, update prom config, and it works. 

[...]

PR #1710, opened 2022-09-21
network-costs serviceMonitor

jessegoodier:
## What does this PR change?

Add a serviceMonitor for network-costs

- NA

Created serviceMonitor for the network-costs daemonset for users with Prometheus Operator

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1709

## How was this PR tested?

helm install against a test cluster with prometheus operator

Probably should update https://guide.kubecost.com/hc/en-us/articles/4407595941015-Custom-Prometheus
but have an example here for now: https://github.com/kubecost/poc-common-configurations/blob/main/custom-tsdb/values-primary.yaml#L12


mbolt35 (approved):
LGTM - I think as long as we have the proper instruction required to ensure that we are scraping _all_ network-cost pods, then this is good to merge.

PR #1444, opened 2022-05-27
Update cluster-controller to v0.1.0

michaelmdresser:
## What does this PR change?

- Updates cluster-controller to v0.1.0, which includes v2 of cluster-turndown
- Updates the turndown CRD to match v2 of cluster-turndown

This includes a BREAKING CHANGE to the cluster-turndown functionality
that is wrapped by cluster-controller. The breaking changes are
summarized in https://github.com/kubecost/cluster-turndown/releases/tag/v2.0.0
and should be featured prominently in release notes.

The breaking change is necessary because we aren't allowed to use `*.k8s.io` to namespace our CRDs. See the notes in https://github.com/kubecost/cluster-turndown/pull/44 for a further explanation and references.

## Does this PR rely on any other PRs?

- https://github.com/kubecost/cluster-controller/pull/7 and the image build of 0.1.0 that follows

## How does this PR impact users? (This is the kind of thing that goes in release notes!)

- Updates cluster-controller to v0.1.0.
- Removes `apiextensions.k8s.io/v1beta1` CRD `turndownschedules.kubecost.k8s.io`
- Introduces `apiextensions.k8s.io/v1` CRD `turndownschedules.kubecost.com` as replacement.
- :warning: Breaking change for users of the alpha feature "Cluster Turndown" contained in the Cluster Controller. If you have Cluster Controller enabled (`--set clusterController.enabled=true`), please visit the [migration guide](https://github.com/kubecost/docs/blob/main/v1-94-turndown-schedule-migration-guide.md) to check if you have existing resources that must be migrated to maintain expected functionality.

## Links to Issues or ZD tickets this PR addresses or fixes

- Should resolve issues like https://github.com/kubecost/cluster-turndown/issues/40 for those who use turndown through cluster controller. I have not yet seen a report of this, but it can happen.

## How was this PR tested?

Deployed this version of the Helm chart with `--set clusterController.enabled=true`. Saw the `turndownschedules.kubecost.com` CRD with `kubectl get crd`. Saw the cluster-controller Pod start and observed clean logs:
```
→ k logs -n kubecost -l app=kubecost-cluster-controller --tail=-1 -f
W0527 17:34:20.783003 1013105 gcp.go:120] WARNING: the gcp auth plugin is deprecated in v1.22+, unavailable in v1.25+; use gcloud instead.
To learn more, consult https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
[...]

PR #1713, opened 2022-09-26
whitelist PV usage metrics 

nickcurie:
## What does this PR change?
Whitelists `kubelet_volume_stats_used_bytes`

- [OC#1403](https://github.com/opencost/opencost/pull/1403)
- [KCM#963](https://github.com/kubecost/kubecost-cost-model/pull/963)

Whitelisted PV usage metrics into Prometheus

- 
- 

## How was this PR tested?
Locally

### DIFF
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -1263,14 +1263,6 @@ serverFiles:
           - source_labels: [ __name__ ]
             regex: (kubelet_volume_stats_used_bytes)
             action: keep
-          - source_labels: [ container ]
-            target_label: container_name
-            regex: (.+)
-            action: replace
-          - source_labels: [ pod ]
-            target_label: pod_name
-            regex: (.+)
-            action: replace
 
       # Scrape config for service endpoints.
       #
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -1261,7 +1261,7 @@ serverFiles:
 
         metric_relabel_configs:
           - source_labels: [ __name__ ]
-            regex: (kubelet_volume_stats_used_bytes)
+            regex: (kubelet_volume_stats_used_bytes) # this metric is in alpha 
             action: keep
 
       # Scrape config for service endpoints.
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -1216,6 +1216,62 @@ serverFiles:
             regex: (.+)
             action: replace
 
+      # A scrape configuration for running Prometheus on a Kubernetes cluster.
+      # This uses separate scrape configs for cluster components (i.e. API server, node)
+      # and services to allow each to use different authentication configs.
+      #
+      # Kubernetes labels will be added as Prometheus labels on metrics via the
+      # `labelmap` relabeling action.
+
+      - job_name: 'kubernetes-nodes'
+
+        # Default to scraping over https. If required, just disable this or change to
+        # `http`.
+        scheme: https
+
+        # This TLS & bearer token file config is used to connect to the actual scrape
+        # endpoints for cluster components. This is separate to discovery auth
+        # configuration because discovery & scraping are two separate concerns in
+        # Prometheus. The discovery auth config is automatic if Prometheus runs inside
+        # the cluster. Otherwise, more config options have to be provided within the
+        # <kubernetes_sd_config>.
+        tls_config:
+          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+          # If your node certificates are self-signed or use a different CA to the
+          # master CA, then disable certificate verification below. Note that
+          # certificate verification is an integral part of a secure infrastructure
+          # so this should only be disabled in a controlled environment. You can
+          # disable certificate verification by uncommenting the line below.
+          #
+          insecure_skip_verify: true
+        bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+
+        kubernetes_sd_configs:
+          - role: node
+
+        relabel_configs:
+          - action: labelmap
+            regex: __meta_kubernetes_node_label_(.+)
+          - target_label: __address__
+            replacement: kubernetes.default.svc:443
+          - source_labels: [__meta_kubernetes_node_name]
+            regex: (.+)
+            target_label: __metrics_path__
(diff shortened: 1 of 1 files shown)


cost-analyzer/charts/prometheus/values.yaml
```diff
@@ -1216,6 +1216,62 @@ serverFiles:
...
+          - target_label: __address__
+            replacement: kubernetes.default.svc:443
+          - source_labels: [__meta_kubernetes_node_name]
+            regex: (.+)
+            target_label: __metrics_path__
+            replacement: /api/v1/nodes/$1/proxy/metrics
+
+        metric_relabel_configs:
+          - source_labels: [ __name__ ]
+            regex: (kubelet_volume_stats_used_bytes)
```
michaelmdresser, line 1264: IIRC `kubelet_volume_stats_used_bytes` is in alpha status as a metric. Based on the [documentation](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#metric-lifecycle), that means that the metric is subject to changes or removal at any time.

Is the code you've written using this metric resilient to the removal/change of the metric? In other words, if the metric were to disappear tomorrow, would the rest of the Asset/Disk pipeline still continue to function as it does today?

For product/eng leadership (@dwbrown2, @nikovacevic): Do we have guidelines on building Kubecost functionality on top of alpha/beta K8s functionality? For context, we've previously had contention over a beta->v1 migration/breaking change with cluster-turndown's CRD: https://github.com/kubecost/cost-analyzer-helm-chart/pull/1444#issuecomment-1145092419.

cost-analyzer/charts/prometheus/values.yaml
```diff
@@ -1216,6 +1216,62 @@ serverFiles:
...
+            action: keep
+          - source_labels: [ container ]
+            target_label: container_name
+            regex: (.+)
+            action: replace
+          - source_labels: [ pod ]
+            target_label: pod_name
+            regex: (.+)
+            action: replace
```
michaelmdresser, line 1273: Are these relabel configs necessary?

cost-analyzer/charts/prometheus/values.yaml
dwbrown2: Are we sure this metric is still alpha? I haven't chased all the way but found this: https://github.com/kubernetes/kubernetes/pull/51553

I think these decisions should be situational. I personally think adding a metric here is valuable as with our past experience it was accurate the vast majority of the time. I do think we should be resilient in the application experience to it being not available on a cluster. We should also consider marking this feature beta in the UI.

cost-analyzer/charts/prometheus/values.yaml
```diff
@@ -1216,6 +1216,62 @@ serverFiles:
...
             action: replace
 
+      # A scrape configuration for running Prometheus on a Kubernetes cluster.
+      # This uses separate scrape configs for cluster components (i.e. API server, node)
+      # and services to allow each to use different authentication configs.
+      #
+      # Kubernetes labels will be added as Prometheus labels on metrics via the
+      # `labelmap` relabeling action.
+
+      - job_name: 'kubernetes-nodes'
```
dwbrown2, line 1226: I would have expected this PR to simply add a new metric to the whitelist, but looks like we need to add an entirely new scrape endpoint. Can you share more?

cost-analyzer/charts/prometheus/values.yaml
michaelmdresser: I have a K8s v1.22 GKE cluster:
```
→ k version 
...
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.12-gke.2300", GitCommit:"e55564cf3a1384026a54920174977659c8c56a50", GitTreeState:"clean", BuildDate:"2022-08-16T09:24:51Z", GoVersion:"go1.16.15b7", Compiler:"gc", Platform:"linux/amd64"}
...
```

When I look for this metric (its a standard kubelet metric, so at a node's kubelet's /metrics, I see that its an ALPHA status:

```
kubectl proxy --port 8033

curl 'localhost:8033/api/v1/nodes/gke-michael-dev-pool-1-92180698-v5zo/proxy/metrics' | grep -i 'kubelet_volume_stats_used_bytes'

...
# HELP kubelet_volume_stats_used_bytes [ALPHA] Number of used bytes in the volume
# TYPE kubelet_volume_stats_used_bytes gauge
```

Do you know if it has moved out of ALPHA status since 1.22? That PR you linked was merged before v1.22.

cost-analyzer/charts/prometheus/values.yaml
AjayTripathy: I tend to agree. @nickcurie can you check if it's on the cadvisor metrics we already scrape? You can learn about that job here: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/charts/prometheus/values.yaml#L1165

cost-analyzer/charts/prometheus/values.yaml
AjayTripathy: I think it's ok to put an alpha metric in but concur that resiliency when the metric is not available is important.

cost-analyzer/charts/prometheus/values.yaml
michaelmdresser: This metric isn't a cadvisor metric, its a kubelet metric. See [docs](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#metrics-in-kubernetes): "In most cases metrics are available on `/metrics` endpoint of the HTTP server ... Note that kubelet also exposes metrics in `/metrics/cadvisor`, `/metrics/resource` and `/metrics/probes` endpoints. Those metrics do not have same lifecycle."

I'm not sure if we already scrape the standard kubelet metrics or not, just wanted to point this out so we're on the same page. The query I ran in https://github.com/kubecost/cost-analyzer-helm-chart/pull/1713#discussion_r984888548 shows which metrics endpoint this comes from.

nickcurie: Unfortunately, `/metrics/cadvisor` doesn't have any metrics on PVs or PVCs. `/metrics` does, so that's why we need to scrape this endpoint. I'm not sure we have another way to monitor PV usage.

AjayTripathy: SG, just wanted to make sure we checked.

michaelmdresser (approved):
Looks good, make sure to nuke those unnecessary relabel configs before merging. We have approval to use the alpha metric.

michaelmdresser:
Maybe just add a comment on the prom scrape for this metric that its in alpha.

PR #1733, opened 2022-10-10
add ports to cost-analyzer deployment

jessegoodier:
## Background

Port definitions for k8s deployments are optional and simply informative. When available, this allows tools like Lens to see the ports from the pods screen and forward from there. 

This was a customer request.

## What does this PR change?

Add port definitions to cost-model container. 
I realize that these ports are for the cost-analyzer-frontend. I put them on cost-model so you don't have to scroll all the way down in Lens to find the ports. 

NA

Allows certain tools to discover the ports and forward them easily

## How was this PR tested?

helm upgrade with local chart. Verified in Lens that it has the intended effect. 

NA

# Screenshot
image

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -278,7 +278,7 @@ spec:
           {{- if .Values.sigV4Proxy.role_arn }}
           - --role-arn
           - {{ .Values.sigV4Proxy.role_arn }}
-          {{- end }}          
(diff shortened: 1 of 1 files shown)


AjayTripathy (approved):

michaelmdresser:
A little late to the party here, but isn't `9090` the frontend port and `9003` the model (API) port?

jessegoodier:
> A little late to the party here, but isn't `9090` the frontend port and `9003` the model (API) port?

better late than never. will fix.

PR #1741, opened 2022-10-13
[networkcosts] Fix default loopback range to include all loopback...

LaikaN57:
… addresses

## What does this PR change?

Updates the loopback address in the `in-zone` classification to include the entire loopback range as registered in IANA.

### Authoritative Sources
IANA IPv4 Address Space Registry: https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml#:~:text=False-,127.0.0.0/8,-Loopback
IANA IPv4 Special-Purpose Address Registry: https://www.iana.org/assignments/ipv4-address-space/ipv4-address-space.xhtml#:~:text=ALLOCATED-,127/8,-IANA%20%2D%20Loopback

No.

Some systems use other loopback addresses such as `127.0.0.53` (for some DNS resolvers), `127.1.0.1`, etc. This change will now classify all addresses in the `127.0.0.0/8` range as `in-zone` by default.

N/A.

## How was this PR tested?

Not tested. I would like some help testing this. I will also try to run a similar config in the next few days and update here once that is done.

Yes.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -588,8 +588,8 @@ networkCosts:
       # In Zone contains a list of address/range that will be
       # classified as in zone.
       in-zone:
-        # Loopback
-        - "127.0.0.1"
+        # Loopback Addresses in "IANA IPv4 Special-Purpose Address Registry"
+        - "127.0.0.0/8"
         # IPv4 Link Local Address Space
         - "169.254.0.0/16"
         # Private Address Ranges in RFC-1918


cost-analyzer/values.yaml
```diff
@@ -588,8 +588,8 @@ networkCosts:
       # In Zone contains a list of address/range that will be
       # classified as in zone.
       in-zone:
-        # Loopback
-        - "127.0.0.1"
+        # Loopback Addresses in "IANA IPv4 Special-Purpose Address Registry"
```
corydolphin, line 591: To confirm, this would've incorrectly excluded this subset of loopback and would've appeared to have charged us?

corydolphin: What happens on line 608? What classification do we have now for cross zone?

LaikaN57: > To confirm, this would've incorrectly excluded this subset of loopback and would've appeared to have charged us?

Correct. It would have been excluding these addresses: `127.0.0.0` and `127.0.0.2` to `127.255.255.255`.

AjayTripathy:
Thank you so much for this PR @LaikaN57 . This LGTM but I'd like @mbolt35 to take a final review.

PR #1745, opened 2022-10-17
Update values.yaml

Tushendra:

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-197) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/charts/grafana/values.yaml
+++ b/cost-analyzer/charts/grafana/values.yaml
@@ -1,6 +1,6 @@
 rbac:
   create: true
-  pspEnabled: true
+  pspEnabled: false
   pspUseAppArmor: true
 serviceAccount:
   create: true


AjayTripathy:
Hey @Tushendra while PSPs are deprecated, this feels like a sweeping change to make without an announcement of some sort that we're removing PSP support by default in an upcoming version. You can set yours to false for now in your local install and we'll discuss getting this into v1.99 (release after next)

thomasvn:
Note: not only are PSPs now deprecated, since Kubernetes v1.25 they have been [removed](https://kubernetes.io/docs/concepts/security/pod-security-policy/).

It may be important to monitor which versions of Kubernetes EKS, GKE, and AKS are beginning to default to. I think it's important we default to `pspEnabled: false` before wide adoption of Kubernetes v1.25.

thomasvn:
Additionally, instead of making a code change to the subchart's `values.yaml`, we can make the following change to Kubecost's `values.yaml`.

- `.Values.grafana.rbac.pspEnabled: false`
- `.Values.podSecurityPolicy.enabled: false`

chipzoller:
These changes have already been made.

PR #1752, opened 2022-10-21
Soft Memory Limit Field

mbolt35:
## What does this PR change?
* Adds a `kubecostModel.softMemoryLimit` field which sets the `GOMEMLIMIT` flag for the cost-model backend. 

* Users can optionally enable impactful memory thresholds to ensure the Go runtime GC throttles at more aggressive frequencies at or approaching the soft limit. For example, examining that the cost-analyzer pod can reach `1GiB` of memory during high usage periods, but can idle as low as `250MiB`. Setting the `kubecostModel.softMemoryLimit=500MiB` may keep the pod's memory from surpassing 500MiB. 
* There is not a one-size fits all value here, and users looking to tune the parameters should be aware that lower values may reduce overall performance if setting the value too low. If users set the the `resources.requests` memory values appropriately, using the same value for `softMemoryLimit` is advisable, as this will instruct the go runtime to keep it's heap acquisition and release within the same bounds as the expectations of the pod memory use. 
* More on this environment variable can be found here: https://tip.golang.org/doc/gc-guide  

## How was this PR tested?
```
helm3 template kubecost --namespace=kubecost ./cost-analyzer -f ./cost-analyzer/values.yaml --set kubecostModel.softMemoryLimit=40GiB > temp.yaml
```

Output:
```
 - name: GOMEMLIMIT
    value: 40GiB
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -383,6 +383,10 @@ spec:
             - name: APP_CONFIGMAP_NAME
               value: {{ .Values.appConfigmapName }}
             {{- end }}
+            {{- if .Values.kubecostModel.softMemoryLimit }}
+            - name: GOMEMLIMIT
+              value: {{ .Values.kubecostModel.softMemoryLimit }}
+            {{- end }}
             {{- if .Values.assetReportConfigmapName }}
             - name: ASSET_REPORT_CONFIGMAP_NAME
               value: {{ .Values.assetReportConfigmapName }}


dwbrown2:
This LGTM, my only question is should we also include in default values file? I could go either way and we could also add later. 

Proposal is we add a section here too: https://guide.kubecost.com/hc/en-us/articles/6446286863383-Tuning-Resource-Consumption -- your PR description looks like a great start!

AjayTripathy (approved):

mbolt35:
>  I could go either way and we could also add later.

@dwbrown2 I think we can proceed incrementally here, what I wanted to avoid was dropping an arbitrary "40GiB" limit. It stuck out like a sore thumb 😄 - I think as we collect more information from users, we can form stronger opinions on how we expose. Just my take. Totally agree on the docs!

dwbrown2:
@mbolt35 despite not putting `40GiB` in values, are you actually setting that as default or leaving unset?

bstuder99:
Ok, I'm gonna take a crack at this. Currently juggling a few other PRs, but I can get a draft ready to go with the official release. This issue and you two will be tagged in the upcoming PR.

mbolt35:
> @mbolt35 despite not putting 40GiB in our default values file, are you actually setting that as default or leaving unset?

@dwbrown2 Leaving it unset for now.

bstuder99:
Docs update: https://github.com/kubecost/docs/pull/424

Issue #1758, opened 2022-10-24
Allow use of `config.prefix` in `kubecostModel.etlBucketConfigSecret` secret by bumping to Thanos 0.27


LaikaN57:
**What problem are you trying to solve?**

I am trying to have multiple clusters backup ETL data to the same S3 bucket using `kubecostModel.etlBucketConfigSecret`. However when using a secret like this:

```yaml
apiVersion: v1
kind: Secret
type: Opaque
stringData:
  object-store.yaml: |
    type: S3
    config:
      bucket: "my-bucket"
      prefix: "my-prefix"
      endpoint: "s3.amazonaws.com"
      region: "us-east-1"
      insecure: false
      signature_version2: false
      put_user_metadata:
        "X-Amz-Acl": "bucket-owner-full-control"
```

I get the following warning:

```log
0000-00-00T22:57:58.831642422Z WRN ETL: failed to create bucket storage: create S3 client: yaml: unmarshal errors:
  line 4: field prefix not found in type storage.S3Config
```

However if we look at similar documenation to the [doc referenced](https://guide.kubecost.com/hc/en-us/articles/4407601811095-ETL-Backup#:~:text=This%20configuration%20secret%20follows%20the%20same%20layout%20documented%20for%20Thanos%20here%3A%20https%3A//thanos.io/v0.21/thanos/storage.md), in these [similar docs](https://github.com/thanos-io/objstore#s3:~:text=The%20field%20prefix,which%20blocks%20came) we can see that a `config.prefix` key is mentioned. (NOTE: I think the documentation actually has an error and the `prefix` key should be under the `config` key.)

**Describe the solution you'd like**

Extend the unmarshalling functionality for the `storage.S3Config` type to be able to handle the `config.prefix` key. This may be as simple as just updating a dependency to a newer version (not sure).

**Describe alternatives you've considered**

None.

**How would users interact with this feature?**

Users would be able to include a `config.prefix` key in their `kubecostModel.etlBucketConfigSecret` secret to be able to select a custom prefix inside of a custom bucket.

LaikaN57:
I am not 100% sure if this is the correct repo top open this issue against. If there is a better place, please LMK.

LaikaN57:
...and just to confirm, I also tried moving the `prefix` key up a level (as it was in the doc example originally) and got the same warning:

```log
[...]

PR #1776, opened 2022-11-02
Upgrade Thanos subchart v0.24 -> v0.29

thomasvn:
## What does this PR change?

This change primarily bumps the Thanos image version from v0.24 -> v0.29 in our existing Thanos subchart. In the process, there were some additional changes that needed to be made:

- remove the `blockSyncConcurrency` argument for the `thanos-compact` deployment since it was [deprecated in v0.27](https://thanos.io/v0.21/thanos/changelog.md/#removed-2).

It would be good to monitor this change in nightly for an extended amount of time before potentially releasing this change in Kubecost v1.99.

No

For most users, there will be minimal impact. If they were setting up special configs for Thanos, it's possible that their configs will not be backwards compatible in this new version of Thanos.

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1758

## How was this PR tested?

Deployed to my local Thanos cluster and monitor logs for any outlier error messages.

Not yet. However I don't believe this requires an addition to https://docs.kubecost.com.

### DIFF
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -20,4 +20,4 @@ dependencies:
   - condition: global.thanos.enabled
     name: thanos
     repository: file://./charts/thanos
-    version: "~0.28.1"
+    version: "~0.29.0"
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "1.92.0"
+appVersion: "1.97.0"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: "1.92.0"
+version: "1.97.0"
 annotations:
   "artifacthub.io/links": |
     - name: Homepage
@@ -20,4 +20,4 @@ dependencies:
   - condition: global.thanos.enabled
     name: thanos
     repository: file://./charts/thanos
-    version: "~0.24.0"
+    version: "~0.28.1"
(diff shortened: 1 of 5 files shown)


cost-analyzer/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "1.92.0"
+appVersion: "1.97.0"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: "1.92.0"
+version: "1.97.0"
```
LaikaN57, line 6: Please update to `1.98.0` as this is the current version of the `develop` branch.

LaikaN57 (approved):
LGTM

cost-analyzer/Chart.yaml
michaelmdresser: Would you omit the appVersion update? I think we'll hit a merge conflict between this and https://github.com/kubecost/cost-analyzer-helm-chart/pull/1793. https://github.com/kubecost/cost-analyzer-helm-chart/pull/1793 is more tightly scoped to the appVersion update so I'd prefer to do it there.

michaelmdresser (approved):

cost-analyzer/Chart.yaml
thomasvn: Good point, will do!

thomasvn:
Before merging, I'd like to upgrade the Thanos image version from v0.28.1 to v0.29.0 which was just released a few days ago.

AjayTripathy:
Approved, let's merge then make sure we get this tested on the scale cluster. Might be a good time to transition/knowledge share this with you @Sean-Holcomb can you help Thomas get set up with it?

AjayTripathy (approved):

thomasvn:
Holding off on merging this PR since we've discovered that this change would not actually fix Issue https://github.com/kubecost/cost-analyzer-helm-chart/issues/1758

AjayTripathy:
May as well do this upgrade since the works been done even if its not a fix for 1758

PR #1789, opened 2022-11-06
fix kubecost metrics service wrong app label

zzorica:
## What does this PR change?

Teste 1.98 helm chart and serviceMonitor for metrics service targets `app=cost-analyzer` label and it should target `app=$name-metrics` label.

If using custom Prometheus with operator it cannot scrape kubecost metrics pod using serviceMonitor.

## How was this PR tested?
Installed helm chart on k8s cluster with existing prometheus operator. Also rendered helm value to test the change.

Before:
```
---
# Source: cost-analyzer/templates/kubecost-metrics-service-template.yaml
apiVersion: v1
kind: Service
metadata:
  name: kubecost-metrics
  labels:

    helm.sh/chart: cost-analyzer-1.98.0
    app: cost-analyzer

```
Now:
```
---
# Source: cost-analyzer/templates/kubecost-metrics-service-template.yaml
apiVersion: v1
kind: Service
metadata:
  name: kubecost-metrics
  labels:

    helm.sh/chart: cost-analyzer-1.98.0
    app: kubecost-metrics

```

No

### DIFF
--- a/cost-analyzer/templates/kubecost-metrics-service-template.yaml
+++ b/cost-analyzer/templates/kubecost-metrics-service-template.yaml
@@ -7,7 +7,8 @@ kind: Service
 metadata:
   name: {{ template "kubecost.kubeMetricsName" . }}
   labels:
-{{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+    {{ unset (include "cost-analyzer.commonLabels" . | fromYaml) "app" | toYaml | nindent 4 }}
+    app: {{ template "kubecost.kubeMetricsName" . }}
 {{- if (or .Values.kubecostMetrics.exporter.service.annotations $prometheusScrape) }}
   annotations:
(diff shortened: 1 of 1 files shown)


AjayTripathy:
@mbolt35 checking in-- is this the right label to use here? If not do we need a flag to flip between app:cost-analyzer and app:kubecost-metrics?

mbolt35:
> @mbolt35 checking in-- is this the right label to use here? If not do we need a flag to flip between app:cost-analyzer and app:kubecost-metrics?

We definitely want to have a different label, as the kubecost-metrics pod can be run separate from the cost-model.

speedfl:
Hello.

Any news on this one ? It break the usage of Prometheus Operator

AjayTripathy (approved):

AjayTripathy:
Sorry for the late response here. Approved + merged.

PR #1633, opened 2022-08-11
Bump the default timeout to 60s

AjayTripathy:
## What does this PR change?
Bumps the prometheus scrape timeout to 60s to avoid issues where at high-scale clusters, kubecost can take additional time to respond to prometheus scrapes.

## Does this PR rely on any other PRs?
No

## How does this PR impact users? (This is the kind of thing that goes in release notes!)
Additional stability on large scale clusters

## Links to Issues or ZD tickets this PR addresses or fixes
ZD #2354

## How was this PR tested?
Manual change of scrape timeout on a high scale cluster-- noted that the Prometheus kubecost scrape jobs turned from red to green.

## Have you made an update to documentation?
Not necessary

PR #1792, opened 2022-11-08
bump default values.yaml prometheus scrape_timeout

keithhand:
## What does this PR change?
A previous [PR](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1633) bumped the scrape_timeout in the Prometheus sub chart, but it was still set in the primary values.yaml, causing it to be overwritten. This updates the default values.yaml, as well.

Bump default Prometheus scrape timeout to 60s.


AjayTripathy:
Thanks Keith!

AjayTripathy (approved):

PR #1801, opened 2022-11-11
Update kubecost-metrics-deployment-template.yaml due to missing condition

daper:
## What does this PR change?
Fix bug caused by missing condition on kubecost-metrics-deployment-template.yaml

No

## How was this PR tested?
With `helm template`.

Not necessary.


AjayTripathy:
Thanks for this! While this scenario is generally unlikely since we often need some product configs, I support this change.

AjayTripathy (approved):

Issue #1742, opened 2022-10-13
More implicit scraping of required Prometheus metrics


keithhand:
**What problem are you trying to solve?**
If you deploy Kubecost to a cluster that has many services with the Prometheus annotation `prometheus.io/scrape: true`, our bundled Prometheus will attempt to auto-scrape these endpoints, potentially leading to false positives on the diagnostics page for "Prometheus endpoints"

**Describe the solution you'd like**
More implicit scraping of our metrics and removing the reliance on the default `prometheus.io/scrape` annotation

**Describe alternatives you've considered**
Ignore false positives, remove scraping from existing services (not possible with infrastructure)

**How would users interact with this feature?**
Install Kubecost with bundled Prometheus, and it doesn't scrape existing Prometheus metric endpoints.

thomasvn:
More specifically, this issue occurs for the [`kubernetes-service-endpoints` scrapeconfig](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/charts/prometheus/values.yaml#L1278). Prometheus will scrape all service endpoints available, and it may error out attempting to scrape API servers or monitoring stacks that require authentication.

As a potential solution, we can add to the [Kubernetes service discovery config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config) to include `__meta_kubernetes_namespace: kubecost` or (`__meta_kubernetes_endpoints_name: kubecost-prometheus-node-exporter` and `kubecost-kube-state-metrics`) to filter for only the targets Kubecost cares about.

AjayTripathy:
I support this and like the proposed solution @thomasvn , this feel like something you can pick up for v1.99? adding the label and cc-ing @Adam-Stack-PM

thomasvn:
Thanks for the feedback @dwbrown2 and @AjayTripathy! Yep this is something I can work on for v1.99!

akramwaseem:
+1, This is something we need as well. This is causing lot of duplicates on metrics as well and adding our own custom config or scrape config seems more resource intensive and have to maintain in the long run.

dwbrown2:
@thomasvn still think you'll have capacity to pick for 1.99?

One other thought: if we use `__meta_kubernetes_endpoints_name` filter we should think about network costs deamonset as well I believe. Not sure if there are others, but Nvidia's GPU utilization exporter may be one.

thomasvn:
[...]

Issue #1787, opened 2022-11-04
Reintroduce `gpuRequestAverage` and `gpuUsageAverage` to the Allocation API Schema


thomasvn:
### What problem are you trying to solve?
Currently when calling the Allocation API (example below), we are only returned `gpuCount`, `gpuHours`, `gpuCost`, and `gpuCostAdjustment`.

To get a better understanding of how the GPUs are actually used, it would be nice to add `gpuRequestAverage` and `gpuUsageAverage` to the Allocation API Schema (similar to how `cpuCoreRequestAverage` and `cpuCoreUsageAverage` exist in the schema)

```
/model/allocation?window=yesterday&accumulate=true&shareIdle=true&reconcile=true&idleByNode=true&aggregate=node
```

### Background
This feature was previously introduced in [PR #944](https://github.com/opencost/opencost/pull/944), however it needed to be rolled back in [PR #965](https://github.com/opencost/opencost/pull/965).

Given that Kubecost now has a backwards-compatible ETL [PR #624](https://github.com/kubecost/kubecost-cost-model/pull/624), this new Allocation Schema can be re-introduced.

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-209) by [Unito](https://www.unito.io)

AjayTripathy:
Just as a note, those aren't standalone; they require deploying a metric collector to run that isn't straightforward. I'm happy to bring those back, just know that getting the relevant exporters running on clusters can be a challenge. I think we can look to bring this back in an alpha state though for v1.99 and provide a guide for spinning up the gpu usage exporters. @kaelanspatel thoughts?

kaelanspatel:
Agree for reintroduce, would be nice to support. 

Like Ajay says, this was initially designed to use the `DCGM_FI_DEV_GPU_UTIL` metric from the Nvidia DCGM. Old setup document [here](https://docs.google.com/document/d/1fsbV55wTbpfWy4m9_FVETqHgtTMahfUa2w5MzV9gngc/edit). If we find some time, we might want to look at other metrics and/or the state of the DCGM, the `DCGM_FI_DEV_GPU_UTIL` metric was moved to off by default in the DCGM and I wonder if there's been an alternative since for a raw usage number.

AjayTripathy:
[...]

PR #1805, opened 2022-11-15
Adjust bundled Prometheus to scrape for only essential metrics

thomasvn:
## What does this PR change?

The `scrape_config` titled `kubernetes-service-endpoints` currently attempts to scrape all service endpoints available in the cluster which include the `prometheus.io/scrape: true` annotation

Because Kubecost only needs metrics from `kubecost-kube-state-metrics`, `kubecost-prometheus-node-exporter`, and `kubecost-network-costs` ... Prometheus is scraping more than it needs to. This results in duplicate metrics, and errors when attempting to scrape a service endpoint which it doesn't have permissions to scrape.

This PR adds a filter in the `kubernetes-service-endpoints` `scrape_config` so that it is only scraping for the metrics required by Kubecost.

prometheus

No

If users were querying any metrics from Kubecost’s bundled prometheus that are not listed here: [https://github.com/kubecost/docs/blob/main/user-metrics.md](https://github.com/kubecost/docs/blob/main/user-metrics.md), they will be affected.

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1742

## How was this PR tested?

Using the following `values.yaml`, I verified on the Prometheus server that all metrics endpoints required by Kubecost were still being scraped, while all excess metrics endpoints (e.g. `kube-dns`) were no longer scraped.

`kubectl port-forward svc/kubecost-prometheus-server 8080:80`

```yaml
# values.yaml
kube-state-metrics:
  enabled: true
nodeExporter:
  enabled: true
networkCosts:
  enabled: true
```

No

### DIFF
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -1284,6 +1284,9 @@ serverFiles:
           - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]
             action: keep
             regex: true
+          - source_labels: [__meta_kubernetes_endpoints_name]
+            action: keep
+            regex: (kubecost-kube-state-metrics|kubecost-prometheus-node-exporter|kubecost-network-costs)
           - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
             action: replace
             target_label: __scheme__


jessegoodier:
Love this. testing now.

AjayTripathy:
This is all we need for now, but as a note we need to add the DCGM exporter when we reintroduce GPU usage metrics: https://docs.google.com/document/d/1fsbV55wTbpfWy4m9_FVETqHgtTMahfUa2w5MzV9gngc/edit

AjayTripathy (approved):

AjayTripathy:
Approved, @jessegoodier @thomasvn you all feel good to merge this?

thomasvn:
Good to merge. After further review, here are some future todo items:
- Test the Nvidia GPU metric `DCGM_FI_DEV_GPU_UTIL` once the feature is reintroduced. #1787 
- Consider removing `kubernetes-service-endpoints-slow` `scrape_config` because this only applies to the annotation `prometheus.io/scrape-slow`, which is not used by Kubecost targets. Slower scrape intervals can be configured in Kubecost’s `.Values.prometheus.server.global.scrape_interval`.
- Consider removing the `kubernetes-nodes` `scrape_config`. I believe its metrics are not currently used by Kubecost (I may be wrong).

Adam-Stack-PM:
So cool @thomasvn. Thanks for pushing this forward.

jessegoodier:
> Approved, @jessegoodier @thomasvn you all feel good to merge this?

yes, working in all of my tests. much cleaner. nice work @thomasvn

PR #1810, opened 2022-11-17
Optional Heap Monitoring for Next Network Costs Release

mbolt35:
## What does this PR change?
* Adds a poor man's heap monitoring solution for the network costs pod which will log a warning and write out a heap dump if total heap memory surpasses a specific threshold


AjayTripathy (approved):

mbolt35:
v16.3 released

PR #1814, opened 2022-11-21
CPU limit on network-cost DS

JesseGoodier:
## What does this PR change?

Add default request and limits to network costs daemonset to reduce CPU usage under load. 

See discussion here: https://github.com/kubecost/docs/pull/447

- 
- 

Add default 500m CPU limit to reduce CPU usage on network cost daemonset

- 
- 

## How was this PR tested?
Helm install on test cluster, 
```
k describe ds kubecost2-network-costs |grep Limit -A4
    Limits:
      cpu:  500m
    Requests:
      cpu:     50m
      memory:  20Mi
```

https://github.com/kubecost/docs/pull/447

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
@@ -47,6 +47,13 @@ spec:
 {{- if .Values.networkCosts.resources }}
         resources:
 {{ toYaml .Values.networkCosts.resources | indent 10 }}
+{{- else }}
+        resources:
+          limits:
+            cpu: 500m
+          requests:
+            cpu: 50m
+            memory: 20Mi
 {{- end }}
         env:
         {{- if .Values.networkCosts.hostProc }}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -586,14 +586,18 @@ networkCosts:
   # Traffic Logging will enable logging the top 5 destinations for each source
   # every 30 minutes.
   trafficLogging: true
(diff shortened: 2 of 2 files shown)


AjayTripathy (approved):

dwbrown2:
Does this PR say that even if you delete the limits/requests block in values we will still set these defaults for you? If so, is that a common practice?

jessegoodier:
> Does this PR say that even if you delete the limits/requests block in values we will still set these defaults for you? If so, is that a common practice?

Good question, thanks for reviewing. 

Upon further review, the else does not need to be set in the template. 

It actually has no effect in any case I have tested.

In any case, remove the limit with:

```
networkCosts:
  resources:
    limits: null
```

I'll PR to remove else statement and add a comment in values.yaml to remove limits to reduce confusion.
Will also update docs.

Adam-Stack-PM:
@jessegoodier, if relevant could you drop the v1.99 tag on this?

llamahunter:
So, this cpu resource limits causes CPUThrottlingHigh alerts to fire for the kubecost-network-cost daemonset pods. Probably should not have a limit? It's request is only 50m, and any 'excess cpu' is divied up between competing pods based on their relative requests weights. So if some 2000m container wants to run more, only 2.5% of the excess CPU would get siphoned off to kubcost-network-cost, no?

jessegoodier:
This is a low priority, async workload. 

You can certainly weigh the cost/nuisance of that alarm with letting this workload use more CPU real-time.

If any others have thoughts, please feel free to correct me.

PR #1815, opened 2022-11-22
Update cluster-controller to v0.3.0 and fix template problems

michaelmdresser:
## What does this PR change?

- Title.

- Also fixes an error with the cluster-controller template where if `.Values.clusterController.kubescaler.defaultResizeAll` is unset, the `default false` kicks in which is a bool, not a string, so the following error occurs:
`Error: UPGRADE FAILED: cannot patch "kubecost-cluster-controller" with kind Deployment:  "" is invalid: patch: Invalid value: .......(huge json blob)..... json: cannot unmarshal bool into Go struct field EnvVar.spec.template.spec.containers.env.value of type string`

  Had to "quote" it to make sure.

- Also fixes a problem with the COST_MODEL_PATH env var. Missing the protocol caused the following error: `FTL Kubescaler setup failed error="creating a Kubescaler: recommendation service unavailable: failed to execute request: Get \"kubecost-cost-analyzer.kubecost:9090/model/savings/requestSizingV2\": unsupported protocol scheme \"kubecost-cost-analyzer.kubecost\""`

N/A (internal APIs and the bad default false hasn't been shipped yet)

## How was this PR tested?

Updated cluster using this version of the chart. Observed cluster-controller running with the correct image, endpoints (request sizer and new kubescaler) available, and logs for turndown running.

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -222,7 +222,7 @@ spec:
           value: {{ $serviceName }}.{{ .Release.Namespace }}:{{ .Values.service.targetPort | default 9090 }}/model
         {{- if .Values.clusterController.kubescaler }}
         - name: CC_KUBESCALER_DEFAULT_RESIZE_ALL
-          value: {{ .Values.clusterController.kubescaler.defaultResizeAll | default false }}
+          value: {{ .Values.clusterController.kubescaler.defaultResizeAll | default "false" | quote }}
         {{- end }}
         ports:
         - name: http-server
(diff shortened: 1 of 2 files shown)


cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -219,10 +219,10 @@ spec:
         - name: CC_LOG_LEVEL
           value: {{ .Values.clusterController.logLevel | default "info" }}
         - name: CC_KUBESCALER_COST_MODEL_PATH
-          value: {{ $serviceName }}.{{ .Release.Namespace }}:{{ .Values.service.targetPort | default 9090 }}/model
+          value: http://{{ $serviceName }}.{{ .Release.Namespace }}:{{ .Values.service.targetPort | default 9090 }}/model
```
saweber, line 222: Is there any https support on the horizon here?

michaelmdresser: I'm not sure! I don't know if we've thought about it at all -- I assume that if any users are using https they set up and manage an ingress that terminates the TLS bit.

I haven't heard any chatter about https internally, but if you think it'd be a good idea to start thinking about it I think it would make for an educational conversation!

saweber (approved):

PR #1816, opened 2022-11-22
Network costs resource limits 2

JesseGoodier:
## What does this PR change?

remove extra else statement from network-costs DS template
add comment for howto remove limit in values.yaml

## Testing:
helm template kubecost2 --namespace kubecost2 ./cost-analyzer -n kubecost --create-namespace --set networkCosts.enabled=true |grep 500m -B4
```
image: gcr.io/kubecost1/kubecost-network-costs:v16.3
        imagePullPolicy: Always
        resources:
        limits:
          cpu: 500m
```

helm template kubecost2 --namespace kubecost2 ./cost-analyzer -n kubecost --create-namespace --set networkCosts.enabled=true --set networkCosts.resources.limits=null |ag 500m -B4
```
<null>
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
@@ -45,15 +45,7 @@ spec:
         imagePullPolicy: Always
 {{- end }}
 {{- if .Values.networkCosts.resources }}
-        resources:
-{{ toYaml .Values.networkCosts.resources | indent 10 }}
-{{- else }}
-        resources:
-          limits:
-            cpu: 500m
-          requests:
-            cpu: 50m
-            memory: 20Mi
+        resources: {{- toYaml .Values.networkCosts.resources | nindent 8 }}
 {{- end }}
         env:
         {{- if .Values.networkCosts.hostProc }}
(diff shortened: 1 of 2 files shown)


AjayTripathy:
Could you also fill this out with how it was tested etc. ?

cost-analyzer/values.yaml
```diff
@@ -592,7 +592,7 @@ networkCosts:
   port: 3001
   # this daemonset can use significant resources on large clusters: https://guide.kubecost.com/hc/en-us/articles/4407595973527-Network-Traffic-Cost-Allocation
   resources:
-    limits:
+    limits: # remove the limits by setting limits: null
```
mbolt35: I think the common null value here is just an empty object: `{}`  although I'll be honest, semantically, null makes way more sense :)

cost-analyzer/templates/cost-analyzer-network-costs-template.yaml
```diff
@@ -45,15 +45,7 @@ spec:
...
-        resources:
-{{ toYaml .Values.networkCosts.resources | indent 10 }}
-{{- else }}
-        resources:
-          limits:
-            cpu: 500m
-          requests:
-            cpu: 50m
-            memory: 20Mi
+        resources: {{- toYaml .Values.networkCosts.resources | nindent 8 }}
```
mbolt35, line 48: This change looks good - thank you

cost-analyzer/values.yaml
jessegoodier: fixed. agree for consistency.

jessegoodier:
> Could you also fill this out with how it was tested etc. ?


done

AjayTripathy (approved):

Issue #1349, opened 2022-04-04
Remove thanos and grafana ingress in the bundled charts 


AjayTripathy:
The old ingresses have been disabled and it is beyond the scope of this chart to run an ingress for thanos or grafana-- those are accessible via service addresses.

dwbrown2:
Is it fair to call this low priority?

Adam-Stack-PM:
Closing based on linked issues and priority status. Re-open if needed.

Issue #1262, opened 2022-02-23
add support for installing on kubernetes 1.22 clusters


mhrabovcin:
I am trying to install on k8s 1.22 cluster. The k8s version had removed some deprecated APIs and one of those is `extensions/v1beta1` and `networking.k8s.io/v1beta1` which are used by embedded charts for `Ingress` objects.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

A simple grep for `extensions/v1beta1` shows:
```
cost-analyzer/charts/thanos/templates/store-ingress.yaml:2
cost-analyzer/charts/thanos/templates/query-frontend-ingress.yml:1
cost-analyzer/charts/thanos/templates/query-ingress.yml:2
cost-analyzer/charts/thanos/templates/bucket-ingress.yaml:1
cost-analyzer/charts/grafana/templates/ingress.yaml:1
cost-analyzer/charts/prometheus/templates/pushgateway-ingress.yaml:1
cost-analyzer/charts/prometheus/templates/_helpers.tpl:4
cost-analyzer/charts/prometheus/templates/server-vpa.yaml:1
cost-analyzer/charts/prometheus/templates/alertmanager-ingress.yaml:1
cost-analyzer/charts/prometheus/templates/server-ingress.yaml:1
```

When installing with helm I get the following error message:

```
Helm install failed: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Ingress" in version "extensions/v1beta1"
```

The configuration I've used to install chart:

```
cost-analyzer:
      fullnameOverride: "kommander-kubecost-cost-analyzer"
      global:
        prometheus:
          fqdn: http://kubecost-prometheus-server.kommander.svc.cluster.local/
          enabled: false
        thanos:
          enabled: true
          queryService: http://kommander-kubecost-thanos-query-http.kubecost.svc.cluster.local:10902/
          # The wait time before Kommander begins querying cost data for all attached clusters
          queryOffset: 5m
          query:
            deploymentAnnotations:
              secret.reloader.stakater.com/reload: kommander-kubecost-thanos-client-tls
        grafana:
          enabled: false
          # Use kommander monitoring Grafana instance
          domainName: centralized-grafana.${releaseNamespace}.svc.cluster.local
      # For Thanos Installs, Allow Higher Concurrency from Cost-Model
      # Still may require tweaking for some installs, but the thanos-query-frontend
[...]

PR #1820, opened 2022-11-23
feat: Support thanos/grafana ingress in updated apis

gracedo:
## What does this PR change?
Adds support for ingresses in the thanos and grafana subcharts that are running in more recent kubernetes versions in which the original `extensions/v1beta1` version has been deprecated and removed

Thanos and Grafana ingresses can now be deployed in newer versions of Kubernetes (1.22+) using the chart

- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1262
- https://github.com/kubecost/cost-analyzer-helm-chart/issues/1349

## How was this PR tested?
Tested installing local version of the chart with the Thanos Query ingress defined in values (on k8s 1.24)

No

### DIFF
--- a/cost-analyzer/charts/thanos/templates/query-ingress.yml
+++ b/cost-analyzer/charts/thanos/templates/query-ingress.yml
@@ -1,7 +1,15 @@
 ---
 {{ if .Values.global.thanos.enabled }}
 {{- if and .Values.query.enabled .Values.query.http.ingress.enabled }}
+{{- $apiV1 := false -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+{{- $apiV1 = true -}}
+apiVersion: networking.k8s.io/v1
+{{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
+{{ else }}
 apiVersion: extensions/v1beta1
+{{ end -}}
 kind: Ingress
 metadata:
   name: {{ include "thanos.componentname" (list $ "query") }}-http
@@ -19,6 +27,9 @@ metadata:
   annotations: {{ toYaml . | nindent 4 }}
     {{- end }}
 spec:
+{{- if .Values.query.http.ingress.className }}
+  ingressClassName: {{ .Values.ingress.query.http.className }}
+{{- end }}
   {{- if .Values.query.http.ingress.tls }}
   tls:
     {{- range .Values.query.http.ingress.tls }}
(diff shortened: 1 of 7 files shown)


gracedo:
@AjayTripathy Let me know what you think about these changes, or if you still prefer that ingress support is removed entirely from the subcharts!

gracedo:
cc @mbolt35 @dwbrown2 mind taking a look? Or is the plan to remove ingress support instead?

AjayTripathy:
Hey @gracedo apologies for the late response, been out for thanksgiving and catching up. I'll take a look.

Adam-Stack-PM:
@AjayTripathy, Friendly ping. See above.

cost-analyzer/charts/thanos/templates/query-ingress.yml
```diff
@@ -80,10 +112,20 @@ spec:
...
+              service:
+                name: {{ include "thanos.componentname" (list $ "query") }}-grpc
+                port:
+                  number: {{ $.Values.query.grpc.port }}
+          {{- else }}
           - path: {{ $.Values.query.grpc.ingress.path }}
             backend:
               serviceName: {{ include "thanos.componentname" (list $ "query") }}-grpc
-              servicePort: {{ $.Values.query.http.port }}
+              servicePort: {{ $.Values.query.grpc.port }}
```
michaelmdresser, line 127: I think this is probably a great bug fix, just want to check w/ @AjayTripathy in case it was intentional

cost-analyzer/charts/thanos/templates/store-ingress.yaml
```diff
@@ -76,10 +108,20 @@ spec:
...
+            service:
+              name: {{ include "thanos.componentname" (list $ "store") }}-grpc
+              port:
+                number: {{ $.Values.store.grpc.port }}
+        {{- else }}
         - path: {{ $.Values.store.grpc.ingress.path }}
           backend:
             serviceName: {{ include "thanos.componentname" (list $ "store") }}-grpc
-            servicePort: {{ $.Values.store.http.port }}
+            servicePort: {{ $.Values.store.grpc.port }}
```
michaelmdresser, line 123: Likewise

michaelmdresser (approved):
LGTM, thanks for the contribution!

cost-analyzer/charts/thanos/templates/query-ingress.yml
AjayTripathy: Uhmm you can't look at the UI via the GRPC port, so I'm not sure this was intentional. Not saying this was wrong, just wondering @gracedo what was the thinking here?

cost-analyzer/charts/thanos/templates/store-ingress.yaml
AjayTripathy: Ditto.

cost-analyzer/charts/thanos/templates/query-ingress.yml
gracedo: Referencing the upstream thanos helm chart, this ingress also exists https://github.com/bitnami/charts/blob/main/bitnami/thanos/templates/query/ingress-grpc.yaml likely because some use cases require being able to access the gRPC service externally (e.g. hierarchical thanos where a central querier may need to connect to various queriers running on different cluster). I imagine this ingress was originally carried over from the upstream chart when bundled into the subchart here

Adam-Stack-PM:
@nealormsbee, It will be your call if you want to cherry-pick this PR or wait until v1.100.

nealormsbee:
I'm not real Thanos-savvy. @michaelmdresser can you characterize the issue potentially fixed by the http -> grpc port change? If it isn't P0~P0.5 then I'd vote to punt to v1.100. Seems like it might be, though.

michaelmdresser:
The http -> grpc thing isn't the main thing being changed here, its actually a K8s API versioning issue on more recent versions of K8s. I think the http -> grpc thing is something we want to have clarity on before merging.

It would be amazing to have this fixed in v1.99 because otherwise ingresses for Thanos and Grafana apparently don't work, but I'm not sure if its P0.

Adam-Stack-PM:
@michaelmdresser, How risky do you think the change is?  It is Neal's call if we cherry pick but think we would all appreciate your perspective.

AjayTripathy:
This feels low risk enough to just cherrypick in-- @nealormsbee can you do that?

nealormsbee:
Will do! +1 from Adam -- I don't really know what I'm looking at here so leaning on y'all. If you feel it's sufficiently low-risk then let's go for it 🙌

PR #1821, opened 2022-11-23
Add CORS Access-Control-Allow-Methods header for /cluster

michaelmdresser:
## What does this PR change?
Adds CORS allow header for all standard REST verbs targeting cluster-controller (`/cluster`).

Not having this header was causing problems for local UI development against some cluster-controller APIs.

Adds `Access-Control-Allow-Methods` header for the `/cluster` (cluster-controller) routes.

## How was this PR tested?
Used by @wolfeaustin during local dev. Fixed his problems.


teevans (approved):
LGTM.

Issue #1210, opened 2022-01-21
Spot datafeed is being ignored


thetaiter:
**Describe the bug**
Kubecost does not seem to be able to read the data in my spot datafeed S3 bucket. I can confirm that the datafeed is dumping files to the S3 bucket. I can also confirm that Kubecost does have permissions to read the bucket. I can also confirm that we do have active spot instances in our AWS account. However in the settings in Kubecost, it just says `No spot instances detected` forever. In the logs, I see these messages every 15 minutes:
```
I0121 12:17:49.943883       1 awsprovider.go:2207] Found 0 spot data files from yesterday
I0121 12:17:49.956519       1 awsprovider.go:2216] Found 0 spot data files from today
I0121 12:17:49.956581       1 awsprovider.go:920] Spot Pricing Refresh scheduled in 15.00 minutes.
```

**To Reproduce**
1. Set up spot datafeed to write to an S3 bucket in AWS.
2. Setup the AWS values in this helm chart so that Kubecost can read it:
```
kubecostProductConfigs:
  clusterName: "<cluster-name>"
  createServiceKeySecret: true
  projectID: <aws-account-id>
  awsSpotDataRegion: us-west-2
  awsSpotDataBucket: <s3-bucket-name>
  awsServiceKeyName: <aws-acess-key-id>
  awsServiceKeyPassword: <aws-secret-access-key>
```
3. Deploy Kubecost and go to settings to verify that there is a warning that says `No spot instances detected`.
4. Spin up an EKS node group with a spot autoscaling group spanning multiple availability zones.
5. Wait for new a new datafile to be written to the spot datafeed S3 bucket.
6. Wait for Kubecost to refresh (every 15 minutes) spot data.
7. Go to settings in kubecost and it still says `No spot instances detected`.

**Expected behavior**
I expect Kubecost to detect my spot EC2 instances using the data found in the spot datafeed S3 bucket.

**Screenshots**
![image](https://user-images.githubusercontent.com/6549935/150570794-a7dbf098-bb70-4d84-8d8f-6fc9219d3bb9.png)

**Collect logs (please complete the following information):**

 - Run `helm ls` and paste the output here:
```
|> helm3 ls
NAME    	NAMESPACE	REVISION	UPDATED                            	STATUS  	CHART               	APP VERSION
kubecost	kubecost 	17      	2021-05-03 07:54:07.03152 -0500 CDT	deployed	cost-analyzer-1.79.1	1.79.1
```
[...]

PR #1830, opened 2022-12-01
Quote the projectID to avoid exponential notation in Helm

roee-landesman:
## What does this PR change?
It quotes the `projectID` string in Helm to avoid conversion to exponential notation when templating.

N/A
- 
- 

It should be a bug fix for users.

https://github.com/kubecost/cost-analyzer-helm-chart/issues/1210
- 
- 

## How was this PR tested?
N/A

N/A

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml
+++ b/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml
@@ -102,7 +102,7 @@ data:
     spotDataPrefix: "{{ .Values.kubecostProductConfigs.awsSpotDataPrefix }}"
   {{- end -}}
   {{- if .Values.kubecostProductConfigs.projectID }}
-    projectID: "{{ .Values.kubecostProductConfigs.projectID }}"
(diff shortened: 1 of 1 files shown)


michaelmdresser:
Hi @roee-landesman, thanks for the contribution! How was this tested? Even just a `helm template` to show this change in action would be great!

roee-landesman:
I did not test this - it seems to be a straightforward change. Documentation here: https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#pipelines

AjayTripathy:
I don't think this actually works. The PR as-written does this:

Error: YAML parse error on cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml: error converting YAML to JSON: yaml: line 12: did not find expected key

Pull out the quote to make the pipeline run like this:
   ` projectID: {{ .Values.kubecostProductConfigs.projectID | quote }}`
And you still get this:
```
helm template ./cost-analyzer | grep projectID
    projectID: "1.23456789e+08"
```

I think that the best thing to do is just leave a comment in the helm value that this must be a string.

PR #1885, opened 2022-12-19
Remove historical builds

nealormsbee:
## What does this PR change?
This commit removes content from the historical-builds directory and replaces it with a README explaning how historical versions of kubecost.yaml should be accessed.

It looks like we haven't been committing these for a few releases anyway (v1.92-v1.99), and all new releases will be appropriately tagged in Git, so going forward there's no use for this directory. Older, untagged historical builds will remain accessible by checking out to any commit prior to this one.

- Correct tagging going forward relies on https://github.com/kubecost/release-scripts/pull/35, and that PR's approval relies on this one's existence.

- Users who want to view the source for historical releases of `kubecost.yaml` will have to check out to commits before this one in order to see them.
- Going forward, users will be able to get these files by checking out to the appropriate release tag.

- 
-


AjayTripathy (approved):

AjayTripathy:
LGTM thanks!

michaelmdresser (approved):
Huge, thanks Neal.

PR #1898, opened 2023-01-09
add budgetcontroller scaffolding

AjayTripathy:
## What does this PR change?
Adds budget controller scaffolding/SSL public cert

KCM admissions controller PR https://github.com/kubecost/kubecost-cost-model/pull/1137

one-click turn on support for budget controllers!

## How was this PR tested?
helm install and manually check the deployment controllers

No; will need to follow with a blog post etc.

### DIFF
--- a/cost-analyzer/templates/budget-controller-service-template.yaml
+++ b/cost-analyzer/templates/budget-controller-service-template.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.budgetController -}}
+{{- if .Values.budgetController.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-server
+  namespace: kubecost
+spec:
+  selector:
+    {{ include "cost-analyzer.selectorLabels" . | nindent 4 }}
+  ports:
+    - port: 443
+      targetPort: 8443
+{{- end -}}
+{{- end -}}
--- a/cost-analyzer/templates/budget-controller-template.yaml
+++ b/cost-analyzer/templates/budget-controller-template.yaml
@@ -1,0 +1,30 @@
+{{- if .Values.budgetController -}}
+{{- if .Values.budgetController.enabled -}}
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: deployment-validation
+webhooks:
+  - name: "deployment-validation.kubecost.svc"
+    namespaceSelector:
+      matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values: [ "kubecost" ]
+    rules:
+      - operations: [ "CREATE", "UPDATE" ]
+        apiGroups: [ "apps" ]
+        apiVersions: [ "v1" ]
+        resources: [ "deployments" ]
+        scope: "Namespaced"
+    clientConfig:
+      service:
+        namespace: kubecost
+        name: webhook-server
+        path: "/validate"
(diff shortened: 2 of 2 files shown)


AjayTripathy:
@dwbrown2 had a naming suggestion that we just call this kubecostadmissioncontroller

dwbrown2:
Just an idea... thinking is that this admission controller is already being used for more than budgets.

dwbrown2:
@AjayTripathy are we able to get this functionality merged today for 1.100? I'm also assuming we want an accompanying values flag?

cc @Adam-Stack-PM @teevans for tracking

cost-analyzer/templates/budget-controller-service-template.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.budgetController -}}
```
michaelmdresser, line 1: I'm not sure what Helm best practices are here, but what do you think about combining these two templates into one file? We do that with Cluster Controller but maybe its an antipattern.

The advantage I see is that a combined file helps us avoid forgetting about files when making future changes.

cost-analyzer/templates/budget-controller-service-template.yaml
```diff
@@ -0,0 +1,15 @@
...
+{{- if .Values.budgetController.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-server
```
michaelmdresser, line 6: This is a little too generic. Maybe `RELEASENAME-admission-controller` or `RELEASENAME-budget-controller`?

cost-analyzer/templates/budget-controller-template.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.budgetController -}}
+{{- if .Values.budgetController.enabled -}}
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: deployment-validation
```
michaelmdresser, line 6: This is alright -- maybe prepend the Helm release name like we do for other things? Totally nonblocking idea.

michaelmdresser (approved):
Other than changing the `webhook-server` name, this is ready to roll. `failurePolicy: Ignore` is great.

Ajay and I discussed the `namespaceSelector` and `rules` sections in Slack, documentation explaining our choices should be coming soon.

I like the idea of changing the Helm value for enabling this to `admissionController.enabled`. I'm definitely supportive of making that change in this PR, but feel free to override if you have a strong conviction.

cost-analyzer/templates/budget-controller-template.yaml
AjayTripathy: there are some interesting rules and naming conventions that need to be hardcoded for the ssl cert to automagically work since it's generated based on a URL. let me get back to this in a v2

cost-analyzer/templates/budget-controller-service-template.yaml
AjayTripathy: there are some interesting rules and naming conventions that need to be hardcoded for the ssl cert to automagically work since it's generated based on a URL. let me get back to this in a v2

AjayTripathy:
@dwbrown2 merging to make the first rc.0 release cut. If you have other ideas on naming let me know and I can change, but it would require a new SSL cert to be generated.

cost-analyzer/templates/budget-controller-service-template.yaml
michaelmdresser: I think this is going to bite us if we don't commit to changing it as soon as possible. Names are sticky. When will there be a patch ready (even if its for 1.101)?

AjayTripathy: I think i might be able to batch all the name changes requested and get it in under the RC

michaelmdresser: That would be amazing, shoot me a PR whenever you have one

PR #1904, opened 2023-01-16
secret store CSI driver for agent key

corax15:
## What does this PR change?
A SecretProviderClass manifest can be deployed for the agent key (when deploying the chart in the kubecost-agent configuration) that can be used to retrieve the agent key from a secret store (e.g. Azure Key Vault).
In case this option is used the no agent key has to be specified explicitly and no secret is created in Kubernetes.

No.

Users have now the option to retrieve the agent key from a secret store that supports the SecretProviderClass manifest.

## How was this PR tested?
Deployed on one of our clusters and tested against Kubecost Cloud.

No. What would be the right place?

### DIFF
--- a/cost-analyzer/templates/kubecost-agent-secretprovider-template.yaml
+++ b/cost-analyzer/templates/kubecost-agent-secretprovider-template.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.agentCsi.enabled}}
+{{- if .Capabilities.APIVersions.Has "secrets-store.csi.x-k8s.io/v1" }}
+apiVersion: secrets-store.csi.x-k8s.io/v1
+{{- else }}
+apiVersion: secrets-store.csi.x-k8s.io/v1alpha1
+{{- end }}
+kind: SecretProviderClass
+metadata:
+  name: {{ .Values.agentCsi.secretProvider.name }}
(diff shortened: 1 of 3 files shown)


corax15:
Is it correct to add the label "app: {{ template "kubecost.kubeMetricsName" . }}" explicitly to the SecretProviderClass? I copied that from the deployment manifest.

jessegoodier:
Wow, thank you so much for this! 
We will review this week and get back to you as soon a possible.

teevans:
@AjayTripathy - Would you mind reviewing this week?

AjayTripathy (approved):

AjayTripathy:
Code itself lgtm , @jessegoodier any suggestions on where to document?

thomasvn:
Ran into the following error:

> Error: template: cost-analyzer/templates/kubecost-agent-secretprovider-template.yaml:1:14: executing "cost-analyzer/templates/kubecost-agent-secretprovider-template.yaml" at <.Values.agentCsi.enabled>: nil pointer evaluating interface {}.enabled

I believe we will need to set the following as a default in `values.yaml` as well:

```yaml
agentCsi:
  enabled: false
```

corax15:
@thomasvn Created #1911 
I excluded the SecretProviderClass from rendering when the agent isn't enabled in the first place according to other template files.

PR #1906, opened 2023-01-16
Set clusterAccountID in configmap

Sean-Holcomb:
## What does this PR change?
Add ClusterAccountID Value and set in config map
see https://github.com/opencost/opencost/pull/1613 for config details

* https://github.com/opencost/opencost/pull/1613

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-21) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -966,6 +966,7 @@ costEventsAudit:
 #    pod_external_label: "kubernetes_pod"
 #  grafanaURL: ""
 #  clusterName: "" # clusterName is the default context name in settings.
+#  clusterAccountID: "" # Manually set Account property for assets
 #  currencyCode: "USD" # official support for USD, AUD, BRL, CAD, CHF, CNY, DKK, EUR, GBP, INR, JPY, NOK, PLN, SEK
 #  azureBillingRegion: US # Represents 2-letter region code, e.g. West Europe = NL, Canada = CA. ref: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
 #  azureSubscriptionID: 0bd50fdf-c923-4e1e-850c-196dd3dcc5d3
--- a/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml
+++ b/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml
@@ -47,6 +47,9 @@ data:
   {{- if .Values.kubecostProductConfigs.clusterName }}
     clusterName: "{{ .Values.kubecostProductConfigs.clusterName }}"
   {{- end -}}
+  {{- if .Values.kubecostProductConfigs.clusterAccountID }}
+    clusterAccountID: "{{ .Values.kubecostProductConfigs.clusterAccountID }}"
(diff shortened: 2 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -896,6 +896,7 @@ federatedETL:
 #    pod_external_label: "kubernetes_pod"
 #  grafanaURL: ""
 #  clusterName: "" # clusterName is the default context name in settings.
+#  clusterAccount: "" # Manually set Account property for assets and allocations from this deployment
```
nikovacevic: @dwbrown2 just wanted to draw your attention to this new flag. We needed a way to enforce a single `Account` value, because there are inconsistencies cropping up across the various providers, given that we don't have a standard way to configure this. Sean can share more details, but that's the short version.

nikovacevic (approved):

cost-analyzer/values.yaml
dwbrown2: What happens if you don't set? 

Also, see any paths to instead pull this automatically from providers?

cc @kwombach12

Sean-Holcomb: AWS:
It cannot be pulled for AWS, because AWS doesn't make it available in the k8s api. Currently for AWS it is being pulled from `AthenaProjectID`, which causes issues when you have a master payer account set up for your CUR because the incorrect value gets set. If you do not set `ClusterAccount` on AWS it will still check if `AthenaProjectID` is set to support users current set ups.

Azure:
We can detect the subscriptionID, so it is fine not to set. If you do set it, it works as a manually override.

All Others:
We don't do account for other providers, including GCP which just uses ProjectID. As it stands this PR doesn't do anything on other providers, but we could make this value set account regardless of provider.

dwbrown2: I like the idea idea of having it as a consistent manual override vs different behavior across providers. Can we mark this as optional in comment if my understanding is correct that it's true?

Also, do you like `clusterAccount`  > `clusterAccountName`?

AjayTripathy: I think this is an ID, not a name.

Sean-Holcomb: The value is definitely more of an ID. I would push back against adding a suffix on it though because this value maps directly to `Asset.Properties.Account`. The intention of the name `clusterAccount` is that it would be the `Account` for the cluster you are on. Anytime you see `AccountID` as a variable name floating around the system it is a hold over from when an incorrect use of `ProjectID`, `AthenaProjectID` being the prime example and upstream source of many of these instances.

Sean-Holcomb: In terms of making it a universal manual override for all providers, I can add it to the other providers before merging. And of course we will also need a doc update, and to communicate with other teams that this is the value to use.

Sean-Holcomb: I went with clusterAccountID. based off the definitions from this document 
https://github.com/finopsfoundation/finops-open-cost-usage-spec/blob/main/specification_sheet_import.md

PR #1911, opened 2023-01-20
fixed SecretProviderClass for non-agent deploy

corax15:
## What does this PR change?
Fixed an issue with the SecretProviderClass being enabled for non-agent deployments which causes an error. The 

- #1904 

Fixes the referenced PR so non-agent deployment does not fail without explicitly disabling the agentCsi flag.

## How was this PR tested?
Local helm rendering

No.


thomasvn:
@dhuesmann89 Thanks for the fix! I like this approach more than the one I recommended in #1904 as it doesn't require adding _more_ to the default `values.yaml`. I'll test this as soon as possible.

thomasvn (approved):

PR #1912, opened 2023-01-20
Cloud Agent

mbolt35:
## What does this PR change?
* Adds CloudAgent deployment options

* Requires PRs for including the cloud-agent logic in the cost-model image, but can use custom image to avoid requirement. 

* No impact on users

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-19) by [Unito](https://www.unito.io)


mbolt35:
@teevans This would require merging all kubecost-cloud agent code into mainline before this will work for the release image, but we still have the option of overriding the image as we've done in our current deployment bash scripts.

teevans:
Plan here is to merge KCM work into main line before we merge this.

teevans (approved):

PR #1916, opened 2023-01-24
grafana updates

JesseGoodier:
## What does this PR change?

Adding:
- networkCosts metrics dashboard
- multi-cluster pod utilization dashboard
- Grafana templates for customers using their own Grafana (not in helm install code path)

Updating:
- Added serve_from_sub_path: true to values.yaml which is how we have /grafana configured.
- Update attached disk (linked in UI from /low-disk) to support multi-cluster Thanos

- No

Added dashboards for networkCosts metrics and multi-cluster pod utilization
Adding Grafana templates for customers using their own Grafana instances

- NA

## How was this PR tested?

multi-cluster environment and local helm install with base install:

```
helm upgrade kubecost ./cost-analyzer --namespace kubecost --create-namespace --set networkCosts.enabled=true
```

We could consider a blog post for networkCosts. 

Link from: https://docs.kubecost.com/install-and-configure/install/custom-grafana to here?

I'm not sure if we need anything else?


thomasvn:
@jessegoodier Looking good! Some comments:

* Do you have a test instance available to view the new dashboards you're adding in here?
* You added two new templated ConfigMaps `cost-analyzer/templates/grafana-networkcosts-metrics-template.yaml` and `cost-analyzer/templates/grafana-multi-cluster-pod-utilization-template.yaml` to instantiate your Grafana dashboards. However I think [this template](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.99/cost-analyzer/charts/grafana/templates/dashboards-json-configmap.yaml) would be sufficient? I believe it loads all JSON files in the `cost-analyzer/templates` directory, but not 100% sure.
* Did you add a new directory `cost-analyzer/grafana-templates` for templates that shouldn't be loaded by default, but are available in this repo if users want to use them?
* You have two files you're committing here that are named `multi-cluster-pod-utilization.json`. Was that intended?

jessegoodier:
@thomasvn 
I totally missed your comments yesterday, thanks for the review.

The new multi-cluster dashboard is not compatible with pod-utilization.json, which is why it is new.

Will take a look at https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.99/cost-analyzer/charts/grafana/templates/dashboards-json-configmap.yaml

sent DM about dashboard qa url

jessegoodier:
I couldn't get the template folder .json files to take. All the dashboards seem to have the configmap of their own. 

Which is fine. I'm sure we can figure it out, but super important at the moment.

I added another dashboard that works like the networkCosts with changing the aggregation. 

Still not ready for formal review. Please let me know if you have any more ideas though!

AjayTripathy:
> Which is fine. I'm sure we can figure it out, but super important at the moment.

@jessegoodier ...did you mean not super important? If it is super important lets get you help.

jessegoodier:
> > Which is fine. I'm sure we can figure it out, but super important at the moment.
> 
> @jessegoodier ...did you mean not super important? If it is super important lets get you help.

haha, yes, not. 
The automated [load](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.99/cost-analyzer/charts/grafana/templates/dashboards-json-configmap.yaml), seems possible. 
But we don't appear to be using it for any dashboard yet and I have bigger fish at the moment. Current method WAI.

jessegoodier:
> * Did you add a new directory `cost-analyzer/grafana-templates` for templates that shouldn't be loaded by default, but are available in this repo if users want to use them?
> * You have two files you're committing here that are named `multi-cluster-pod-utilization.json`. Was that intended?

@thomasvn 

For the new directory: these templates allow the user to import to a self-managed grafana and select a datasource on load.
I added a readme to help clarify.

Great point, I have now cleaned that up. I believe the kubernetes-resource-efficiency.json to be a "better" way to look at inefficient resources, allowing you to aggregate at a high level and drill in.

I intend to create some documentation on how to use it and the networkCosts dashboard.

Please review when you have a moment.

jessegoodier:
Tagged Webb and Ajay as it looks like you are the only ones that have done much with Grafana. 

Please let me know if you are good with these changes or if there is a better place for this.

thomasvn (approved):
Talked to Jesse about `.Values.grafana.grafana.ini.server.serve_from_sub_path=true`. Seems to have no effect whether it is set or not. However [this Grafana tutorial](https://grafana.com/tutorials/run-grafana-behind-a-proxy/) sets is. Looks good to me!

Issue #1629, opened 2022-08-10
Parameterize namespace values using Values.yaml file


linhlam-kc:
**What problem are you trying to solve?**
- Customer and partner report that many of our deployment templates does not have the namespace explicitly specified in the metadata. It generates issue when deploying Kubecost using their in-house deployment tool or Gitops tool like ArgoCD which has no option to omit namespace value. The template that does not have the namespace specified will be deploy into 'default' namespace.

**Describe the solution you'd like**
 - Update all of our resources templates in all of the helm charts and subcharts in consistent format with namespace is explicitly specified. Customer can specify the namespace to deploy Kubecost via a Values file

- Example logic:
```
{{- if .Values.namespaceOverride -}}
{{- .Values.namespaceOverride -}}
{{- else -}}
{{- .Release.Namespace -}}
{{- end -}}
```
**Describe alternatives you've considered**
- Rendering Helm charts to template file and overriding namespace value in post-operation using Kubect:
- Example:
```
helm template | kubectl create --namespace foo -f -
```
 **How would users interact with this feature?**
Customers can either deploy Kubecost using Helm with the `--namespace` flag or set it in values.yaml file when deploy Kubecost via their Gitops tool.

**Related issue**
https://github.com/kubecost/cost-analyzer-helm-chart/issues/1527

Adam-Stack-PM:
@linhlam-kc Thanks for this idea. To help move this forward, do you see any downsides to your proposed solution? 

Assuming we went with the current approach for a specific reason (not saying it was a good or bad) but always helpful knowing the other side of a change.

linhlam-kc:
@Adam-Stack-PM From my local testing, I have not seen the downside of the proposed solution compare to the current approach yet.

linhlam-kc:
Closing the issue as it was resolved with the PR https://github.com/kubecost/cost-analyzer-helm-chart/pull/1924

PR #1924, opened 2023-01-25
Update chart to use Helm release namespace

linhlam-kc:
Signed-off-by: Linh Lam <linh@kubecost.com>

## What does this PR change?

Update chart to use helm release namespace to support the deployment method that utilizes Helm engine to generate manifests such as Amazon EKS add-on. Without these changes, the generated manifests have no metadata.namespace, which results in deploying resources to the `default` namespace

N/A

Update chart to use helm release namespace to support the deployment method that utilizes Helm engine to generate manifests such as Amazon EKS add-on. Without these changes, the generated manifests have no metadata.namespace, which results in deploying resources to the `default` namespace

https://github.com/kubecost/cost-analyzer-helm-chart/issues/1629

## How was this PR tested?

- command: `helm template kubecost cost-analyzer/ -n kubecost > /tmp/default-namespace-test`
- Results:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kubecost-cost-analyzer
  namespace: kubecost
....
```

N/A

### DIFF
--- a/cost-analyzer/charts/grafana/templates/configmap-dashboard-provider.yaml
+++ b/cost-analyzer/charts/grafana/templates/configmap-dashboard-provider.yaml
@@ -13,6 +13,7 @@ metadata:
 {{ toYaml . | indent 4 }}
 {{- end }}
   name: {{ template "grafana.fullname" . }}-config-dashboards
+  namespace: {{ .Release.Namespace }}
 data:
   provider.yaml: |-
     apiVersion: 1
--- a/cost-analyzer/charts/grafana/templates/configmap.yaml
+++ b/cost-analyzer/charts/grafana/templates/configmap.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "grafana.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "grafana.name" . }}
     chart: {{ template "grafana.chart" . }}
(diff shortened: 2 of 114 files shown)


jessegoodier:
Wow, this is huge, but looks low-risk. 

Can you think of any issues that might cause with upgrades?

AjayTripathy (approved):

AjayTripathy:
I don't think this causes any issues on upgrades, or at least, can't think of any path that this would affect. Seems strictly positive to me.

linhlam-kc:
I tested the upgrade path and it didn't cause any issues. This one has been discussing for along time and many other Opensource charts like KSM and Prometheus have been adopting it. A few templates inside our chart had metadata.namespace for a long time, so I think it should not cause any issues.

thomasvn:
@linhlam-kc How does the value for `{{ .Release.Namespace }}` get set by Helm? Would it be an issue if deploying Kubecost to a namespace not strictly titled `kubecost`?

linhlam-kc:
> @linhlam-kc How does the value for `{{ .Release.Namespace }}` get set by Helm? Would it be an issue if deploying Kubecost to a namespace not strictly titled `kubecost`?

@thomasvn `{{ .Release.Namespace }}` will take the value from `-n` or `--namespace` flag from the Helm commands like `install`,`upgrade`, or `template`. The limitation here is the `helm template` which is usually used for generating the flat manifests. If there are no `metadata.namespace` defined in the templates, `helm template` doesn't add the namespace to metadata of resources which results in deploying resources into `default` namespace. It cause the issue to some tools like [Spinnaker](https://spinnaker.io/docs/guides/user/kubernetes-v2/deploy-helm/) or Amazon EKS add-on.

linhlam-kc:
Adding P0, v1.100 after discussing with Adam as it is required for our integration with Amazon EKS add-on.

PR #1925, opened 2023-01-26
Add ALLOCATION_NODE_LABEL_ENABLED and ALLOCATION_NODE_LABEL_INCLUDE_LIST env vars

nikovacevic:
Required for https://github.com/opencost/opencost/pull/1618

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -714,6 +714,18 @@ spec:
               value: {{ (quote .Values.kubecostModel.etlAssetReconciliationEnabled) | default (quote true) }}
             - name: ETL_USE_UNBLENDED_COST
               value: {{ (quote .Values.kubecostModel.etlUseUnblendedClost) | default (quote false) }}
+            {{- if .Values.kubecostModel }}
+            {{- if .Values.kubecostModel.allocation }}
+            {{- if .Values.kubecostModel.allocation.nodeLabels }}
+            {{- with .Values.kubecostModel.allocation.nodeLabels }}
+            - name: ALLOCATION_NODE_LABELS_ENABLED
+              value: {{ (quote .enabled) | default (quote true) }}
+            - name: ALLOCATION_NODE_LABELS_INCLUDE_LIST
+              value: {{ (quote .includeList) }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
             {{- with .Values.kubecostModel.cloudCost }}
             - name: CLOUD_COST_ENABLED
               value: {{ (quote .enabled) | default (quote false) }}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -344,6 +344,14 @@ kubecostModel:
   # For deploying kubecost in a cluster that does not self-monitor
   etlReadOnlyMode: false
 
+  allocation:
+    # Enables or disables adding node labels to allocation data (i.e. workloads).
+    # Defaults to "true" and starts with a sensible includeList for basics like
+    # topology (e.g. zone, region) and instance type labels.
+    # nodeLabels:
+    #   enabled: true
+    #   includeList: "node.kubernetes.io/instance-type,topology.kubernetes.io/region,topology.kubernetes.io/zone"
+
   # Enables or disables the ContainerStats pipeline, used for quantile-based
   # queries like for request sizing recommendations.
   # ContainerStats provides support for quantile-based request right-sizing


dwbrown2:
Should this default list be moved to our values file? cc @kwombach12

kwombach12:
Absolutely! Is this the 'values' file you're referring to: https://github.com/kubecost/cost-analyzer-helm-chart/blob/master/README.md#config-options? @dwbrown2

nikovacevic:
Yep. Just putting it here until we have a name for it, within `values.yaml` -- what do we think?

```yaml
kubecostModel:
  allocation:
    nodeLabels:
      enabled: true
      whiteList: "label1,label2,etc."
```

kwombach12:
I think the name: `nodeLabels.whitelist` would make sense. Users could pass a list of the labels they want access to

michaelmdresser:
Not blocking, but if its not too late, perhaps we could call it `allowlist` in the allow/deny fashion?

nikovacevic:
Doesn't make a difference to me. In fact, I think Sean once called a Cloud Cost thing an "include list" -- do you have a preference @kwombach12 @dwbrown2 ?

kwombach12:
@nikovacevic I like `includeList`. People have been moving away from white/black list for a variety of reasons so that makes a lot of sense to me.

saweber (approved):

dwbrown2:
@nikovacevic aiming for 1.100 with this?

nikovacevic:
@dwbrown2 yep!

kc-adawson:
@dwbrown2 @nikovacevic @kwombach12 Hey guys, checking on the status of this one - Looks like it merged, is this now available as of 1.100.1?

kwombach12:
@kc-adawson I believe it is! @nikovacevic correct me if im wrong

kc-adawson:
@kwombach12 Confirmired this works on 1.100.1. 



Last question, if the includeList is not set, will it collect all node labels by default?

kwombach12:
@kc-adawson I don't think it collects all node labels by default. By default, I believe we collect these labels:
`node.kubernetes.io/instance-type`, `topology.kubernetes.io/region`, `topology.kubernetes.io/zone`

nikovacevic:
Kai is correct. It's deployed and there is a small "include list" that can be replaced in helm configuration.

kc-adawson:
@nikovacevic @kwombach12 We have this deployed but the user is only seeing it working in /allocations and not anywhere else - When they go to filter Request Rightsizing by the same label, the page does not load anything. Would these labels be expected to work within Request Rightsizing? If not this may be a bug.

nikovacevic:
Request Sizing uses the Allocation API, so if it's in Allocation it should be good-to-go for Request Sizing without any other work. And IIRC @srpomeroy even tested that exact use case when the feature had been merged, and was able to get it working.

Do you have a repro we can see @kc-adawson?

kc-adawson:
@nikovacevic Repro here - http://acdc9ca48678c4e938348860722a922f-1516114798.us-east-2.elb.amazonaws.com:9090/request-sizing?window=7d&algorithmCPU=max&algorithmRAM=max&qCPU=0.95&qRAM=0.95&targetCPUUtilization=0.8&targetRAMUtilization=0.8&filter=W3sicHJvcGVydHkiOiJsYWJlbCIsInZhbHVlIjoiYmV0YS5rdWJlcm5ldGVzLmlvL29zIiwiZXF1YWxpdHkiOiI6In1d&profile=custom

Using "beta.kubernetes.io/os" which works in Allocation:

http://acdc9ca48678c4e938348860722a922f-1516114798.us-east-2.elb.amazonaws.com:9090/allocations?agg=label%3Abeta.kubernetes.io%2Fos

kc-adawson:
@nikovacevic Hey Niko wanted to ping on this - I have updated the repro above to v1.100.2 and still see the filtering issue on the Rightsizing UI

kc-adawson:
@nikovacevic So I realized the API returns the following when I use that nodeLabel, it seems to not auto-detect the containers that are related to the node label so theres nothing to show in the list:

`curl -G -d 'algorithmCPU=max' -d 'algorithmRAM=max' -d 'targetCPUUtilization=0.8' -d 'targetRAMUtilization=0.8' -d 'window=3d' --data-urlencode 'filter=label[beta.kubernetes.io/os]:"linux"' http://acdc9ca48678c4e938348860722a922f-1516114798.us-east-2.elb.amazonaws.com:9090/model/savings/requestSizingV2`

Response: Recommending requests from data: CPU algorithm set to max but no max data is available

kc-adawson:
@nikovacevic @kwombach12 @dwbrown2 Another thing here - I added "beta.kubernetes.io/os" to the Product label group instead of a directly added label. This allows me to see the containers associated with "beta.kubernetes.io/os:linux"

http://acdc9ca48678c4e938348860722a922f-1516114798.us-east-2.elb.amazonaws.com:9090/allocations?agg=product

Kapture 2023-02-16 at 09 23 27

If I had the ability to filter by "Product" in Rightsizing, I would be able to see these associated containers with the node Label

 example on API - 'filter=product:"linux"' (this is not currently supported in filtersV2)
 
 Mockup:

kc-adawson:
Right now the ask is for some kind of filter on rightsizing to include these new "nodeLabels" in features

I have a "nodeLabel" that is getting collected on this sandbox called "beta.kubernetes.io/ios". 

image

I can add it directly to allocation as a label, but then it does not show me the associated containers.

To get those, I've added it to the "Product" label map.



You can see my one value here, "Linux" when I sort by product

image

When I drill into details, we do know which containers are associated with this "linux" tag due to being in the "Product" group. 



We have no option to sort by product in Rightsizing, so I was trying to filter by the label of "beta.kubernetes.io/os=linux" for rightsizing, or even containers with tag key "beta.kubernetes.io/os" and any value. I also tried this with a simple app:kubecost: Doing this seems to crash the page with the following errors:

image

I think the "Label" filter anywhere in the product (rightsizing/allocation/alerts) should allow people to filter by a combo of any "label" we collect - whether its a Label, nodeLabel, or Annotation, they should all be available as a "label"

kc-adawson:
Simplified outstanding issue into this new issue - https://github.com/kubecost/cost-analyzer-frontend/issues/1728. Boiled it down to 2 API calls.

PR #1934, opened 2023-01-31
Ajay tripathy bc scaffolding

AjayTripathy:
## What does this PR change?
Rename files + controller name and add values yaml

No

Use a more clear naming convention

## How was this PR tested?
Manually deployed files and saw results of admissions controller

No;todo


cost-analyzer/templates/kubecost-admission-controller-service-template.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.kubecostAdmissionController -}}
+{{- if .Values.kubecostAdmissionController.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-server
```
michaelmdresser, line 6: I assume we'll be revisiting this name in a future PR?

michaelmdresser (approved):
As far as I can tell, this is just moving the templates enabled by `budgetController.enabled=true` to templates enabled by `kubecostAdmissionController.enabled=true`. LGTM on that front.

Do we have any plans to remove the old templates?

cost-analyzer/templates/kubecost-admission-controller-service-template.yaml
AjayTripathy: this is baked into the SSL key. I'd have to regenerate everything...so we can revisit in a future PR.

AjayTripathy:
> As far as I can tell, this is just moving the templates enabled by `budgetController.enabled=true` to templates enabled by `kubecostAdmissionController.enabled=true`. LGTM on that front.
> 
> Do we have any plans to remove the old templates?

I don't see them in the source tree.. am I missing something?

michaelmdresser:
> I don't see them in the source tree.. am I missing something?

https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/budget-controller-template.yaml

AjayTripathy:
https://github.com/kubecost/cost-analyzer-helm-chart/tree/AjayTripathy-bc-scaffolding/cost-analyzer/templates

Hahaha my bad. I guess I need to commit the git rm and not just move them here^

AjayTripathy:
https://github.com/kubecost/cost-analyzer-helm-chart/pull/1935

Issue #1910, opened 2023-01-19
Installation of Kubecost on Kubernetes 1.25 fails when installing without Helm


Adam-Stack-PM:
Installation of Kubecost on Kubernetes 1.25 still fails when installing without Helm:

Proposed solution from @AjayTripathy:  Generate that file from a template where we assume we're on helm less than 1.25. 

Linked Issues and context: https://github.com/kubecost/cost-analyzer-helm-chart/issues/1773#issuecomment-1397618832

dwbrown2:
@AjayTripathy is this still relevent?

AjayTripathy:
Yes, this is still relevant. 

We probably need to set 
```  -a, --api-versions strings                       Kubernetes api versions used for Capabilities.APIVersions```

See https://helm.sh/docs/helm/helm_template/

when we run helm template in our buildscripts to generate the unbundled yaml file:
https://github.com/kubecost/release-scripts/blob/2674fbd636e8ddee627373ea2bdac55fdeb6097d/create_release_tags.py#L182

A small release fix like this would normally fall to the buildmaster, maybe you can help triage @teevans  ?

**Comment by dwbrown2** [comment]
How common is it for users to install without helm at this point?

cc @kwombach12

**Comment by AjayTripathy** [comment]
We have at least one from #1773  :)

I doubt its super common so we perhaps can mark as p1

**Comment by zioproto** [comment]
> How common is it for users to install without helm at this point?
> 
> cc @kwombach12

Hello, I am the person that raised the issue about the installation without Helm still failing in #1773 

I discovered this problem because I am doing a content refresh of this Microsoft documentation page, that provides the option to install Kubecost on AKS without Helm:

https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/scenarios/app-platform/aks/cost-governance-with-kubecost

Deprecating the installation without Helm is a possible solution. But if the Kubecost project wants to support the installation by Kubernetes manifest files, then this issue should be fixed.

Please clarify if long term you want to support only the Helm installation method, so I can refresh the Microsoft docs with the correct information.

Thanks

**Comment by teevans** [comment]
@zioproto - We'll get this added to our tracking and try to get a fix out as quick as we can!

**Comment by michaelmdresser** [comment]
Repro with error output:

```
k3d cluster create --image rancher/k3s:v1.25.3-rc3-k3s1 1.25
...
kubectl create namespace kubecost
...
[...]

PR #1937, opened 2023-01-31
Re-Templated kubecost.yaml to support 1.25+ clusters.

teevans:
## What does this PR change?

Retemplates our kubecost.yaml to support V1.25+ versions of K8s

No

Allows them to install via the generated manifests instead of helm on clusters V1.25 and above.

[Issue](https://github.com/kubecost/cost-analyzer-helm-chart/issues/1910)

## How was this PR tested?
Tested with raw file committed:
image

## Command Executed

```bash
helm template kubecost ./cost-analyzer --namespace kubecost > kubecost.yaml
```

### DIFF
--- a/kubecost.yaml
+++ b/kubecost.yaml
@@ -1,72 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  labels:
-    kubernetes.io/metadata.name: kubecost
-  name: kubecost
----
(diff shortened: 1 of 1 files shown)


michaelmdresser (approved):
Would you please make a note of what command you ran?

> Also, what are your current thoughts on how to keep this file up to date?

teevans:
Great question - lots of ideas here. probably need to have a more thorough review and have a good permanent url for the file so we can add it to the automation scripts.

PR #1958, opened 2023-02-07
Added namespace configuration for network cost exporter

dr460neye:
## What does this PR change?
Adds the option to configure a namespace for the network cost pod

At the moment, only release namespace was supported. This is not helpful for more complex product setups, in specific as prometheus is not scanning every namespace for monitors by default.

## How was this PR tested?
Only local rendering was tested. A review is required in relation to the policy changes.

Added value to the README

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-244) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -8,49 +8,50 @@ While Helm is the [recommended install path](http://kubecost.com/install), these
 <a name="config-options"></a><br/>
 The following table lists the commonly used configurable parameters of the Kubecost Helm chart and their default values.
 
-Parameter | Description | Default
---------- | ----------- | -------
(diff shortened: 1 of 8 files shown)


thomasvn:
@dr460neye Thanks for the PR! I'm curious, how are you currently deploying the network-costs daemonset? If done through Helm and [`.values.networkCosts`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.100/cost-analyzer/values.yaml#L595-L597), wouldn't the daemonset always deploy to `{{ Release.Namespace }}`?

AjayTripathy:
I think the author of the PR is saying they want to deploy the network costs pod to a namespace other than the one kubecost runs in? I think that's something we can support.

chipzoller:
What is the status with this one? Conflicts need to first be resolved.

PR #1970, opened 2023-02-09
Ajay tripathy update sslkey

AjayTripathy:
## What does this PR change?
Update SSL Key to not expire for 2y. Also remove/fix namespace selector so that the scope is truly cluster level

https://github.com/kubecost/cost-analyzer-helm-chart/pull/1970

Easy demo version of the kubecost budget controller-- the previous one shipped a 30d expiry ssl key

## How was this PR tested?
Manually, new ssl key deployed 

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-256) by [Unito](https://www.unito.io)


cost-analyzer/templates/kubecost-admission-controller-template.yaml
```diff
@@ -7,11 +7,6 @@ metadata:
 webhooks:
   - name: "kubecost-deployment-validation.kubecost.svc"
     failurePolicy: Ignore
-    namespaceSelector:
-      matchExpressions:
-        - key: kubernetes.io/metadata.name
-          operator: In
-          values: [ "kubecost" ]
```
michaelmdresser: Woah. I thought we were intentionally tightly scoping it to avoid security holes and then we were going to provide guidance on manual cert generation for users who want to increase the scope. What changed?

AjayTripathy: Just general feedback/ confusion from people deploying to other namespaces and not seeing data. But I can revert and we can discuss later.

michaelmdresser (approved):

Issue #1975, opened 2023-02-10
Kubecost no longer supports BYO prometheus-node-exporter or kube-state-metrics


aaroniscode:
**Describe the bug**  

With the introduction of PR #1805 customers are missing metrics when they have existing prometheus-node-exporter and kube-state-metrics and follow the [documentation](https://docs.kubecost.com/install-and-configure/install/custom-prom#disable-node-exporter-and-kube-state-metrics-recommended) which says:

### Disable node-exporter and kube-state-metrics (recommended)

If you have node-exporter and/or KSM running on your cluster, follow this step to disable the Kubecost included versions. Additional detail on [KSM requiments](/ksm-metrics.md).

> **Note**: In contrast to our recommendation above, we do recommend disabling the Kubecost's node-exporter and kube-state-metrics if you already have them running in your cluster.

  ```sh
  helm upgrade --install kubecost \
    --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer \
    --namespace kubecost --create-namespace \
    --set prometheus.nodeExporter.enabled=false \
    --set prometheus.serviceAccounts.nodeExporter.create=false \
    --set prometheus.kubeStateMetrics.enabled=false
  ```

But when you follow the documentation above, kubecost's Prometheus won't scrape the existing prometheus-node-exporter and kube-state-metrics due to:

```
          - source_labels: [__meta_kubernetes_endpoints_name]
            action: keep
            regex: (kubecost-kube-state-metrics|kubecost-prometheus-node-exporter|kubecost-network-costs)
```

Code link [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/eb5010d956aae3a764a9f943c371b3f9e6cda850/cost-analyzer/charts/prometheus/values.yaml#L1287) Typically installs have names like `kube-state-metrics` and `prometheus-node-exporter`. Perhaps we want to add to `kube-state-metrics|prometheus-node-exporter` to the regex?  Like:

```
          - source_labels: [__meta_kubernetes_endpoints_name]
            action: keep
            regex: (kubecost-kube-state-metrics|kubecost-prometheus-node-exporter|kubecost-network-costs|kube-state-metrics|prometheus-node-exporter)
```

[...]

PR #1982, opened 2023-02-14
fix bring your own prometheus-node-exporter and kube-state-metrics

aaroniscode:
## What does this PR change?

updates the scrape config job `kubernetes-service-endpoints`

No

Bug fix

- #1975 

## How was this PR tested?
tested with `helm template` and installed. Both with kubecost included prometheus-node-exporter & kube-state-metrics and then again with BYO prometheus-node-exporter & kube-state-metrics 

┆Issue is synchronized with this [Jira Task](https://kubecost.atlassian.net/browse/GIB-267) by [Unito](https://www.unito.io)

### DIFF
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -1289,7 +1289,7 @@ serverFiles:
             regex: true
           - source_labels: [__meta_kubernetes_endpoints_name]
             action: keep
-            regex: (kubecost-kube-state-metrics|kubecost-prometheus-node-exporter|kubecost-network-costs)
+            regex: (.*kube-state-metrics|.*prometheus-node-exporter|kubecost-network-costs)
           - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scheme]
             action: replace
             target_label: __scheme__


teevans:
@nikovacevic - Can you have someone on your team review?

nikovacevic:
This looks fine to me. I'll just ask for a second opinion from @AjayTripathy and @mbolt35 if they have time.

thomasvn:
Just tested in my own environment. Working well for me! Prometheus is still scraping `kubecost-kube-state-metrics` as well as any other KSM service endpoints such as `kubecost-custom2-kube-state-metrics`.

thomasvn:
Tested this with users and it seemed to work for them as well! cc @keithhand

thomasvn (approved):

keithhand:
Overall, this LGTM. I am curious if implementing this change is almost reversing the [original purpose](https://github.com/kubecost/cost-analyzer-helm-chart/pull/1805) it was added in the first place, though. Are we not effectively reversing the "allow list" with this wild card?

aaroniscode:
@keithhand I _think_ the original purpose of the PR you mention was to stop monitoring other application metrics and only include Kubecost related metrics. The wildcard only allows bring your own Prometheus Node Exporter and kube-state-metrics. As the wildcard is only for those matching services that end with those service names.

thomasvn:
Right, the original purpose was to reduce Kubecost's Prometheus from scraping _all_ service endpoints with `prometheus.io/scrape: true`. This change still limits the service endpoints which are being scraped, but would scrape multiple nodeExporter and kubeStateMetric if there are multiple deployed. However, Kubecost is [resilient to that redundancy](https://docs.kubecost.com/architecture/ksm-metrics).

PR #1998, opened 2023-02-24
Update Network Costs to v16.5

mbolt35:
## What does this PR change?
* Updates `network-costs` version to `v16.5`:
    * Adds an internet filter to the configuration, allow it to be the first operation checked during filtering.

* Adds an internet filter to the configuration, allow it to be the first operation checked during filtering.

# Tickets Addressed
* https://github.com/kubecost/kubecost-cost-model/issues/1095

## How was this PR tested?
Live environment 

* Inline documentation on the field in configuration stubs. 
```
# Internet contains a list of address/range that will be
# classified as internet traffic. This is synonymous with traffic
# that cannot be classified within the cluster. 
# NOTE: Internet classification filters are executed _after_ 
# NOTE: direct-classification, but before in-zone, in-region,
# NOTE: and cross-region. 
```

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -663,6 +663,14 @@ networkCosts:
       # classified as non-internet egress from one region to another.
       cross-region: []
 
+      # Internet contains a list of address/range that will be
+      # classified as internet traffic. This is synonymous with traffic
+      # that cannot be classified within the cluster. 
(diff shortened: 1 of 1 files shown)


AjayTripathy (approved):

JesseGoodier (approved):
please merge!

PR #2018, opened 2023-03-02
Change registry url to registry.k8s.io

surajnarwade:
## What does this PR change?

as k8s.gcr.io will be frozen and will need to replace it with `registry.k8s.io`

Ref: https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/

No

there is no direct impact on the users.

- 
-


thomasvn:
Thanks for the PR @surajnarwade!

I've pulled both images and have validated they are the same by checking the hash:

* `k8s.gcr.io/kube-state-metrics/kube-state-metrics:v1.9.8`
* `registry.k8s.io/kube-state-metrics/kube-state-metrics:v1.9.8`

Looks good to me!

thomasvn (approved):

PR #2019, opened 2023-03-02
minor fixes for hosted agent

JesseGoodier:
## What does this PR change?

This only impacts users that are applying this file during helm installs of the hosted-agent. It is not in any other install path.

## Fixes 

`Failed to get upgrade info: template: cost-analyzer/templates/kubecost-metrics-deployment-template.yaml:47:24: executing "cost-analyzer/templates/kubecost-metrics-deployment-template.yaml" at <.Values.agentCsi.enabled>: nil pointer evaluating interface {}.enabled
`

and 

Prometheus targets broken after https://github.com/kubecost/cost-analyzer-helm-chart/pull/1805
1805 is a good cange. this workaround for hosted agents is simple enough. 

NA

Only hosted kubecost

## How was this PR tested?

Multiple customers using directly from this branch

NA

### DIFF
--- a/cost-analyzer/values-agent.yaml
+++ b/cost-analyzer/values-agent.yaml
@@ -73,7 +73,7 @@ prometheus:
     enableAdminApi: true
     sidecarContainers:
     - name: thanos-sidecar
-      image: thanosio/thanos:v0.22.0
+      image: thanosio/thanos:v0.29.0
       securityContext:
         runAsNonRoot: true
         runAsUser: 1001
--- a/cost-analyzer/values-agent.yaml
+++ b/cost-analyzer/values-agent.yaml
@@ -1,5 +1,14 @@
 # Kubecost running as an Agent is designed for external hosting. The current setup deploys a
 # kubecost-agent pod, low data retention prometheus server + thanos sidecar, and node-exporter.
+networkCosts:
+  enabled: false
+  # config:
+  #   services:
+  #     amazon-web-services: true
+  #     google-cloud-services: true
+  #     azure-cloud-services: true
+
+
 global:
   thanos:
     enabled: false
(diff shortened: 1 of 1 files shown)


jessegoodier:
@thomasvn just tagged you because you were the author of 1805, maybe you have other ideas for the prometheus part?

Either way though, this is a minor change.

thomasvn:
@jessegoodier 

* Any reason why your `extraScrapeConfigs` uses service discovery based on pod? Instead of default service discovery [based on DNS](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.100/cost-analyzer/values.yaml#L490-L494)?
* No issue with bumping the `thanos-sidecar` version!
* What's the reasoning behind disabling nodeExporter?

jessegoodier:
@thomasvn nodeExporter is useless on multi-cluster.
only used on reserved instances, which is local cluster only. See https://docs.google.com/spreadsheets/d/1JPLCRIcQTrxySgCHTGqkV4X7Wt5z_fRgnfR8B7j2lWw/edit#gid=0

please correct me if you know/find otherwise.

Researching your link now, thanks!

jessegoodier:
> @jessegoodier
> 
> * Any reason why your `extraScrapeConfigs` uses service discovery based on pod? Instead of default service discovery [based on DNS](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.100/cost-analyzer/values.yaml#L490-L494)?
> * No issue with bumping the `thanos-sidecar` version!
> * What's the reasoning behind disabling nodeExporter?

updated with dns configuration. I don't think we have a template name for the service, 

What do you think now?

thomasvn:
Looks good!

I think last question. When deploying the `kubecost-agent` is its service name (i.e. `kubectl get svc`) `kubecost-agent-agent`?

jessegoodier:
> Looks good!
> 
> I think last question. When deploying the `kubecost-agent` is its service name (i.e. `kubectl get svc`) `kubecost-agent-agent`?

yep. tested on a clean cluster to verify it picked up the target and was 1/1

thomasvn (approved):

PR #2030, opened 2023-03-06
Set up helm param for Cost Audit Event pipeline

biancaburtoiu:
## What does this PR change?
This PR introduces a `.Values.costEventsAudit.enabled` helm value linked to an `COST_EVENTS_AUDIT_ENABLED` environment variable. This enables/disables the cost event audit pipeline. 

KCM PR: https://github.com/kubecost/kubecost-cost-model/pull/1273

Users can control whether the cost event audit pipeline is enabled/disabled via this control. Particularly useful for large scale customers where the impact of the Kubernetes API watchers can be unexpected.

N/A

## How was this PR tested?
Local helm template and upgrade in local cluster, verifying that if value is set in values.yaml, the same boolean is reflected in the container envvar.

Public docs should include this note. PR TBD.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -857,6 +857,13 @@ federatedETL:
 kubecostAdmissionController:
   enabled: false
 
+# Enables or disables the Cost Event Audit pipeline, which tracks recent changes at cluster level
+# and provides an estimated cost impact via the Kubecost Predict API.
+#
+# It is disabled by default to avoid problems in high-scale environments.
+costEventsAudit:
+  enabled: false
+
 # readonly: false # disable updates to kubecost from the frontend UI and via POST request
 
 # These configs can also be set from the Settings page in the Kubecost product UI
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -929,6 +929,10 @@ spec:
             - name: GOOGLE_ANALYTICS_TAG
               value: {{ .Values.reporting.googleAnalyticsTag }}
             {{- end }}
+            {{- if .Values.costEventsAudit }}
+            - name: COST_EVENTS_AUDIT_ENABLED
+              value: {{ (quote .Values.costEventsAudit.enabled) | default (quote false) }}
+            {{- end }}
             {{- /*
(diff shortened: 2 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -857,6 +857,13 @@ federatedETL:
 kubecostAdmissionController:
   enabled: false
 
+# Enables or disables the Cost Event Audit pipeline, which tracks recent changes at cluster level
+# and provides an estimated cost impact via the Kubecost Predict API.
+#
+# It is disabled by default to avoid problems in high-scale environments.
+costEventsAudit:
```
teevans, line 864: @dwbrown2 / @kwombach12 - What are your thoughts here? Maybe enabled by default and disabled if it's a problem?

dwbrown2: Yeah, would propose enabled by default unless there are concerns uncovered during testing.

teevans: @biancaburtoiu - We may come back here and change this, but let's go ahead and set this to enabled. We'll do our diligence in scale testing this week. If we find anything concerning, we'll revert to this.

biancaburtoiu: I've just converted it to default `true`, thanks for the feedback!

kwombach12: +1 to enabling by default

AjayTripathy: I do not feel comfortable enabling this by default. We don't have the ability to test this at scale right now effectively since our simulations all scale to number of sets in prometheus, not churn and availability of the kubernetes API. What's more, we should also update this where the env variable is consumed to be a consistent default with what we pick, which I strongly believe should be "false" at this point.

If this doesn't use the kubernetes API via new watchers or saving data out of new watchers, please let me know @biancaburtoiu

dwbrown2: @AjayTripathy @teevans @biancaburtoiu can you sync to see if we can come up with a scale testing plan you're comfortable with?

AjayTripathy: we cannot come up with a scale testing plan in time for 101. Will need some time for scale testing plan. Currently sized at 1wk

teevans (approved):
LGTM

PR #2038, opened 2023-03-08
Add primaryClusterID to federator configuration

kaelanspatel:
## What does this PR change?
Adds `federator.primaryClusterID` to federated ETL configs, so that the federator can avoid overwriting extant merged data on the primary.

- https://github.com/kubecost/kubecost-cost-model/pull/1287
- 

See linked KCCM PR

- 
- 

## How was this PR tested?
See linked KCCM PR

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -853,6 +853,8 @@ federatedETL:
     # federator.clusters is an optional whitelist of clusters by cluster id.
     # If not set, the federator will attempt to federated all clusters pushing to the federated storage.
     clusters: []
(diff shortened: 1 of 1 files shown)


AjayTripathy:
Fine for now, but Is there no way to detect what the primary cluster is?

AjayTripathy (approved):

kaelanspatel:
> Fine for now, but Is there no way to detect what the primary cluster is?

I don't think so, unfortunately. There's nothing that distinguishes the primary from the others w.r.t. the federator.

PR #2079, opened 2023-03-23
Add NodeExporter dnsPolicy option.

kahirokunn:
## What does this PR change?

I have taken the actions described in the following document.

> Pods running with hostNetwork, you should explicitly set its DNS policy to "ClusterFirstWithHostNet".

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

- 
- 

- 
- 

## How was this PR tested?

on local machine

yes

### DIFF
--- a/cost-analyzer/charts/prometheus/README.md
+++ b/cost-analyzer/charts/prometheus/README.md
@@ -206,6 +206,7 @@ Parameter | Description | Default
 `initChownData.resources` | init-chown-data pod resource requests & limits | `{}`
 `kube-state-metrics.disabled` | If false, create kube-state-metrics sub-chart, see the [kube-state-metrics chart for configuration options](https://github.com/helm/charts/tree/master/stable/kube-state-metrics) | `false`
 `nodeExporter.enabled` | If true, create node-exporter | `true`
+`nodeExporter.dnsPolicy` | node-exporter dns policy | `ClusterFirstWithHostNet`
 `nodeExporter.name` | node-exporter container name | `node-exporter`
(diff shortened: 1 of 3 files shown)


Adam-Stack-PM:
@kahirokunn, Thanks for this PR. We will get this routed to an engineer for review very soon.  

Please note this it may take a few days as our team is traveling to a company event this week. 

CC: @allyleonce

thomasvn:
Thanks for the PR @kahirokunn! I believe the default configuration for node-exporter does not include `hostNetwork: true`. Therefore, node-exporter should be defaulting to the `default` dnsPolicy and not `ClusterFirstWithHostNet`?

kahirokunn:
@thomasvn default configuration for node-exporter `include hostNetwork: true`.
image

thomasvn:
Ah, thanks for pointing me to that @kahirokunn! Yes, so I agree that updating node-exporter to `dnsPolicy: ClusterFirstWithHostNet` would be in alignment with the k8s docs.

Can you think of a way to test this change? Would it be worthwhile running `nslookup` in the node-exporter pod before and after the change?

kahirokunn:
It works well in my environment, so I don't think it is necessary to go that far.

thomasvn:
@kahirokunn I’ve tested in my environment and also see no issue with metrics emitted by node-exporter, and Prometheus scraping node-exporter. Approving as these changes are in accordance to k8s docs.

thomasvn (approved):

PR #2100, opened 2023-04-01
1.102.0 updates

JesseGoodier:
## What does this PR change?
update chart and kubecost.yaml to 1.102.0 
@ausmith this is something that we commonly miss. If you do this, feel free to close this PR.

### DIFF
--- a/kubecost.yaml
+++ b/kubecost.yaml
@@ -20233,11 +20233,6 @@ metadata:
     chart: grafana-1.17.2
     heritage: Helm
     release: kubecost
-rules:
-- apiGroups:      ['extensions']
-  resources:      ['podsecuritypolicies']
-  verbs:          ['use']
-  resourceNames:  [kubecost-grafana]
 ---
 # Source: cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
 apiVersion: rbac.authorization.k8s.io/v1
--- a/kubecost.yaml
+++ b/kubecost.yaml
@@ -60,7 +60,7 @@ metadata:
   labels:
 
     app.kubernetes.io/name: cost-analyzer
-    helm.sh/chart: cost-analyzer-1.101.3
+    helm.sh/chart: cost-analyzer-1.102.0
     app.kubernetes.io/instance: kubecost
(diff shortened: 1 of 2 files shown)


ausmith:
The following issues already track the automation gaps here, so we'll need to do this one manually for 102 as we won't be getting the release scripts fixed up by EOD: 50 and 49 (links removed due to private repo reference)

jessegoodier:
ok, could you approve this so we can merge?

ausmith:
When we have cut release 1.102.0, yes I will approve this PR and merge it.

kubecost.yaml
```diff
@@ -20317,28 +20303,6 @@ subjects:
...
-      app: cost-analyzer
-roleRef:
-    apiGroup: rbac.authorization.k8s.io
-    kind: Role
-    name: kubecost-cost-analyzer-psp
-subjects:
-- kind: ServiceAccount
-  name: kubecost-cost-analyzer
-  namespace: kubecost
----
```
ausmith, line 20341: Does the helm chart also remove these or are we manually removing from the rendered content?

kubecost.yaml
```diff
@@ -20317,28 +20303,6 @@ subjects:
     name: kubecost-cost-analyzer
     namespace: kubecost
 ---
-# Source: cost-analyzer/templates/cost-analyzer-psp-rolebinding.template.yaml
-apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
-metadata:
-    name: kubecost-cost-analyzer-psp
```
ausmith, line 20324: Wouldn't we also need to remove the PSP itself, not just the role and rolebinding that used them?

kubecost.yaml
jessegoodier: this will not remove. users would need to manually remove.

just to be clear- this file is an alternative to helm.

kubecost.yaml
jessegoodier: good find! I'm looking at this now.

jessegoodier:
should be good to go

ausmith (approved):
I'll handle merging once the release is completed as a final step.

PR #2133, opened 2023-04-10
update rules to use current cadvisor labels

JesseGoodier:
## What does this PR change?
remove `_name` from recording rules to support modern cadvisor.

Note that this is compatible with the default install of kubecost with the bundled prometheus as well.

Our check already uses the modern label:
image

- NA
 
update recording rules to use current cadvisor labels. Previously they would need to be modified manually

## How was this PR tested?

test-1 cluster:
default kubecost in namespace `jesse-cadvisor-update`
custom prometheus test in namespace `jesse`

NA

### DIFF
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "1.102.0"
+appVersion: "1.102.1"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor
   cloud costs.
 name: cost-analyzer
-version: "1.102.0"
+version: "1.102.1"
 annotations:
   "artifacthub.io/links": |
     - name: Homepage
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -609,15 +609,15 @@ prometheus:
       groups:
         - name: CPU
           rules:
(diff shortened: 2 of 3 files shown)


jessegoodier:
QA: https://cadvisor.test-1.kubecost.xyz/diagnostics

thomasvn:
@jessegoodier This looks good to me!

What's the level of impact here should these new recording rules fail? In other words, what would be the impact on the product if the `kubecost_container_cpu_usage_irate` or `kubecost_container_memory_working_set_bytes` metrics suddenly become empty?

jessegoodier:
@thomasvn 

per: https://kubecost.slack.com/archives/G01MN4R06MB/p1679959189351259
```
If the rule does not exist, the Allocation pipeline build will fall back to a Prometheus subquery when building CPU max. 
```

We have a ton of non-bundled Prometheus users that don't have any recording rules because it wasn't in the documentation until the past couple of weeks.

thomasvn:
@jessegoodier Thanks! One last request. Could we check that the following metrics have nonzero values in your Prometheus for cadvisor.test-1.kubecost.xyz:

* `kubecost_container_cpu_usage_irate`
* `kubecost_container_memory_working_set_bytes`
* `kubecost_cluster_memory_working_set_bytes`

jessegoodier:
All look healthy:
image
image
image

thomasvn (approved):

PR #2134, opened 2023-04-10
Don't quote sharedNamespaces

michaelmdresser:
Allows it to be overwritten with an empty string

- 
- 

- 
-


michaelmdresser:
Doesn't work, K8s needs values in the `data:` section to be quoted, so this will break for most normal use-cases.

Issue #2067, opened 2023-03-20
Kubecost and its Frontend not working with google managed prometheus (gmp)


funkymcb:
**Describe the bug**  
Hello there. I am trying to apply the kubecost and its frontend to our cluster which runs a google managed prometheus (gmp).
However many metrics are missing and the "Overview" page does not load properly (just a blank white screen).

After running the kubecost/diagnostics, the prometheus health check stated the following error message:
"404: The requested URL `/v1/projects/PROJECT_ID/location/global/prometheus/api/v1/targets` was not found on this server."

Due to this exception the frontend fails to load properly and missing metrics cant't be scraped.
The issue is, that kubecost tries to query the `targets` endpoint.
Unfortunately GMP only provides the following endpoints:
- `/api/v1/query`
- `/api/v1/query_range`
- `/api/v1/metadata`
- `/api/v1/labels`

Is there any possibility to alter the prometheus health check to not query the `api/v1/targets` endpoint?
Or is the only solution to go back to kubecosts own prometheus setup?

**To Reproduce**  
1. Deploy kubecost on a cluster with GMP
2. Disable kubecost prometheus and configure gmp
3. run /kubecost/diagnostics in the frontend

**Expected behavior**  
Prometheus healthcheck succeeds and frontend works properly

**our kubecost helm values.yaml**
```yaml
global:
  thanos:
    enabled: false
  grafana:
    enabled: false
    domainName: grafana.monitoring
    proxy: false # If true, the kubecost frontend will route to your grafana through its service endpoint
  prometheus:
    enabled: false
    fqdn: http://prometheus-server-proxy.monitoring.svc:9090
  additionalLabels:
    unit: peng
prometheus:
  nodeExporter:
    enabled: false
grafana:
  sidecar:
    dashboards:
      enabled: true
    datasources:
      enabled: false
kubecostMetrics:
  emitPodAnnotations: true
  emitNamespaceAnnotations: true
kubecostProductConfigs:
  clusterProfile: "${kubecost_cluster_profile}"
  currencyCode: "EUR"
  grafanaURL: "[...]"
  sharedNamespaces: "kube-system, […]"
  labelMappingConfigs:
    team_label: "unit"
  shareTenancyCosts: true
serviceAccount:
  annotations:
    iam.gke.io/gcp-service-account: "[...]"
nodeSelector:
  unit: peng
```

thomasvn:
[...]

PR #2157, opened 2023-04-17
Update Helm chart to support GCP MP Enterprise integration

linhlam-kc:
## What does this PR change?

Add necessary changes to support Kubecost Enterprise product listed on the Google Cloud Marketplace. So, it can help to accelerate the update process for this integration.

N/A

This is required code changes for our product listed on GCP Marketplace to transact with customers using GCP MP Private Offers

[2067](https://github.com/kubecost/cost-analyzer-helm-chart/issues/2067)

## How was this PR tested?

The integration was tested ad approved on the GCP MP at https://console.cloud.google.com/marketplace/product/kubecost-public/kubecost-ent 

Yes, at the specific repo for GCP Marketplace https://github.com/kubecost/gcp-marketplace

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -22,6 +22,10 @@ global:
     proxy: true # If true, the kubecost frontend will route to your grafana through its service endpoint
 #    fqdn: cost-analyzer-grafana.default.svc
 
+  # Enable only when you are using GCP Marketplace ENT listing. Learn more at https://console.cloud.google.com/marketplace/product/kubecost-public/kubecost-ent    
+  gcpstore:
+    enabled: false
+
   # Amazon Managed Service for Prometheus
   amp:
     enabled: false # If true, kubecost will be configured to remote_write and query from Amazon Managed Service for Prometheus.
@@ -893,6 +897,11 @@ kubecostAdmissionController:
 # It is disabled by default to avoid problems in high-scale environments.
 costEventsAudit:
   enabled: false
+
+# Reporting Secret for GCP Marketplace integration.
+# The value will be replaced automatically when Kubecost is deployed following GCP MP instructions.
+# Do not change the value manually.
+reportingSecret: "kubecost-reporting-secret"
 
 # readonly: false # disable updates to kubecost from the frontend UI and via POST request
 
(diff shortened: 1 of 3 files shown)


cost-analyzer/values.yaml
```diff
@@ -894,6 +898,11 @@ kubecostAdmissionController:
 costEventsAudit:
   enabled: false
 
+# Reporting Secret for GCP Marketplace integration.
+# The value will be replaced automatically when Kubecost is deployed following GCP MP instructions.
+# Do not change the value manually.
+reportingSecret: "kubecost-reporting-secret"
```
thomasvn: @linhlam-kc If this value shouldn't be changed manually, should we set this as a the default in `cost-analyzer/templates/gcpstore-config-map-template.yaml` and remove it from this `values.yaml`?

cost-analyzer/values.yaml
```diff
@@ -22,6 +22,10 @@ global:
     proxy: true # If true, the kubecost frontend will route to your grafana through its service endpoint
 #    fqdn: cost-analyzer-grafana.default.svc
 
+  # Enable only when you are using GCP Marketplace ENT listing. Learn more at https://console.cloud.google.com/marketplace/product/kubecost-public/kubecost-ent
```
thomasvn, line 25: @linhlam-kc Could you provide more comment on what the purpose of the `ubbagent` is and why it is necessary when enabling Kubecost via the GCP Marketplace?

cost-analyzer/values.yaml
linhlam-kc: It is required to be parameterized and to be in the Helm chart values file, so GCP MP deployment tool can substitute the value at the deployment time. GCP MP does not use Helm client to deploy Helm Chart in normal way.

thomasvn: Would it work if we removed these lines from `values.yaml`, and instead did something like this in the template:

```yaml
          - name: AGENT_ENCODED_KEY
            valueFrom:
              secretKeyRef:
                name: {{ default "kubecost-reporting-secret" .Values.reportingSecret }}
                key: reporting-key
```

linhlam-kc: Thank you Thomas ! It should work, I have tested locally and apply the changes in the latest commit [8285d6c](https://github.com/kubecost/cost-analyzer-helm-chart/pull/2157/commits/8285d6cbc83225ba14eb28271601519cb4b44170)

thomasvn (approved):

PR #2173, opened 2023-04-22
EV-59

nik-kc:
## What does this PR change?
Adds support for Query Service Replicas.

- https://github.com/kubecost/kubecost-cost-model/pull/1392

If Query Service Replicas are disabled, functionality is the same. Otherwise, /model/allocation and /model/assets endpoints will be forward to a StatefulSet of Query Services, round robined by a load balancer.

- https://kubecost.atlassian.net/browse/EV-59

## How was this PR tested?
Deployed to test-1 in GKE, with QS replicas enabled and disabled. When enabled, verified that traffic is routed to load balancer, and that load balancer is properly balancing requests between replicas.

No

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -5,6 +5,9 @@ Expand the name of the chart.
 {{- define "cost-analyzer.name" -}}
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+{{- define "query-service.name" -}}
+{{- default "query-service" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
 
 {{/*
 Create a default fully qualified app name.
@@ -22,6 +25,10 @@ If release name contains chart name it will be used as a full name.
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 {{- end -}}
+{{- end -}}
+
+{{- define "query-service.fullname" -}}
+{{- printf "%s-%s" .Release.Name "query-service" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/*
@@ -96,6 +103,12 @@ helm.sh/chart: {{ include "cost-analyzer.chart" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
+{{- define "kubecost.queryService.chartLabels" -}}
+app.kubernetes.io/name: {{ include "query-service.name" . }}
+helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
 
 
 {{/*
@@ -115,6 +128,13 @@ Create the name of the service account
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+{{- define "query-service.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "query-service.fullname" .) .Values.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
 
 {{/*
 Create the common labels.
@@ -126,6 +146,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app: cost-analyzer
 {{- end -}}
+{{- define "query-service.commonLabels" -}}
+{{ include "kubecost.queryService.chartLabels" . }}
+app: query-service
+{{- end -}}
 
 {{/*
 Create the selector labels.
@@ -134,6 +158,11 @@ Create the selector labels.
 app.kubernetes.io/name: {{ include "cost-analyzer.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app: cost-analyzer
+{{- end -}}
+{{- define "query-service.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "query-service.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app: query-service
 {{- end -}}
 
 {{/*
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -101,6 +101,12 @@ data:
     }
 {{- end }}
 
+  {{- if and (.Values.kubecostDeployment) (.Values.kubecostDeployment.queryServiceReplicas) (gt (.Values.kubecostDeployment.queryServiceReplicas | toString | atoi) 0) }}
+    upstream queryservice {
+        server {{ .Release.Name }}-query-service-load-balancer.{{ .Release.Namespace }}:9003;
+    }
+  {{- end }}
+
     server {
         server_name _;
         root /var/www;
@@ -283,6 +289,32 @@ data:
         {{- if .Values.saml.rbac.enabled }}
         location /authrbac {
             proxy_pass http://model/isAdminAuthenticated;
+        }
+    {{- end }}
+
+    {{- if and (.Values.kubecostDeployment) (.Values.kubecostDeployment.queryServiceReplicas) (gt (.Values.kubecostDeployment.queryServiceReplicas | toString | atoi) 0) }}
+        location /model/allocation {
+            proxy_connect_timeout       180;
+            proxy_send_timeout          180;
+            proxy_read_timeout          180;
+            proxy_pass http://queryservice/allocation;
+            proxy_redirect off;
+            proxy_http_version 1.1;
+            proxy_set_header Connection "";
+            proxy_set_header  X-Real-IP  $remote_addr;
+            proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
+        }
+
+        location /model/assets {
+            proxy_connect_timeout       180;
+            proxy_send_timeout          180;
+            proxy_read_timeout          180;
+            proxy_pass http://queryservice/assets;
+            proxy_redirect off;
+            proxy_http_version 1.1;
+            proxy_set_header Connection "";
+            proxy_set_header  X-Real-IP  $remote_addr;
+            proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
     {{- end }}
     }
--- a/cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
@@ -1,0 +1,34 @@
+{{- if and (not .Values.agent) (not .Values.cloudAgent) (.Values.kubecostDeployment) (.Values.kubecostDeployment.queryServiceReplicas) }}
+{{- if gt .Values.kubecostDeployment.queryServiceReplicas 0 }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ template "query-service.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{ include "query-service.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "query-service.serviceAccountName" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "query-service.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "query-service.serviceAccountName" . }}
+  labels:
+    {{ include "query-service.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "query-service.serviceAccountName" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "query-service.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }} 
+{{- end }}
+{{- end }}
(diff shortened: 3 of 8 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -285,5 +291,31 @@ data:
             proxy_pass http://model/isAdminAuthenticated;
         }
     {{- end }}
+
+    {{- if and (.Values.kubecostDeployment.queryServiceReplicas) (.Values.kubecostDeployment.queryServiceReplicas.enabled) }}
+        location /model/allocation {
```
michaelmdresser, line 296: We won't be using this Helm config with the frontend yet, right? I'm checking because I think this will have minimal impact on FE until we implement things like Allocation Summary, Allocation View, etc.

cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
```diff
@@ -0,0 +1,34 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "query-service.serviceAccountName" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "query-service.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }} 
+{{- end }}
+{{- end }}
```
michaelmdresser, line 34: Why do we have a `ClusterRole`(`Binding`) for query service? What access does it need to resources outside of the release namespace?

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -0,0 +1,113 @@
+{{- if and (not .Values.agent) (not .Values.cloudAgent) (.Values.kubecostDeployment) }}
+{{- if and (.Values.kubecostDeployment.queryServiceReplicas) (.Values.kubecostDeployment.queryServiceReplicas.enabled) }}
+apiVersion: apps/v1
+kind: StatefulSet
```
michaelmdresser: Why is query service a StatefulSet? I apologize for my lack of context. If it is actually supposed to be/needs to be a StatefulSet, we should probably change the file name from `deployment-template` to `statefulset-template`.

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -0,0 +1,113 @@
...
+            failureThreshold: 200
+          imagePullPolicy: Always
+          args: ["query-service"]
+          ports:
+            - name: tcp-model
+              containerPort: 9003
+              protocol: TCP
+            - name: tcp-frontend
+              containerPort: 9090
+              protocol: TCP
```
michaelmdresser: Does query service actually have a frontend (or anything) running on 9090?

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -0,0 +1,113 @@
...
+          volumeMounts:
+            - name: persistent-configs
+              mountPath: /var/configs
+            - name: etl-bucket-config
+              mountPath: /var/configs/etl
+              readOnly: true
+            - name: persistent-db
+              mountPath: /var/db
```
michaelmdresser: Naively, I would assume query service would only need read access to all persistent data, not just ETL. What am I missing?

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -0,0 +1,113 @@
...
+            - name: ETL_FILE_STORE_ENABLED
+              value: "true"
+            - name: ETL_TO_DISK_ENABLED
+              value: "false"
+            {{- if $etlBackupBucketSecret }}
+            - name: ETL_BUCKET_CONFIG
+              value: "/var/configs/etl/object-store.yaml"
+            {{- end }}
+            - name: ETL_TO_DISK_ENABLED
+              value: "true"
```
michaelmdresser: Looks like we have `ETL_TO_DISK_ENABLED` twice with conflicting definitions. Which should it be?

kubecost.yaml
```diff
@@ -506,13 +506,15 @@ data:
     upstream api {
         server kubecost-cost-analyzer.kubecost:9001;
     }
-
```
michaelmdresser: Please don't modify `kubecost.yaml` in a feature PR. It has to be updated with specific `helm template` flags and that generally occurs alongside a release of Kubecost.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
nik-kc: I have all of the `/model/allocation` and (almost) all of the `/model/assets` endpoints working in the Query Service replicas. The replicas are wholly for the use of the frontend in its current iteration, so we'll need this with the frontend.

cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
nik-kc: Well that would be because I still only have a beginner's understanding of roles and such, and misunderstood what this does. I'll take it out, as we don't need to access anything outside of the release namespace.

cost-analyzer/templates/query-service-deployment-template.yaml
nik-kc: This is a conversation I had with Ajay, we're using a StatefulSet so that each replica can have its own PV. And yes, the template name should definitely say statefulset, not deployment.

cost-analyzer/templates/query-service-deployment-template.yaml
nik-kc: Negative, I'll take that out.

cost-analyzer/templates/query-service-deployment-template.yaml
nik-kc: ~Based on my in-progress understanding of how KC uses various data, we should only need ETL. All the QS replicas do is maintain a copy of ETL data from a backup store (populated by the cost model), and use the existing QS implementation in `pkg/query/service.go` to query the data for the frontend.~

Oh, I misread your comment. They should all be read only, yes.

cost-analyzer/templates/query-service-deployment-template.yaml
nik-kc: I'm glad you're looking at this Michael, I don't know how I missed this. It should be true.

kubecost.yaml
nik-kc: Makes total sense! I didn't intend to edit that file at all, must've been an accidental commit. I'll revert that.

cost-analyzer/templates/query-service-deployment-template.yaml
michaelmdresser: I saw you changed the PV mounts on the StatefulSet to both be read only (because I asked about it). If all PV mounts are read only, do we need a StatefulSet? Perhaps this is the question point; Ajay may have insight or knowledge here about PVs/configs of cost-model that I'm forgetting.

nik-kc: Just ran some tests, I was wrong to set them to readOnly to begin with. I've updated that configuration, none of them are readOnly now.

cost-analyzer/templates/_helpers.tpl
```diff
@@ -126,6 +143,13 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app: cost-analyzer
 {{- end -}}
+{{- define "query-service.commonLabels" -}}
+app.kubernetes.io/name: query-service
+helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
ameijer: include the query-service.chartLables here to avoid duplication

cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
ameijer: @nik-kc looks like we still have a clusterRoleBinding here?

cost-analyzer/templates/query-service-cluster-role-template.yaml
```diff
@@ -0,0 +1,109 @@
...
+      - list
+      - watch
+  - apiGroups: 
+      - storage.k8s.io
+    resources: 
+      - storageclasses
+    verbs:
+      - get
+      - list
+      - watch
```
ameijer, line 99: does this query service need all these perms?

cost-analyzer/templates/query-service-service-template.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    helm.sh/chart: {{ include "cost-analyzer.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app: query-service
+spec:
+  selector:
+    app.kubernetes.io/name: query-service
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app: query-service
```
ameijer: define a selectorLabels template here

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -0,0 +1,95 @@
...
+{{- if and (.Values.kubecostDeployment.queryServiceReplicas) (.Values.kubecostDeployment.queryServiceReplicas.enabled) }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "query-service.fullname" . }}
+  namespace: {{ .Release.Name }}
+  labels:
+    {{ include "query-service.commonLabels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.kubecostDeployment.queryServiceReplicas.replicas | default 1 }}
```
ameijer: you don't need a default here. The values.yaml you set serves as the default

cost-analyzer/templates/_helpers.tpl
nik-kc: Fixed.

cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
nik-kc: Fixed, thanks for noticing that.

cost-analyzer/templates/query-service-cluster-role-template.yaml
ameijer: so, apparenty it does. I see you instantiate a clustercache which does the following operation: 

```
	client:                     client,
		namespaceWatch:             NewCachingWatcher(coreRestClient, "namespaces", &v1.Namespace{}, "", fields.Everything()),
		nodeWatch:                  NewCachingWatcher(coreRestClient, "nodes", &v1.Node{}, "", fields.Everything()),
		podWatch:                   NewCachingWatcher(coreRestClient, "pods", &v1.Pod{}, "", fields.Everything()),
		kubecostConfigMapWatch:     NewCachingWatcher(coreRestClient, "configmaps", &v1.ConfigMap{}, kubecostNamespace, fields.Everything()),
		serviceWatch:               NewCachingWatcher(coreRestClient, "services", &v1.Service{}, "", fields.Everything()),
		daemonsetsWatch:            NewCachingWatcher(appsRestClient, "daemonsets", &appsv1.DaemonSet{}, "", fields.Everything()),
		deploymentsWatch:           NewCachingWatcher(appsRestClient, "deployments", &appsv1.Deployment{}, "", fields.Everything()),
		statefulsetWatch:           NewCachingWatcher(appsRestClient, "statefulsets", &appsv1.StatefulSet{}, "", fields.Everything()),
		replicasetWatch:            NewCachingWatcher(appsRestClient, "replicasets", &appsv1.ReplicaSet{}, "", fields.Everything()),
		pvWatch:                    NewCachingWatcher(coreRestClient, "persistentvolumes", &v1.PersistentVolume{}, "", fields.Everything()),
		pvcWatch:                   NewCachingWatcher(coreRestClient, "persistentvolumeclaims", &v1.PersistentVolumeClaim{}, "", fields.Everything()),
		storageClassWatch:          NewCachingWatcher(storageRestClient, "storageclasses", &stv1.StorageClass{}, "", fields.Everything()),
		jobsWatch:                  NewCachingWatcher(batchClient, "jobs", &batchv1.Job{}, "", fields.Everything()),
		pdbWatch:                   NewCachingWatcher(pdbClient, "poddisruptionbudgets", &v1beta1.PodDisruptionBudget{}, "", fields.Everything()),
		replicationControllerWatch: NewCachingWatcher(coreRestClient, "replicationcontrollers", &v1.ReplicationController{}, "", fields.Everything()),

```
		
So, we might actually need a cluster role to get these resources across all namespaces @michaelmdresser ?

cost-analyzer/templates/query-service-service-template.yaml
nik-kc: Done.

cost-analyzer/templates/query-service-deployment-template.yaml
nik-kc: I'm following [this example](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L17), which has a default even though it's defined in values.yaml.

cost-analyzer/templates/query-service-cluster-role-binding-template.yaml
nik-kc: Seems to actually be necessary. I tested without the binding, and I see error logs such as:

`failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:ev-59:ev-59-query-service" cannot list resource "configmaps" in API group "" in the namespace "kubecost"`

cost-analyzer/templates/query-service-cluster-role-template.yaml
michaelmdresser: Aha, yes that could be! Good catch.

I'm a little concerned; if query service needs clustercache (which makes sense) the load on the K8s API might increase noticeably. This could introduce stability problems for large users. I hope we're aware of this when testing!

nik-kc: What's interesting is the the _only_ thing that the clustercache is used for is to retrieve the currency code for the provider. In costmodel, some of the /allocation and /assets handlers take in a cloudprovider (created in part using the clustercache) in order to determine the currency code for PDF exports, and those handlers are reused in queryservice. The only attribute of the cloudprovider that the handlers use is the config, which holds the currency code.

I'm looking into better ways to do this, as we clearly don't need to be watching the cluster just to get the currency code.

nik-kc: For reference, check out the usage of `cp` in this handler:
https://github.com/kubecost/kubecost-cost-model/blob/develop/pkg/query/handlers.go#L335

cost-analyzer/templates/query-service-deployment-template.yaml
nik-kc: I've made some updates to the config, this default has been removed.

cost-analyzer/templates/query-service-cluster-role-template.yaml
nik-kc: @ameijer I talked with Michael, and this is indeed an issue, but it's something that I'll tackle if I have time after addressing all the other feedback on this project.

ameijer: 👍

ameijer (approved):
LGTM

PR #2188, opened 2023-04-27
feat(cost-analyzer): add StatefulSet as option

sebbrandt87:
## What does this PR change?
It enables the option to create a `StatefulSet` with `claimVolumeTemplates` and any `StorageClass`. 

With the current `Deployment` implementation you only have two options, if you want to set up cost-analyzer in HA mode (replicas > 1).

1. You can use `emptyDir`, as this would create a volume mount without a PV on each Kubernetes Node where a Pod is started.
2. You can use a `StorageClass` which enables you to get a PV with [RWM (ReadWriteMany)].

With this feature you are now able to use any `StorageClass` as the `claimVolumeTemplates` creates automatically a PVC for each Pod scaled by the `StatefulSet`.

No

With `.statefulSet.enabled = true` you are able to get a cost-analyzer `StatefulSet` rolled out, instead of a `Deployment` and get automatically a `PersistenVolumeClaim` per scaled `Pod`. As this is a StatefulSet feature.

None created

## How was this PR tested?
Kind cluster and a started StatefulSet as written in the `README.md` with ct (chart-testing) cli command.

Not yet

[RWM (ReadWriteMany)]: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

### DIFF
--- a/cost-analyzer/README.md
+++ b/cost-analyzer/README.md
@@ -99,3 +99,57 @@ Adjusting the log format changes the format in which the logs are output making 
 |--------|----------------------------------------------------------------------------------------------------------------------------|
 | `JSON`   | `{"level":"info","time":"2006-01-02T15:04:05.999999999Z07:00","message":"Starting cost-model (git commit \"1.91.0-rc.0\")"}` |
 | `pretty` | `2006-01-02T15:04:05.999999999Z07:00 INF Starting cost-model (git commit "1.91.0-rc.0")`                                     |
+
+## Testing
+To perform local testing do next:
+- install locally [kind](https://github.com/kubernetes-sigs/kind) according to documentation.
+- install locally [ct](https://github.com/helm/chart-testing) according to documentation.
+- create local cluster using `kind` \
+use image version from https://github.com/kubernetes-sigs/kind/releases e.g. `kindest/node:v1.25.11@sha256:227fa11ce74ea76a0474eeefb84cb75d8dad1b08638371ecf0e86259b35be0c8`
+```shell
+kind create cluster --image kindest/node:v1.25.11@sha256:227fa11ce74ea76a0474eeefb84cb75d8dad1b08638371ecf0e86259b35be0c8
+```
+- perform ct execution
+```shell
+ct install  --chart-dirs="." --charts="."
+```
+
+- perform ct StatefulSet execution
+
+```shell
+# create multiple nodes kind config
+cat > kind-config.yaml <<EOF
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+EOF
+# creaet kind cluster with kind config
+kind create cluster --name kubecost-statefulset --config kind-config.yaml --image kindest/node:v1.25.11@sha256:227fa11ce74ea76a0474eeefb84cb75d8dad1b08638371ecf0e86259b35be0c8
+# deploy an object storage for our testing purpose (https://min.io/docs/minio/kubernetes/upstream/index.html)
+curl --silent https://raw.githubusercontent.com/minio/docs/master/source/extra/examples/minio-dev.yaml  | sed -e "s/kubealpha.local/kubecost-statefulset-worker/" -e "s%minio server /data%mkdir -p /data/kubecost; minio server /data%" | kubectl apply -f -
+# create a headless service to the minio S3 API port
+kubectl create service clusterip -n minio-dev minio --tcp=9000:9000 --clusterip="None"
+# create our testing namespace
+kubectl create namespace kubecost-statefulset
+# create the bucket config 
+cat > etlBucketConfigSecret.yaml <<EOF
+type: s3
+config:
+  bucket: kubecost
+  endpoint: minio.minio-dev:9000
+  insecure: true
+  access_key: minioadmin
+  secret_key: minioadmin
+EOF
+# create the secret with the object-store.yaml
+kubectl create secret generic -n kubecost-statefulset object-store --from-file=object-store.yaml=etlBucketConfigSecret.yaml
+# start our chart-testing
+ct install --namespace kubecost-statefulset --chart-dirs="." --charts="." --helm-extra-set-args="--set=global.prometheus.enabled=true --set=global.grafana.enabled=true --set=kubecostDeployment.leaderFollower.enabled=true --set=kubecostDeployment.statefulSet.enabled=true --set=kubecostDeployment.replicas=2 --set=kubecostModel.etlBucketConfigSecret=object-store"
+# cleanup
+kind delete cluster --name kubecost-statefulset 
+```
+
+
--- a/cost-analyzer/templates/cost-analyzer-db-pvc-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-db-pvc-template.yaml
@@ -3,6 +3,7 @@
 {{- if not .Values.persistentVolume.dbExistingClaim -}}
 {{- if .Values.persistentVolume.enabled -}}
 {{- if .Values.persistentVolume.dbPVEnabled -}}
+{{- if not (and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled) -}}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
@@ -30,6 +31,7 @@ spec:
     {{- else }}
       storage: 32.0Gi
     {{ end }}
+{{- end -}}
 {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -1,7 +1,11 @@
 {{- if and (not .Values.agent) (not .Values.cloudAgent) }}
 {{- $nginxPort := int .Values.service.port | default 9090 -}}
 apiVersion: apps/v1
+{{- if and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled }}
+kind: StatefulSet
+{{- else }}
 kind: Deployment
+{{- end }}
 metadata:
   name: {{ template "cost-analyzer.fullname" . }}
   namespace: {{ .Release.Namespace }}
@@ -17,6 +21,9 @@ metadata:
 spec:
 {{- if .Values.kubecostDeployment }}
   replicas: {{ .Values.kubecostDeployment.replicas | default 1 }}
+{{- end }}
+{{- if and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled }}
+  serviceName: {{ template "cost-analyzer.serviceName" . }}
 {{- end }}
   selector:
     matchLabels:
@@ -264,6 +271,7 @@ spec:
         # Extra volume(s)
         {{- toYaml .Values.extraVolumes | nindent 8 }}
         {{- end }}
+{{- if not (and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled) }}
         - name: persistent-configs
 {{- if .Values.persistentVolume }}
 {{- if .Values.persistentVolume.enabled }}
@@ -280,7 +288,8 @@ spec:
           persistentVolumeClaim:
             claimName: {{ template "cost-analyzer.fullname" . }}
 {{- end }}
-{{- if and (.Values.kubecostModel.etlToDisk | default true) .Values.persistentVolume.dbPVEnabled }}
+{{- end }}
+{{- if and (.Values.kubecostModel.etlToDisk | default true) .Values.persistentVolume.dbPVEnabled (not (and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled)) }}
         - name: persistent-db
 {{- if .Values.persistentVolume }}
 {{- if .Values.persistentVolume.enabled }}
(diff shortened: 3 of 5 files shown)


thomasvn:
Thanks for the PR @sebbrandt87! Could you share more about some of the advantages/disadvantages of running the cost-analyzer as a `StatefulSet`?

Would appreciate any opinions from @ausmith as well!

sebbrandt87:
@thomasvn @ausmith I updated the PR description accordingly. I hope this is enough. 😉

thomasvn:
@sebbrandt87 Thanks for the update! Also for your awareness, Kubecost does have ability to run in HA mode without using a `StatefulSet` https://docs.kubecost.com/install-and-configure/advanced-configuration/high-availability.

@AjayTripathy @mbolt35 What're your thoughts on the viability of `StatefulSet`? Would this effectively create multiple primaries, each with their own persistent storage? If so, I can see the current kubecost-cost-analyzer pod duplicating effort in cost calculations.

sebbrandt87:
> Would this effectively create multiple primaries, each with their own persistent storage? 

Yes it would. And as the etlbackup is an requirement for being activated with `replicas > 1` and is filled then, the followers would restore the data from it. 

> If so, I can see the current kubecost-cost-analyzer pod duplicating effort in cost calculations.

From my understanding only the leader does the calculation, which absolutely makes sense.

michaelmdresser:
> From my understanding only the leader does the calculation, which absolutely makes sense.

@sebbrandt87 I'm not sure that this is true. Do you have a reference for this information?

sebbrandt87:
@michaelmdresser sorry, no. It was just a guess from a quick overview. Actually I think all replicas, anyhow if Deployment or StatefulSet (as this changes nothing logic wise for cost-analyzer itself), do the calculation as long as it is not present in the etl data. But only the elected leader, will write into the etlbackup and the followers will sync calculated data from there. https://docs.kubecost.com/install-and-configure/advanced-configuration/high-availability.

Anyhow regarding 
> If so, I can see the current kubecost-cost-analyzer pod duplicating effort in cost calculations.

Nothing changes to current Deployment `replicas > 1` with a StatefulSet instead. It would do the same as yet but lacks the PR related enhancement of being able to really use multiple PVCs, controlled by the `StatefulSet`.

thomasvn:
@sebbrandt87 Have you tried implementing [High Availability Kubecost](https://docs.kubecost.com/install-and-configure/advanced-configuration/high-availability)? Would that fulfill your use case?

sebbrandt87:
@thomasvn this would not fulfill the use case. We are not going to set `emptyDir` as a volume and then always sync again from storage / bucket the ETL backup files to the new node (if we do not get our Pod running on the same node as before).
The only way to get HA running with PVCs not being a RWM but a RWO volume is with a `StatefulSet` and created PVCs by that `StatefulSet` through the `claimVolumeTemplates`. Anything with `data` that is kind of a persistent thing, should be a `StatefulSet`.

mbolt35:
Apologies for taking a while to respond here. StatefulSets, based on my knowledge, simply provide an ordered naming policy for replicas, and is really meant to assist with `claimVolumeTemplates` to ensure the right volume claims are applied across restarts? I can say with confidence that attempting to modify the storage volumes here to support statefulsets is going to create problems for kubecost. We would've absolutely gone the Statefulset route had it supported our product features fully (or even partially), but there are many issues with the disk bound ETL storage replication that will create lots of churn on prometheus, replication and rebuilding of all internal data stores, and overall, instability and desynchronization in the product. This is the primary reason we built the leader/follower implementation to synchronize across a shared backup storage, and actually implement legitimate read/write policies for each role. We are working tirelessly to improve the way a lot of our internal systems work today such that we can hopefully reach a point where leveraging a shared volume is possible, but we aren't quite there yet. 

Hopefully this is helpful, or at least sheds some light on why these specific decisions were made.

timroh:
Hi @mbolt35,
I have trouble following you here.

What is then the correct way to deploy cost-analyzer with 3 or more replicas using the PV that is described in the [documentation](https://docs.kubecost.com/install-and-configure/install/storage)?

thomasvn:
@TimRoh I believe the correct way would be to deploy following the [Kubecost High Availability Documentation](https://docs.kubecost.com/install-and-configure/advanced-configuration/high-availability). In this case each replica has its own PV.

thomasvn:
@sebbrandt87 @TimRoh Thanks for working with @ivankube and I in helping us understand your use case for StatefulSet. After discussing further with engineering, we believe using a `StatefulSet` would be good _only_ when used in addition to Kubecost's HA LeaderFollower `.Values.kubecostDeployment.leaderFollower.enabled=true`.

What are your thoughts on modifying your changes to the template such that a `StatefulSet` would be created only when the following conditions are true?

```mustache
{{- if and .Values.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled }}
```

sebbrandt87:
@thomasvn you can take a look, how I implemented it into `_helpers.tpl` and added instructions for testing purpose.

thomasvn:
Thanks so much for the continued work on this @sebbrandt87. @ivankube has tested some of these changes, and has an working environment with StatefulSet + LeaderFollower.

Upon some further testing though, it seems that it's still possible to create a `StatefulSet` even when `kubecostDeployment.leaderFollower.enabled!=true`?

```sh
$ helm template ./cost-analyzer --set statefulSet.enabled=true --debug | grep -C3 StatefulSet

---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: release-name-cost-analyzer
  namespace: kubecost

```

Although not very elegant we may need to change the following

```mustache
# From this
{{- if "cost-analyzer.statefulSet" }}

# To this
{{- if and .Values.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled }}
```

sebbrandt87:
@thomasvn so after a longer delay, I finally did 😉 
let me know if it is okay now.

sebbrandt87:
or @ivankube 😉 ☝️

thomasvn:
@sebbrandt87 Again, thanks for this work!! I've tested the following and this looks good to me!

```sh
# This correctly returns nothing.
# We only allow Kubecost to be deployed as a StatefulSet when `kubecostDeployment.leaderFollower.enabled=true`
$ helm template ./cost-analyzer --set statefulSet.enabled=true --debug | grep -C3 StatefulSet
```

```sh
# This deploys Kubecost as a StatefulSet!
$ helm template ./cost-analyzer --set statefulSet.enabled=true \
    --set kubecostDeployment.leaderFollower.enabled=true \
    --set kubecostModel.etlBucketConfigSecret=kubecost-bucket-secret \
    --set kubecostDeployment.replicas=2 | grep -C3 StatefulSet

---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: release-name-cost-analyzer
  namespace: kubecost
```

thomasvn (approved):
@jessegoodier For final review & merge of this PR? Let's target this change for v1.107?

jessegoodier:
I'll give this a test next week, sorry for my delay.

As Thomas said, we will get this in 1.107.

cost-analyzer/values.yaml
```diff
@@ -1010,6 +1010,9 @@ kubecostAdmissionController:
 costEventsAudit:
   enabled: false
 
+# Instead of a kubecost-analyzer Deployment, you can set it to be a StatefulSet as for volumeClaimTemplates usage and real stateful behaviour
+statefulSet:
+  enabled: false
```
chipzoller: Because this setting only impacts the cost-analyzer Deployment, I would prefer to see this under the top-level object `kubecostDeployment`.

cost-analyzer/README.md
chipzoller, line 1: I don't think any of these changes are necessary so I'd remove them. It'll help resolve conflicts as well.

chipzoller (dismissed):
In general this looks fine other than resolving conflicts and some reorg.

cost-analyzer/values.yaml
sebbrandt87: but it isn't a `Deployment` Kubernetes wise - if it should go under `kubecostDeployment` I would say lets rephrase the variable `kubecostDeployment` overall 😉

cost-analyzer/README.md
sebbrandt87: I solved the conflicts. 😄

cost-analyzer/values.yaml
chipzoller: I agree, but doing so now would be a breaking change which I think we want to avoid until we have a more substantial release where we can batch other breaking changes.

sebbrandt87: done ✔️

cost-analyzer/templates/cost-analyzer-db-pvc-template.yaml
```diff
@@ -3,6 +3,7 @@
 {{- if not .Values.persistentVolume.dbExistingClaim -}}
 {{- if .Values.persistentVolume.enabled -}}
 {{- if .Values.persistentVolume.dbPVEnabled -}}
+{{- if and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled -}}
```
AjayTripathy: We shouldn't necessarily block users from installing this to just using a statefulset?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -17,6 +21,9 @@ metadata:
 spec:
 {{- if .Values.kubecostDeployment }}
   replicas: {{ .Values.kubecostDeployment.replicas | default 1 }}
+{{- end }}
+{{- if and .Values.kubecostDeployment.statefulset.enabled .Values.kubecostDeployment.leaderFollower.enabled }}
+  serviceName: {{ template "cost-analyzer.serviceName" . }}
```
AjayTripathy, line 26: Can you help me understand why a servicename field gets added if and only if we use a statefulset?

cost-analyzer/templates/cost-analyzer-pvc-template.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.persistentVolume -}}
 {{- if not .Values.persistentVolume.existingClaim -}}
 {{- if .Values.persistentVolume.enabled -}}
+{{- if and .Values.kubecostDeployment.statefulset.enabled .Values.kubecostDeployment.leaderFollower.enabled -}}
```
AjayTripathy: Don't think we want to force users to run a statefulset to run a PV

AjayTripathy (dismissed):
I think it's ok to make the changes mostly additive, but I don't think we want all users to be forced to use a PV with a statefulset?

cost-analyzer/templates/cost-analyzer-db-pvc-template.yaml
sebbrandt87: good catch @AjayTripathy, before the changes proposed to be done for a better combination with `.Values.kubecostDeployment.leaderFollower.enabled` it was a negated `.Values.kubecostDeployment.statefulSet.enabled` so I added the negation as in the `Deployment` already present as well

cost-analyzer/templates/cost-analyzer-pvc-template.yaml
sebbrandt87: good catch @AjayTripathy, before the changes proposed to be done for a better combination with `.Values.kubecostDeployment.leaderFollower.enabled` it was a negated `.Values.kubecostDeployment.statefulSet.enabled` so I added the negation as in the `Deployment` already present as well

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
sebbrandt87: A StatefulSet needs a `serviceName` by definition https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

sebbrandt87:
@thomasvn @AjayTripathy @chipzoller correct / good to go in now? 🙂

chipzoller:
See CI failures, please.

sebbrandt87:
@chipzoller my mistake, was case-sensitivity issue 😿 - fixed!

thomasvn (approved):
These changes look good to me. Thanks @sebbrandt87! Once merged, we'll continuously test against this in our Nightly environments.

thomasvn:
Because I believe all of @AjayTripathy 's and @chipzoller 's requested changes have been made, I'm going to merge into develop and monitor this in our nightly builds.

thomasvn:
@cliffcolvin @jessegoodier v1.107 release candidate was cut before this PR was merged. Should we cherry-pick this into v1.107 branch?

chipzoller:
Is this being requested for 107?

FernandoMiguel:
I did a quick test with stable 1.107 , without enabling `leaderFollower.enabled` and it failed spectacularly .
I'll see if I have some time this week to test this better.
thanks for the change @sebbrandt87

sebbrandt87:
> I did a quick test with stable 1.107 , without enabling `leaderFollower.enabled` and it failed spectacularly .
> I'll see if I have some time this week to test this better.
> thanks for the change @sebbrandt87 

`leaderFollower.enabled` must be `true` so that you will get a StatefulSet. Additional the setting to enable the StatefulSet as well. See within the values.yaml

FernandoMiguel:
@sebbrandt87 i saw it later.
shouldnt this be caught by an helper or test condition if the statefulset is enabled?
why require the practitioner to change something in two places, if they have an hard dependency?

sebbrandt87:
@FernandoMiguel It could be and makes totally sense. I think we can push this enhancement into next. 😉

sebbrandt87:
@FernandoMiguel it seems the commit wasn't cherry picked into the `v1.107.0` release, I already informed kubecost team about it. Hoping for a quick fix with a `v1.107.1` release.

dwbrown2:
I believe this was picked up in v1.107.1 release, but let me know if I'm mistaken.

FernandoMiguel:
@dwbrown2 i don't see it mentioned or linked in the changelog.
I didn't spend time diff'ing tho.

dwbrown2:
See it here: https://github.com/kubecost/cost-analyzer-helm-chart/releases/tag/v1.107.1

@cliffcolvin to confirm

image

FernandoMiguel:
Thanks.
I'm sure I missed it.
I'll give it a try next week.

Tho, like mentioned before, having to enable two options that should be a conditional toggle, would make this feature much easier to use.

PR #2193, opened 2023-04-27
Always set GOOGLE_APPLICATION_CREDENTIALS in CC

michaelmdresser:
## What does this PR change?
These if statements were not consistent with our public install instructions for Cluster Controller (they do not specify that gcpSecretName should be set) and also are not consistent with this template for the deployment because even if gcpSecretName is set, it is not mounted anywhere. 

By making this change, following our full Cluster Controller install instructions on GKE will work correctly.

Fixed a bug in the Helm chart when using Cluster Controller in GKE with full capabilities (turndown, etc.). The GCP service account credentials were not being correctly passed to the application, resulting in a failure to initialize.

- Fixes https://kubecost.atlassian.net/browse/CORE-209

## How was this PR tested?
Followed the public Kubecost documentation for installing Cluster Controller with turndown enabled on a fresh GKE cluster. Observed correct startup logs, instead of the failure "validating provider" logs observed previously.

N/A, this PR makes functionality consistent with our documentation.

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -212,12 +212,8 @@ spec:
           value: {{ .Release.Namespace }}
         - name: TURNDOWN_DEPLOYMENT
           value: {{ template "kubecost.clusterControllerName" . }}
-        {{- if .Values.kubecostProductConfigs }}
-        {{- if .Values.kubecostProductConfigs.gcpSecretName }}
         - name: GOOGLE_APPLICATION_CREDENTIALS
           value: /var/keys/service-key.json
-        {{- end }}
-        {{- end }}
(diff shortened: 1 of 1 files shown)


michaelmdresser:
This is because GCP libs authenticate with Application Default Credentials (ADC) https://cloud.google.com/docs/authentication/application-default-credentials

mbolt35:
Looks like these checks were added by a contributor: https://github.com/kubecost/cost-analyzer-helm-chart/pull/1260

(Which was broken). @michaelmdresser The funny part is that you realized it broke _something_ in cluster controller, so you fixed it: https://github.com/kubecost/cost-analyzer-helm-chart/commit/167626bfef7a69821e51224b0dd9673658dac1e3

But of course, as fate has it, here we are hitting rewind back to the original changeset. Thanks for looking into this!

mbolt35 (approved):
LGTM

PR #2203, opened 2023-04-29
Remove unnecessary ConfigMap for productKey when secretname is specified 

### LINKED ISSUE #2177: Remove unnecessary ConfigMap for `productKey`
When setting `.Values.kubecostProductConfigs.productKey.enabled`, the Helm template will always create a ConfigMap, even when you are providing your productKey via a Secret (`.Values.kubecostProductConfigs.productKey.secretname`).

Related to:
* https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.102/cost-analyzer/templates/cost-analyzer-pkey-configmap.yaml#L3
* https://github.com/kubecost/cost-analyzer-helm-chart/pull/2057

When I provide the productKey via secret, don't generate an empty configmap.

No functionality is currently broken, however there is an unnecessary resource created.

Cleaner UX

ivankube:
## What does this PR change?
When user provide the productKey via secret, don't generate an empty `product-configs` [configmap](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.102/cost-analyzer/templates/cost-analyzer-pkey-configmap.yaml#L3).

N/A

No direct impact on users

Fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/2177

## How was this PR tested?
Manually. If kubecostProductConfigs.productKey.secretname  exists, Kubecost is not  creating `product-configs` Configmap:

```
❯ kubectl get cm product-configs -n kubecost
Error from server (NotFound): configmaps "product-configs" not found`
```

N/A


thomasvn:
@ivankube Looks good! Do you think it's worth commenting in the [values.yaml](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.102/cost-analyzer/values.yaml#L991-L995) that if the `secretname` is specified, a configmap with the `key` will not be created? This can also be added as a comment to the `cost-analyzer-pkey-configmap.yaml`

ivankube:
@thomasvn great suggestion. I added the comments.

thomasvn:
@ivankube When testing this change, I found that the ConfigMap was still created even when `.Values.kubecostProductConfigs.productKey.key` was not set. Should we check to make sure it the key exists?

```
helm template ./cost-analyzer --set "kubecostProductConfigs.productKey.enabled=true" | less
```

If possible, it would also probably be good to halt the install here if `key` or `secretName` doesn't exit, but `.Values.kubecostProductConfigs.productKey.enabled=true`

ivankube:
great catch! I just added this condition.

thomasvn:
@ivankube Looks great, thanks!

thomasvn (approved):

PR #2204, opened 2023-04-29
Split grafana sidecar image name and image version config

### LINKED ISSUE #2191: Split grafana sidecar image name and image version config
I'm trying to update the `kiwigrid/k8s-sidecar` image to leverage a pull through cache registry, to remove dependencies on Docker Hub. Since the image name and version are [stored in the same key/value](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/charts/grafana/values.yaml#L259), updating this reference will prevent us from receiving automatic updates to this image version when it's incremented in the helm chart, and will require us to increment the image version manually as we upgrade helm chart versions.

Split the image name and version into separate key/value configs, similar to how [other images are being configured](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/charts/grafana/values.yaml#L27-L28) in this same chart.

Pinning to the specific image version and taking on the manual effort to keep it up to date when we upgrade helm charts. My team most likely will not allow this.

This would allow users to override the grafana sidecar image name without taking on the responsibility of updating image versions.

ivankube:
## What does this PR change?
This PR would allow users to override the Grafana sidecar image name without taking on the responsibility of updating image versions.

No

No direct impact on users. 

Fixes https://github.com/kubecost/cost-analyzer-helm-chart/issues/2191

## How was this PR tested?
Manually. Checked that Grafana sidecar still has the correct repo and tag:

```
❯ kubectl get deploy kubecost-grafana -n kubecost -o yaml | grep sidecar
        image: kiwigrid/k8s-sidecar:1.23.1
```

N/A

### DIFF
--- a/cost-analyzer/charts/grafana/templates/deployment.yaml
+++ b/cost-analyzer/charts/grafana/templates/deployment.yaml
@@ -73,8 +73,8 @@ spec:
       containers:
 {{- if .Values.sidecar.dashboards.enabled }}
         - name: {{ template "grafana.name" . }}-sc-dashboard
-          image: "{{ .Values.sidecar.image }}"
-          imagePullPolicy: {{ .Values.sidecar.imagePullPolicy }}
+          image: "{{ .Values.sidecar.image.repository }}:{{ .Values.sidecar.image.tag }}"
+          imagePullPolicy: {{ .Values.sidecar.image.pullPolicy }}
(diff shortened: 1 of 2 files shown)


thomasvn (approved):
Looks great, thanks!

PR #2209, opened 2023-05-01
Add schema for status to TurndownSchedule CRD

michaelmdresser:
## What does this PR change?
Fixes a bug where the CR (turndownschedule) status would never be updated because calls to UpdateStatus were no-ops. The "bad state" cleanup routine for turndownschedules would then delete the schedule because it thought it was in a bad state, but in fact we just failed to update the CR with the state being held by the controller.

With the schema, they are no longer no-ops and turndown completes successfully.

- Mirrors OSS: https://github.com/kubecost/cluster-turndown/pull/55

- https://kubecost.atlassian.net/browse/CORE-195

## How was this PR tested?

Installed Helm chart with CC enabled on a fresh GKE cluster. Scheduled turndown. Observed correct full schedule success -> turn down -> turn up -> reschedule cycle, including statuses on the TDS in-cluster.

GKE cluster config:
```
gcloud container clusters create \
    "<redacted>" \
    --project "<redacted>" \
    --region "us-central1-c" \
    --network "<redacted>" \
    --subnetwork "us-central1" \
    --release-channel "stable" \
    --num-nodes 2
```

I can provide other specific testing commands if desired.

Not necessary.

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -294,6 +294,34 @@ spec:
                 repeat:
                   type: string
                   enum: [none, daily, weekly]
+            status:
+              type: object
+              properties:
+                state:
+                  type: string
+                lastUpdated:
+                  format: date-time
+                  type: string
+                current:
+                  type: string
+                scaleDownId:
+                  type: string
+                nextScaleDownTime:
+                  format: date-time
+                  type: string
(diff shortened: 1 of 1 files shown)


michaelmdresser:
~Still have a problem with where we make new node and reschedule successfully, but then scale down repeats infinitely. Investigating.~ Just had to tweak the schema a bit to match the `json:` struct tags on the TDS struct definition.

mbolt35 (approved):
Looks good!

PR #2235, opened 2023-05-09
Add cluster-controller-turndown-config ConfigMap.

nik-kc:
## Release notes headline
* Adds the ConfigMap for nsturndown, to allow for the frontend to start work on their end.

## Does this PR relate to any other PRs?
* https://github.com/kubecost/cluster-controller/pull/40

## Does this PR address any GitHub or Zendesk issues?
* https://kubecost.atlassian.net/jira/software/c/projects/CORE/boards/1?modal=detail&selectedIssue=CORE-231

## How was this PR tested?
* Tested in new turndown test cluster, verified ConfigMap is created properly and handler returns 200.

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -245,9 +245,6 @@ spec:
           # partially or fully initialize based on the presence of these keys
           # and their validity.
           optional: true
-      - name: cluster-controller-turndown-conf
-        configMap:
-          name: cluster-controller-turndown-conf
 ---
 apiVersion: v1
 kind: Service
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -168,7 +168,7 @@ subjects:
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: cluster-controller-turndown-config
+  name: cluster-controller-nsturndown-config
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -165,6 +165,14 @@ subjects:
     name: {{ template "kubecost.clusterControllerName" . }}
     namespace: {{ .Release.Namespace }}
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-controller-turndown-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -237,6 +245,9 @@ spec:
           # partially or fully initialize based on the presence of these keys
           # and their validity.
           optional: true
+      - name: cluster-controller-turndown-conf
+        configMap:
+          name: cluster-controller-turndown-conf
```
ameijer: I don't think you need to mount this volume

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -165,6 +165,14 @@ subjects:
...
     namespace: {{ .Release.Namespace }}
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-controller-turndown-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
+---
```
ameijer, line 193: This is an interesting approach. I had been thinking that we would just let the mapstore create the object, but this works too. The difference in behavior is that now the configmap is managed by helm, so if the helm release is uninstalled, then this configmap would also be uninstalled since you are creating it here via helm. If you did not create this here, then the mapstore would create it, and it would not be managed by helm so would survive if the helm release was uninstalled. 

I am not sure what the right approach is - do we want the scheduled turndowns to survive a helm uninstall? Helm takes the approach of safety with PVCS, see https://github.com/helm/helm/issues/5156#issuecomment-812963730 ... they are not deleted with a helm delete. What do you think?

ameijer (changes requested):

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
nik-kc: Thanks for pointing that out, I've updated that.

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -165,6 +165,14 @@ subjects:
     name: {{ template "kubecost.clusterControllerName" . }}
     namespace: {{ .Release.Namespace }}
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-controller-turndown-config
```
michaelmdresser: Please don't use this name. Turndown is already a thing in Cluster Controller and it has a very specific associated with functionality (https://docs.kubecost.com/install-and-configure/advanced-configuration/controller/cluster-turndown), a repository https://github.com/kubecost/cluster-turndown, and a CRD https://github.com/kubecost/cost-analyzer-helm-chart/blob/1723242ea6744b2d20cdda6dea9a4a59ffe22faa/cost-analyzer/templates/kubecost-cluster-controller-template.yaml#L258-L273.

---

We could call it `nsturndown` or `ns-turndown` to disambiguate.

nik-kc: Fair enough, I'll update the names as such.

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
nik-kc: I can certainly see both sides. Given the comments in that thread, I'd lean towards maintaining the configmap on helm install, but I'm not entirely convinced. I can also see the benefit to removing the configmap, preventing a zombie configmap that may hang around forever, if let unchecked.

ameijer (approved):

ameijer:
@michaelmdresser do these changes look good to you?

michaelmdresser (approved):
LGTM. I'm curious how we'll feel about the ConfigMap choice in a little while; I'm not sure what the right answer is.

PR #2240, opened 2023-05-12
Document CloudCost in Helm chart. Default disabled.

thomasvn:
## What does this PR change?

Adds `cloudCost` feature to Helm chart.

- No

- More thoroughly documented `values.yaml` file

- 

## How was this PR tested?

```
# The following environment variables were created
$ helm template ./cost-analyzer
...
            - name: CLOUD_COST_ENABLED
              value: "false"
            - name: CLOUD_COST_IS_INCLUDE_LIST
              value: "false"
            - name: CLOUD_COST_LABEL_LIST
              value: ""
            - name: CLOUD_COST_TOP_N
              value: "1000"
...
```

No

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -418,6 +418,16 @@ kubecostModel:
   # For deploying kubecost in a cluster that does not self-monitor
   etlReadOnlyMode: false
 
+  ## Feature to view your out-of-cluster costs and their k8s utilization
+  ## Ref: https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer
+  cloudCost:
+    enabled: false
+    labelList:
+      isIncludeList: false
+      # format labels as comma separated string (ex. "label1,label2,label3")
+      labels: ""
+    topNItems: 1000
+
   allocation:
     # Enables or disables adding node labels to allocation data (i.e. workloads).
     # Defaults to "true" and starts with a sensible includeList for basics like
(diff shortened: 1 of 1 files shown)


thomasvn:
@kwombach12 Could you share your thoughts on surfacing these configs to `values.yaml`? They are currently [documented here](https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer), but not yet in the Helm Chart.

kwombach12:
@thomasvn I think its a good idea. @teevans @Sean-Holcomb do you see any issues with this?

MikeFindsThings:
Bump, customer [brought up missing values in values.yaml](https://kubecost.zendesk.com/agent/tickets/4208)

cost-analyzer/values.yaml
```diff
@@ -418,6 +418,16 @@ kubecostModel:
   # For deploying kubecost in a cluster that does not self-monitor
   etlReadOnlyMode: false
 
+  ## Feature to view your out-of-cluster costs and their k8s utilization
+  ## Ref: https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/cloud-costs-explorer
+  cloudCost:
+    enabled: false
+    labelList:
+      isIncludeList: false
```
jessegoodier: This should be a capital `I`
--set kubecostModel.cloudCost.labelList.IsIncludeList=true
no?

unless we want to change the template? because camelCase would be lowercase i?

thomasvn: @jessegoodier Great catch, thank you!

I'd prefer for it to be `isIncludeList` to be consistent with how all other configs are named. However I think it's best we don't adjust the template to do so. Don't want to cause any breaking changes for existing users already configuring `IsIncludeList`.

JesseGoodier (approved):
LGTM

PR #2241, opened 2023-05-12
fix: secret objects from csi secret provider

corax15:
## What does this PR change?
Previously when using the csi secret provider, the agent key secret has only been mounted directly to the agent pod. The prometheus pods, however, need the agent key to be stored in an actual k8s secret.
This change offers the possibility to specify corresponding information in the secret provider class.

No.

Users using the csi secret provider to retrieve the agent key yaml file from a secret store will have to specify secret object information in the secret provider class.

## How was this PR tested?
Test deployment in local cluster.

No.

### DIFF
--- a/cost-analyzer/templates/kubecost-agent-secretprovider-template.yaml
+++ b/cost-analyzer/templates/kubecost-agent-secretprovider-template.yaml
@@ -15,7 +15,11 @@ spec:
   provider: {{ required "Specify a valid provider." .Values.agentCsi.secretProvider.provider }}
   {{- if .Values.agentCsi.secretProvider.parameters }}
   parameters:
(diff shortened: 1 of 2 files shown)


jessegoodier:
thanks, this looks great to me.

Notes:
tested that does not interfere with an install without they key.
```
helm upgrade --install kubecost-agent \
  --kube-context $CONTEXT \
  . \
  --version 1.103.2 \
  --namespace kubecost-agent --create-namespace \
  -f values-agent.yaml \
  --set prometheus.server.global.external_labels.cluster_id=$CONTEXT \
  --set kubecostProductConfigs.clusterName=$CONTEXT \
  --set-file agentKey=kubecost-agent.key \
  --set agentKeySecretName=kubecost-agent-object-store
```

JesseGoodier (approved):
LGTM

PR #2268, opened 2023-05-23
Update cluster-controller ClusterRole RBAC permissions.

nik-kc:
## What does this PR change?

- Updates cluster-controlloer ClusterRole RBAC permissions for NS Turndown.

- [Yes](https://github.com/kubecost/cluster-controller/pull/48).

TODO

- [Yup.](https://kubecost.atlassian.net/browse/CORE-233)

### DIFF
--- a/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
+++ b/cost-analyzer/templates/kubecost-cluster-controller-template.yaml
@@ -71,6 +71,7 @@ rules:
       - watch
       - update
       - patch
+      - delete
   - apiGroups:
       - ''
     resources:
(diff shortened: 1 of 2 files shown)


ameijer (approved):
LGTM but would appreciate @michaelmdresser 's input as well...

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -169,6 +170,17 @@ rules:
...
+  # When cleaning a namespace, we need the ability to remove
+  # arbitrary resources (since we helm uninstall all releases in that NS first)
+  {{- if .Values.clusterController.namespaceTurndown.rbac.enabled }}
+  - apiGroups: ["*"]
+    resources: ["*"]
+    verbs:
+      - list
+      - get
+      - delete
+  {{- end }}
```
michaelmdresser, line 183: Pretty intense permissions! I think I understand why.

michaelmdresser (approved):

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
ameijer: yeah. We tried to keep these constrained as much as possible - but there is no way we can defend against people trying to delete Custom Resources and stuff like that. 

Our defensive approach was to at least make this controlled under its own toggle in the helm vars

PR #2315, opened 2023-06-07
federator pod template

nickcurie:
## What does this PR change?
Removed federator from cost-model, now runs in own pod

- https://github.com/kubecost/kubecost-cost-model/pull/1554

N/A

- https://kubecost.atlassian.net/browse/CORE-313

## How was this PR tested?
* Updated helm chart
* Observed typical federator logs in new federator pod
* Did not observe typical federator logs in cost-analyzer pod

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -252,13 +252,6 @@ spec:
         {{- end }}
         {{- end }}
         {{- end }}
-        {{- if .Values.federatedETL.federator }}
-        {{- if .Values.federatedETL.federator.enabled }}
-        - name: federator-config
-          configMap:
-            name: {{ template "cost-analyzer.fullname" . }}-federator
-        {{- end }}
-        {{- end }}
         {{- if .Values.extraVolumes }}
         # Extra volume(s)
         {{- toYaml .Values.extraVolumes | nindent 8 }}
@@ -484,12 +477,6 @@ spec:
             - name: {{ .Values.kubecostAdmissionController.secretName }}
               mountPath: /certs
             {{- end }}
-            {{- end }}
-            {{- end }}
-            {{- if .Values.federatedETL }}
-            {{- if .Values.federatedETL.federator.enabled }}
-            - name: federator-config
-              mountPath: /var/configs/federator
             {{- end }}
             {{- end }}
             {{- if .Values.kubecostProductConfigs }}
@@ -795,18 +782,6 @@ spec:
             {{- end}}
             {{- if .Values.federatedETL.redirectS3Backup }}
             - name: FEDERATED_REDIRECT_BACKUP
-              value: "true"
-            {{- end}}
-            {{- if .Values.federatedETL.useExistingS3Config }}
-            - name: FEDERATED_USE_EXISTING_CONFIG
-              value: "true"
-            {{- end}}
-            {{- if .Values.federatedETL.federator.enabled }}
-            - name: FEDERATED_FEDERATOR_ENABLED
-              value: "true"
-            {{- end}}
-            {{- if .Values.federatedETL.federator.useMultiClusterDB }}
-            - name: CURRENT_CLUSTER_ID_FILTER_ENABLED
               value: "true"
             {{- end}}
             - name: ETL_STORE_READ_ONLY
--- a/cost-analyzer/templates/federator-pod-template.yaml
+++ b/cost-analyzer/templates/federator-pod-template.yaml
@@ -1,0 +1,66 @@
+{{- if .Values.federatedETL.federator }}
+{{- if .Values.federatedETL.federator.enabled }}
+apiVersion: v1
+kind: Pod
+metadata: 
+  name: {{ template "cost-analyzer.fullname" . }}-federator
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: federator
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app: federator
+spec: 
+  restartPolicy: Always
+  volumes: 
+    - name: federator-config
+      configMap:
+        name: {{ template "cost-analyzer.fullname" . }}-federator
+    {{- if .Values.kubecostModel.federatedStorageConfigSecret }}
+    - name: federated-storage-config
+      secret:
+        defaultMode: 420
+        secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret }}
+    {{- end }}
+  containers:
+    - name: federator
+      {{- if .Values.kubecostModel }}
+      {{- if .Values.kubecostModel.fullImageName }}
+      image: {{ .Values.kubecostModel.fullImageName }}
+      {{- else if .Values.imageVersion }}
+      image: {{ .Values.kubecostModel.image }}:{{ .Values.imageVersion }}
+      {{- else }}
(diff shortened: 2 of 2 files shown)


cost-analyzer/templates/federator-pod-template.yaml
```diff
@@ -0,0 +1,80 @@
...
+      secret:
+        defaultMode: 420
+        secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret }}
+    {{- end }}
+  containers:
+    - name: federator
+      {{- if .Values.kubecostModel }}
+      {{- if .Values.kubecostModel.openSourceOnly }}
+      image: quay.io/kubecost1/kubecost-cost-model:{{ .Values.imageVersion }}
+      {{- else if .Values.kubecostModel.fullImageName }}
```
Sean-Holcomb: I don't think there is an openSourceOnly version of federator

cost-analyzer/templates/federator-pod-template.yaml
```diff
@@ -0,0 +1,80 @@
...
+          path: /healthz
+          port: 9001
+        initialDelaySeconds: 30
+        periodSeconds: 10
+        failureThreshold: 200
+      env:
+        - name: CONFIG_PATH
+          value: /var/configs/
+        - name: DB_PATH
+          value: /var/db/
```
Sean-Holcomb, line 60: We have this value in two env variables, lets remove the one we are not using

cost-analyzer/templates/federator-pod-template.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - name: ETL_FILE_STORE_ENABLED
+          value: "true"
+        - name: ETL_TO_DISK_ENABLED
+          value: "true"
```
Sean-Holcomb: can remove these values

cost-analyzer/templates/federator-pod-template.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - name: FEDERATED_STORE_CONFIG
+          value: "/var/configs/etl/federated/federated-store.yaml"
+        {{- end }}
+        {{- if or .Values.federatedETL.federatedCluster .Values.kubecostModel.federatedStorageConfigSecret }}
+        - name: FEDERATED_CLUSTER
+          value: "true"
+        {{- end}}
+        - name: ETL_PATH_PREFIX
+          value: "/var/db"
+        - name: CLOUD_PROVIDER_API_KEY
```
Sean-Holcomb: I don't think we need this one either

cost-analyzer/templates/federator-pod-template.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - name: ETL_FILE_STORE_ENABLED
+          value: "true"
+        - name: ETL_TO_DISK_ENABLED
+          value: "true"
+        {{- if .Values.kubecostModel.federatedStorageConfigSecret }}
+        - name: FEDERATED_STORE_CONFIG
+          value: "/var/configs/etl/federated/federated-store.yaml"
+        {{- end }}
+        {{- if or .Values.federatedETL.federatedCluster .Values.kubecostModel.federatedStorageConfigSecret }}
+        - name: FEDERATED_CLUSTER
```
Sean-Holcomb: The federator shouldn't need to know if it is a federated cluster

Sean-Holcomb (commented):
Looks good, there are just some extra stuff that I don't think the new template needs

Sean-Holcomb (commented):
There are still a couple of env variables on here that I don't think are necessary

Sean-Holcomb (approved):
Looks good, if you can just squash the commits and force push to get rid of the git guardian flags

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -797,18 +784,6 @@ spec:
             - name: FEDERATED_REDIRECT_BACKUP
               value: "true"
             {{- end}}
-            {{- if .Values.federatedETL.useExistingS3Config }}
-            - name: FEDERATED_USE_EXISTING_CONFIG
-              value: "true"
-            {{- end}}
```
thomasvn, line 803: @nickcurie Was deleting this line intentional? I believe there are existing users who are using this config. This may break their deployment.

cc @kaelanspatel @jessegoodier Thoughts?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -797,18 +784,6 @@ spec:
...
-            {{- if .Values.federatedETL.federator.enabled }}
-            - name: FEDERATED_FEDERATOR_ENABLED
-              value: "true"
-            {{- end}}
-            {{- if .Values.federatedETL.federator.useMultiClusterDB }}
-            - name: CURRENT_CLUSTER_ID_FILTER_ENABLED
-              value: "true"
-            {{- end}}
```
thomasvn, line 811: @nickcurie I see you delete this config here, but you're not adding it to the federator's template? Is that intentional?

cost-analyzer/templates/federator-pod-template.yaml
```diff
@@ -0,0 +1,66 @@
+{{- if .Values.federatedETL.federator }}
+{{- if .Values.federatedETL.federator.enabled }}
+apiVersion: v1
+kind: Pod
```
thomasvn, line 4: @nickcurie In most cases, we don't want to be directly deploying a pod to a k8s cluster. Instead, we want to be creating a [`Deployment`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) which will manage a set of pods.

The pod spec you have here is good, but I'd recommend you wrapping a Deployment spec around this pod spec. You can use the [QSR template](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.104/cost-analyzer/templates/query-service-deployment-template.yaml) as reference, but instead of a StatefulSet I'd recommend a Deployment!

cc @kaelanspatel @jessegoodier @srpomeroy Thoughts?

jessegoodier:
all good questions @thomasvn 
We should also consider the helm templating added here if you update the pod/deployment spec: 
https://github.com/kubecost/cost-analyzer-helm-chart/pull/2370

cost-analyzer/templates/federator-pod-template.yaml
srpomeroy: Prefer not to have uncontrolled workloads. I agree; Deployment likely makes the most sense.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
srpomeroy: I believe this is intentional, as we don't recommend using ETL Backup with Federated-ETL.

thomasvn: @srpomeroy Ah ok understood. We may need to remove this from our docs, and also include a message in release notes for those who may be using this config.

cc @bstuder99 @cliffcolvin

jessegoodier: This has been the recommended config for some time: 
https://github.com/kubecost/poc-common-configurations/commit/3429a820678590ec94223361f20dd39f50c8fdba
that is from Jan 18.

https://docs.kubecost.com/install-and-configure/install/multi-cluster/federated-etl#using-an-existing-object-store.yaml
excerpt: 
image

Issue #2325, opened 2023-06-10
Add Extra Manifest


speedfl:
**What problem are you trying to solve?**
Lot of charts propose to add extra manifests. This allows for example to dynamically create Secret management resources, networking resources, policies with the chart itself without having to create an umbrella chart.

**Describe the solution you'd like**
Add an `extraObjects` node which takes as input kube manifests

**Describe alternatives you've considered**
Create umbrella chart :/
Install the chart and then extra manifests :/

**How would users interact with this feature?**
NA

kwombach12:
@speedfl Thanks for suggesting this and opening a PR! @AjayTripathy @teevans is this something you can review?

jessegoodier:
Interesting suggestion!

Is this simply adding a template like below? If so, I can see how this could be useful.

Example:
`extra-manifests.yaml`
```yaml
{{- range .Values.extraManifests }}
---
{{ tpl (toYaml .) $ }}
{{- end }}
```

with a sample file of:
`configmap-networkcosts.yaml`
```yaml
extraManifests:
  - apiVersion: v1
    kind: ConfigMap
    metadata:
      name: network-costs-config
      namespace: staging-app
    data:
      config.yaml: |
        destinations:
          cross-region: []
          direct-classification: []
          in-region: []
          in-zone:
          - 127.0.0.0/8
          - 169.254.0.0/16
          - 10.0.0.0/8
          - 172.16.0.0/12
          - 192.168.0.0/16
          internet: []
        services:
          amazon-web-services: false
          azure-cloud-services: false
          google-cloud-services: false
```

test:
```sh
helm template kubecost ./cost-analyzer -f .idea/configmap-networkcosts.yaml.yaml
```

speedfl:
Yes the goal is to allow adding extra manifest for custom needs. Example if you are using a service mesh, external secret operator or any specific needs 🙂

jessegoodier:
Simple enough. created PR #2330

speedfl:
I already created this one 😄  https://github.com/kubecost/cost-analyzer-helm-chart/pull/2326 when I opened the ticket

jessegoodier:
Good to close? will be in 1.105.

speedfl:
PR has been merged so I would say yes!

PR #2326, opened 2023-06-10
fix(2325): Add extra manifests

speedfl:
## What does this PR change?

Add extra manifests to the chart

N/A

Add helm template to allow extra manifests to be passed during install/upgrades.

- #2325 

## How was this PR tested?

Yes. 
- No extra manifests in the values: no new manifests
- Extra manifests in the values: new manifests appended at the end

In the values

### DIFF
--- a/cost-analyzer/templates/extra-manifests.yaml
+++ b/cost-analyzer/templates/extra-manifests.yaml
@@ -1,0 +1,8 @@
+{{ range .Values.extraObjects }}
+---
+{{- if typeIs "string" . }}
+    {{- tpl . $ }}
+{{- else }}
+    {{- tpl (toYaml .) $ }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


jessegoodier:
can you add a comment to this section?
How does this PR impact users? (This is the kind of thing that goes in release notes!)

something like 
```
Add helm template to allow extra manifests to be passed during install/upgrades.
```

jessegoodier:
I have tested and it works as expected both with and without passing `extraObjects`

speedfl:
@jessegoodier all done :)

thomasvn:
@speedfl I just checked out the branch, and also ran `helm template ./cost-analyzer` to test. Looks good to me, thanks!

thomasvn (approved):

PR #2329, opened 2023-06-12
Impl query replica fixes for large scale PoC

ameijer:
(cherry picked from commit 53208349cf1fa771d80c5df7b630f4ee3ad16764)

## What does this PR change?

Adds persistence to query replicas, allows for dedicated query replicas secret, increases timeout

- #2328 
- 

query replicas can now point at federated data

- 
-

### DIFF
--- a/cost-analyzer/templates/query-service-deployment-template.yaml
+++ b/cost-analyzer/templates/query-service-deployment-template.yaml
@@ -9,6 +9,7 @@ metadata:
     {{ include "query-service.commonLabels" . | nindent 4 }}
 spec:
   replicas: {{ .Values.kubecostDeployment.queryServiceReplicas }}
+  serviceName: "query-service"
   selector:
     matchLabels:
       app.kubernetes.io/name: query-service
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -47,7 +62,7 @@ spec:
           volumeMounts:
             - name: persistent-configs
               mountPath: /var/configs
-            - name: persistent-db
+            - name: storage
               mountPath: /var/db
           securityContext:
             runAsUser: 0
```
jessegoodier, line 70: is user 0 needed?

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -47,7 +62,7 @@ spec:
           volumeMounts:
             - name: persistent-configs
               mountPath: /var/configs
-            - name: persistent-db
+            - name: storage
```
jessegoodier: can we use a name that reflects the purpose?

cost-analyzer/templates/query-service-deployment-template.yaml
ameijer: @jessegoodier I actually think it is. This is an init container that sets volume perms before the query replica runs. User 0 will allow it to take volumes with arbitrary perms and ensure they are writable by the pod

nik-kc (approved):

PR #2331, opened 2023-06-12
option to disable trends

JesseGoodier:
## What does this PR change?

add option to disable the trends enpoint to reduce cpu and memory load in large envrionments.

No

add option to disable the trends enpoint to reduce cpu and memory load in large envrionments.

## How was this PR tested?
local helm install

TBD


cliffcolvin (approved):
LGTM

nealormsbee (commented):
🙌

PR #2334, opened 2023-06-13
InitContainer Default Value Correction on KubeCost Deployment

hamidyousefi:
## What does this PR change?
Only Fix missed empty array in front of initContainers

No

IDE Linters does not show errors anymore on this line.

None

## How was this PR tested?
Tested on Self Hosted OKD

## Have you made an update to the documentation?
No need.

### DIFF
--- a/kubecost.yaml
+++ b/kubecost.yaml
@@ -20873,7 +20873,7 @@ spec:
         - name: persistent-configs
           persistentVolumeClaim:
             claimName: kubecost-cost-analyzer
-      initContainers:
+      initContainers: []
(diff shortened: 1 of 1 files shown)


thomasvn:
@hamidyousefi Thanks for the PR! Note, that the `kubecost.yaml` file is automatically generated from a `helm template` command. Therefore even if we merge & fix for the next release, it would be overwritten in future releases. Do you have any recommendations here?

hamidyousefi:
Then it can be closed for now. I will check the Helm Repository.

PR #2353, opened 2023-06-14
Add the capability to override the key inside the GCP secret

gschlueter-jaconi:
## What does this PR change?
Adds the ability to override the previously hard-coded GCP secret key (compute-viewer-kubecost-key.json). The reasoning behind the change is that sometimes we don't control the creation of secrets. In our case, we created a GCP Service Account Key using the Config Connector (https://cloud.google.com/config-connector/docs/reference/resource-docs/iam/iamserviceaccountkey). The created K8s secret will always contain the key "key.json".

- 
- 

Adds the ability to override the key inside the GCP secret. The default remains "compute-viewer-kubecost-key.json".

- 
- 

## How was this PR tested?
Rendered the chart locally using multiple scenarios.

To the values.yaml

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -137,7 +137,7 @@ spec:
           secret:
             secretName: {{ .Values.kubecostProductConfigs.gcpSecretName }}
             items:
-            - key: compute-viewer-kubecost-key.json
+            - key: {{ .Values.kubecostProductConfigs.gcpSecretKeyName | default "compute-viewer-kubecost-key.json" }}
               path: service-key.json
         {{- end }}
         {{- end -}}
--- a/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml
+++ b/cost-analyzer/templates/kubecost-metrics-deployment-template.yaml
@@ -77,7 +77,7 @@ spec:
           secret:
             secretName: {{ .Values.kubecostProductConfigs.gcpSecretName }}
             items:
-            - key: compute-viewer-kubecost-key.json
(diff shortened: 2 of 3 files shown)


thomasvn:
@gschlueter-jaconi Thanks for the PR! I'm good with this change.

Could you share the results of how you tested the chart locally? Also, could you share how k8s secrets are created in your environment and why they default to `key.json`?

gschlueter-jaconi:
I just tested my change using `helm template` as follows:

1. Test without setting the key (defaults to `compute-viewer-kubecost-key.json`):
```
helm template --set kubecostProductConfigs.gcpSecretName=my-secret . | grep -C2 my-secret
```
Output:
```
        - name: gcp-key-secret
          secret:
            secretName: my-secret
            items:
            - key: compute-viewer-kubecost-key.json
```
2. Test setting the key (key is now `my-secret-key`):
```
helm template --set kubecostProductConfigs.gcpSecretName=my-secret --set kubecostProductConfigs.gcpSecretKeyName=my-secret-key . | grep -C2 my-secret
```
Output:
```
        - name: gcp-key-secret
          secret:
            secretName: my-secret
            items:
            - key: my-secret-key
              path: service-key.json
        - name: persistent-configs
```

As for how the secrets are created in our environment: We're using the Config Connector (https://cloud.google.com/config-connector/docs/overview) to create GCP Service Accounts, including Service Account Keys (https://cloud.google.com/config-connector/docs/reference/resource-docs/iam/iamserviceaccountkey).
The Config Connector creates a K8s secret for such keys. However, you cannot specify the name of the key inside this secret. It will always be `key.json`.
There's also an open issue for that problem: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/42

thomasvn:
Thanks @gschlueter-jaconi, this looks good to me!

thomasvn (approved):

PR #2357, opened 2023-06-14
option to use federated store

JesseGoodier:
prefer fed store over etl-backups. fixes crash if both are configured.


cliffcolvin (approved):
LGTM

PR #2379, opened 2023-06-21
update helm chart to support savings cache disable

ameijer:
## What does this PR change?

provides a way to disable the savings cache via helm vars

- 
- 

re-introduces a way for users to disable the savings cache

- 
- 

## How was this PR tested?

placed in cluster, confirmed ram usage was lower with savings disabled

none needed, this uses the existing documented parameters


cliffcolvin (approved):
LGTM!

PR #2388, opened 2023-06-25
Add template to ingress hostname

Matan-Tzuberi:
## What does this PR change?

adding a support to template the ingress hostname

no

shouldn't impact

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingress-template.yaml
@@ -41,7 +41,7 @@ spec:
   {{- end }}
 {{- end }}
   rules:
-  {{- range .Values.ingress.hosts }}
+  {{- range tpl .Values.ingress.hosts . }}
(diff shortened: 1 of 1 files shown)


thomasvn:
Thanks @Matan-Tzuberi! Have you tested the following change using `helm template` and running it in your environment? If so, could you share results?

I did a quick test and found that the [`tpl`](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) function expects a string, whereas currently `.Values.ingress.hosts` is currently a list. Therefore I received an error when using the following configuration:

```yaml
ingress:
  enabled: true
  hosts:
  - 'kubecost.xyz'
  - 'kubecost.io'

# Error: template: cost-analyzer/templates/cost-analyzer-ingress-template.yaml:44:23: executing "cost-analyzer/templates/cost-analyzer-ingress-template.yaml" at <.Values.ingress.hosts>: wrong type for value; expected string; got []interface {}
```

PR #2399, opened 2023-06-28
Add comment to defaultModelPricing

ivankube:
If defaultModelPricing is set to 0. it must to be in quotes "0.00"

Failure to do so will result in the value(s) not being written to the Kubecost cost-model's PV

## What does this PR change?
Add one comment to defaultModelPricing

No

It provides some clarity on using zero default costs via defaultModelPricing

https://kubecost.zendesk.com/agent/tickets/4433

## How was this PR tested?
Locally

N/A

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1017,7 +1017,7 @@ costEventsAudit:
 #       pass: admin
 #   - name: "Cluster B"
 #     address: http://cluster-b.kubecost.com:9090
-#  defaultModelPricing: # default monthly resource prices, used predominately for on-prem clusters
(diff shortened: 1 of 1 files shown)


thomasvn:
Thanks @ivankube! This looks good to me. Could you provide more explanation on what the behavior is when setting `0` or `0.0` without quotes? Is it because it's being interpreted as boolean false here?

https://github.com/kubecost/cost-analyzer-helm-chart/blob/53836363707d815c3846f9847cdc98be4ed70cc2/cost-analyzer/templates/cost-analyzer-pricing-configmap.yaml#L15-L17

thomasvn (approved):

ivankube:
@thomasvn correct - 0 or 0.00 without quotes is being interpreted as false.

PR #2419, opened 2023-07-07
Fix Network Costs DaemonSet resource limits

lukasgomez:
## What does this PR change?
The comment in the Kubecost Helm Chart where indicates how to remove the limits for the Network Costs DaemonSet is wrong. It says to remove the limits you just have to set `limits: {}`, but this solution doesn't work because the parent chart has a default value for the cpu limits. The real solution to remove the limits (for CPU, because for memory the parent chart doesn't set any value), is set `cpu: null`

Old comment:
image

New comment:
image

- No

This will allow users understand how to remove the limits for the Network Costs DaemonSet correctly.

- N/A
- 
## How was this PR tested?

Rendering the chart locally with a Helm template and deploying it in a real cluster. 

No, the proper update in the documentation should be the comment in the Helm Chart

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -727,7 +727,7 @@ networkCosts:
   port: 3001
   # this daemonset can use significant resources on large clusters: https://guide.kubecost.com/hc/en-us/articles/4407595973527-Network-Traffic-Cost-Allocation
   resources:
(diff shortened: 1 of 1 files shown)


jessegoodier:
nice find. thank you!

JesseGoodier (approved):
Thanks!

PR #2429, opened 2023-07-11
Update regex for node exporter

Danial968:
## What does this PR change?

Reduces the strictness of regex for Prometheus to scrape the node-exporter.

No

Update prometheus scrape config to scrape existing node exporters. 

- #1975 

## How was this PR tested?

Running kubecost directly with updated configmap.

### DIFF
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -1297,7 +1297,7 @@ serverFiles:
             regex: true
           - source_labels: [__meta_kubernetes_endpoints_name]
             action: keep
-            regex: (.*kube-state-metrics|.*prometheus-node-exporter|kubecost-network-costs)
+            regex: (.*kube-state-metrics|.*node-exporter|kubecost-network-costs)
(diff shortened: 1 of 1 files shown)


thomasvn:
Thanks for the PR @Danial968! The only edge case I can think of is if a user is running `svc/my-node-exporter` which they didn't want to be scraped by Kubecost, but would now be scraped by Kubecost with this change. However, I anticipate the impact to Kubecost and Prometheus would be low.

@ivankube @jcharcalla @keithhand could you provide comment based on your experience as well? Given this would impact all users, I want to make sure we give it a close look.

jcharcalla:
I've not encountered any users that did not want to scrape an existing node exporter.

Change LGTM as I have instructed users to modify this in the past.

ivankube:
+1. LGTM.

thomasvn (approved):

jessegoodier:
thanks @Danial968- will be in 1.106.

also closes:
https://github.com/kubecost/cost-analyzer-helm-chart/pull/2444

Issue #2425, opened 2023-07-10
Cloud Cost reports in values.yam not generating a report


ChrisMusgrove:
Adding Cloud Cost report to the values.yaml file does not generate a cloud cost report in the ui

**To Reproduce**  

1. Added following to values.yaml:
 assetReports:
    enabled: true # If true, overwrites report parameters set through UI
    reports:
    - title: "Example Asset Report 0"
      window: "7d"
      aggregateBy: "label:squads"
      chartDisplay: "category"
      idle: "separate"
      rate: "cumulative"
      accumulate: false # daily resolution
      # filters:
      #   - property: "cluster"
      #     value: "cluster-one"

 cloudCostReports:
    enabled: false # If true, overwrites report parameters set through UI
    reports:
      - title: "Cloud Cost Report 0"
        window: "today"
        aggregateBy: "type"
        accumulate: false # daily resolution

3. push to our kubecost setup 

4. Asset report generates successfully, no cloud cost report generates

**Expected behavior**  
Cloud Cost report is generated in Kubecost UI

**What impact will this have on your ability to get value out of Kubecost?**
Unable to generate cloud cost reports, with Asset reports being deprecated will be unable to generate required reporting for costs

**Bug report**
[kubecost-bug-report.txt](https://github.com/kubecost/cost-analyzer-helm-chart/files/12000381/kubecost-bug-report.txt)

**Please share the support case, if any**  
4533

kirbsauce:
@kwombach12 , diagnosis by support team feels this needs reviewed by engineering.

kwombach12:
Thanks @kirbsauce. We will take a look!

avrodrigues5:
Hello @ChrisMusgrove ,

Please give cloudCostReport in values.yaml as below

``` 
    cloudCostReports:
      enabled: true # If true, overwrites report parameters set through UI
      reports:
        - title: "Cloud Cost Report 0"
          window: "7d"
          aggregateBy: "service"
          # filters: 
          #   - property: "service"
          #     value: "service-as-seen-in-Cloud Cost UI" 
```

Filter is optional but you are more than welcome to try it. 

https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/saved-reports#managing-reports-via-values.yaml, Per the doc the enabled should be set to true and I can confirm assetReport block that you pasted if enabled is false there it wont create a asset Report as well. 

[...]

PR #2433, opened 2023-07-12
update values.yaml to reflect right aggregate by option for cloudcost report

avrodrigues5:
## What does this PR change?

Aggregate by type is not a valid aggregate by option for cloud cost. It should be one of service, accountID,  provider, providerID or label. When user gives current option in values.yaml in case of [2425](https://github.com/kubecost/cost-analyzer-helm-chart/issues/2425) it leads to a log error in cost-model pod. This is the PR to provide right example in values.yaml.

None
- 
- 

None

https://github.com/kubecost/cost-analyzer-helm-chart/issues/2425
- 
- 

## How was this PR tested?
On Azure cluster creating a cloud cost report with a yaml satisfying below values
```
  cloudCostReports:
    enabled: true # If true, overwrites report parameters set through UI
    reports:
      - title: "Cloud Cost Report 0"
        window: "7d"
        aggregateBy: "service"
        # filters: 
        #   - property: "cluster"
        #     value: "cluster-one"
```

No need

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -207,11 +207,11 @@ global:
     reports:
       - title: "Cloud Cost Report 0"
         window: "today"
-        aggregateBy: "type"
-        accumulate: false # daily resolution
-        filters:
-          - property: "cluster"
-            value: "cluster-one"
+        aggregateBy: "service"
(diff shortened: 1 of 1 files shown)


ghost (approved):
Looks good to me

PR #2438, opened 2023-07-13
Federator runs as separate deployment by default

thomasvn:
## What does this PR change?

1. Federator runs as separate deployment **by default**.
2. Remove federator configs from cost-analyzer-deployment.

## Does this PR rely on any other PRs?

This PR amends the previous 2 PRs: https://github.com/kubecost/cost-analyzer-helm-chart/pull/2397 https://github.com/kubecost/cost-analyzer-helm-chart/pull/2315

## How does this PR impact users? (This is the kind of thing that goes in release notes!)

If you are running the Federated ETL architecture, when upgrading to v1.105 you will see a new `deployment/kubecost-federator` which is a workload dedicated to federating Kubecost's ETL files.

## Links to Issues or ZD tickets this PR addresses or fixes

- 

## How was this PR tested?

As shown, if the Federator is enabled a new `federator-deployment-template.yaml` is rendered

```sh
$ helm template ~/kubecost/cost-analyzer-helm-chart/cost-analyzer --set federatedETL.federator.enabled=true | grep -C3 federator-deployment

            periodSeconds: 10
            failureThreshold: 200
---
# Source: cost-analyzer/templates/federator-deployment-template.yaml
apiVersion: apps/v1
kind: Deployment
metadata:

```

Created a new cluster running Kubecost Federated ETL. `gcloud container clusters get-credentials thomasn-federator-microservice --region us-west1-a`. Confirmed that the federator was running and placing federated ETL files in the `/combined` directory on the bucket.

```sh
$ helm upgrade -i kubecost ~/kubecost/cost-analyzer-helm-chart/cost-analyzer -f ./values.yaml
```

```yaml
# values.yaml
kubecostProductConfigs:
  clusterName: "thomasn-federator-microservice"
  cloudIntegrationSecret: "cloud-integration"
kubecostModel:
  federatedStorageConfigSecret: "kubecost-fedetl"
  fullImageName: "gcr.io/kubecost1/cost-model:prod-1.105.0-rc.10"
  extraEnv:
  - name: "LOG_LEVEL"
    value: "debug"
federatedETL:
  federatedCluster: true
  federator:
    enabled: true
    primaryClusterID: "thomasn-federator-microservice"
  primaryCluster: true
prometheus:
  server:
    global:
      external_labels:
        cluster_id: "thomasn-federator-microservice"
```

## Have you made an update to documentation?

Not yet.

PR #2439, opened 2023-07-14
[Cherry-pick] Federator runs as separate deployment by default (#2438)

thomasvn:
Cherry-pick #2438 into v1.105 branch

* Federator runs as separate deployment by default. Remove federator configs from cost-analyzer-deployment.
* serviceAccount for irsa


cliffcolvin (approved):
LGTM

PR #2444, opened 2023-07-17
To meet node exporter customized values

kalyanteja19:
## What does this PR change?

To meet node exporter customized values. During every kubecost upgrade, to avoid manual update of node exporter endpoint

-  No

This will ease maintenance efforts that has dependency on node-exporter name

- NA
- NA

## How was this PR tested?
This regex expression change, hence this should not cause any issue

No


jessegoodier:
@kalyanteja19 thank you for the contribution- we actually had a request to do this recently. the change will be in 1.106

PR #2302, opened 2023-06-02
Include Thanos endpoint for Grafana Datasource if configured

srpomeroy:
## Does this PR rely on any other PRs?

- 
- 

## Links to Issues or ZD tickets this PR addresses or fixes

- 
-

PR #2466, opened 2023-08-02
grafana- default datasource for thanos / update datasources logic

JesseGoodier:
## What does this PR change?

Change default Grafana datasource to thanos if enabled. 
Update Grafana datasource for Prometheus to allow for Grafana interval variables to work properly

- No

Change default Grafana datasource to thanos if enabled. 
Update Grafana datasource for Prometheus to allow for Grafana interval variables to work properly

## How was this PR tested?

1. Thanos install
2. Prometheus install
3. BYO Prometheus install

No changes needed

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -935,6 +935,15 @@ grafana:
   rbac:
     # Manage the Grafana Pod Security Policy
     pspEnabled: false
+  datasources:
+    datasources.yaml:
+      apiVersion: 1
+      datasources:
+        - name: prometheus-kubecost
+          type: prometheus
+          url: http://kubecost-prometheus-server.kubecost.svc.cluster.local
+          access: proxy
+          isDefault: false
   sidecar:
     dashboards:
       enabled: true
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -863,9 +863,9 @@ kubecostDeployment:
   # Ref: https://docs.kubecost.com/install-and-configure/advanced-configuration/query-service-replicas
   queryServiceReplicas: 0
   queryService:
-    resources: 
-      requests: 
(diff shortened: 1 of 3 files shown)


cost-analyzer/values.yaml
```diff
@@ -935,15 +935,6 @@ grafana:
   rbac:
     # Manage the Grafana Pod Security Policy
     pspEnabled: false
-  datasources:
```
thomasvn, line 938: @jessegoodier Could you explain the reasoning behind removing `.Values.datasources`? I notice it's used here as well:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/3a283b62bb914daf45bc4ead079eddbe0612e212/cost-analyzer/charts/grafana/templates/configmap.yaml#L24-L29

If we do decide to get rid of `datasources.yaml` we should also remove the above code snippet and test?

jessegoodier: I was struggling with the logic.

Can you take a look at the configmap logic now?

srpomeroy:
We also discussed removing the data source defined in the main values file as it duplicates the data sources.
https://github.com/kubecost/cost-analyzer-helm-chart/pull/2302/files#diff-e232ea8a3d165b37f6212249d3718bb96e05b64a591f1fa36d77c9448c747527

thomasvn:
@jessegoodier These changes look good to me! I've also run `helm template` against these changes, and all templates render properly when no additional `values.yaml` is passed.

thomasvn (approved):

jessegoodier:
> We also discussed removing the data source defined in the main values file as it duplicates the data sources. https://github.com/kubecost/cost-analyzer-helm-chart/pull/2302/files#diff-e232ea8a3d165b37f6212249d3718bb96e05b64a591f1fa36d77c9448c747527

done! thanks for checking on that.

PR #2471, opened 2023-08-04
Add ability to set `containerSecurityContext` for Grafana and override `securityContext` for cost-analyzer

JorisAndrade:
## What does this PR change?

- This PR adds missing `containerSecurityContext` to Grafana 
- It also the ability to overrides properties of the default `securityContext` for the `cost-analyzer`

Add ability to set `containerSecurityContext` for Grafana and override `securityContext` for the `cost-analyzer`

## How was this PR tested?

```
$ helm template --set 'grafana.containerSecurityContext.allowPrivilegeEscalation=false' cost-analyzer --show-only charts/grafana/templates/deployment.yaml | yq '.spec.template.spec.containers.[].securityContext'
```

```
$ helm template cost-analyzer --show-only charts/grafana/templates/deployment.yaml | yq '.spec.template.spec.containers.[].securityContext'
null
null
```

```
$ helm template cost-analyzer --show-only templates/cost-analyzer-deployment-template.yaml | yq '.spec.template.spec.securityContext'

fsGroup: 1001
runAsGroup: 1001
runAsUser: 1001
```

```
$ helm template --set 'global.securityContext.runAsNonRoot=true' cost-analyzer --show-only templates/cost-analyzer-deployment-template.yaml | yq '.spec.template.spec.securityContext'

fsGroup: 1001
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
```

```
$ helm template --set 'global.securityContext.runAsUser=2001' cost-analyzer --show-only templates/cost-analyzer-deployment-template.yaml | yq '.spec.template.spec.securityContext'

fsGroup: 1001
runAsGroup: 1001
runAsUser: 2001
```

Yes!

### DIFF
--- a/cost-analyzer/charts/grafana/README.md
+++ b/cost-analyzer/charts/grafana/README.md
@@ -36,6 +36,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `livenessProbe`            | Liveness Probe settings                           | `{ "httpGet": { "path": "/api/health", "port": 3000 } "initialDelaySeconds": 60, "timeoutSeconds": 30, "failureThreshold": 10 }` |
 | `readinessProbe`            | Rediness Probe settings                           | `{ "httpGet": { "path": "/api/health", "port": 3000 } }`|
 | `securityContext`               | Deployment securityContext                    | `{"runAsUser": 472, "fsGroup": 472}`                    |
+| `containerSecurityContext`      | Container securityContext                     | `{}`                                                    |
 | `image.repository`              | Image repository                              | `grafana/grafana`                                       |
 | `image.tag`                     | Image tag. (`Must be >= 5.0.0`)               | `5.3.1`                                                 |
 | `image.pullPolicy`              | Image pull policy                             | `IfNotPresent`                                          |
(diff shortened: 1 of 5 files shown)


jessegoodier:
Nice work!
We are actually in the process of adding pod/container security contexts to make all pods run as nonRoot by default.

Do you think that a global section for this would work for your use case? 

like this, with templating in all pods:
```
global:
  securityContext:
    fsGroup: 65534
    runAsGroup: 65534
    runAsNonRoot: true
    runAsUser: 65534
    seccompProfile:
      type: RuntimeDefault
  containerSecurityContext:
    readOnlyRootFilesystem: true
    capabilities:
      drop:
      - ALL
```

JorisAndrade:
This would work yeah! As long as we can overload to put what we need (like seccomp in our case)

JorisAndrade:
> Nice work! We are actually in the process of adding pod/container security contexts to make all pods run as nonRoot by default.
> 
> Do you think that a global section for this would work for your use case?
> 
> like this, with templating in all pods:
> 
> ```
> global:
>   securityContext:
>     fsGroup: 65534
>     runAsGroup: 65534
>     runAsNonRoot: true
>     runAsUser: 65534
>     seccompProfile:
>       type: RuntimeDefault
>   containerSecurityContext:
>     allowPrivilegeEscalation: false
>     readOnlyRootFilesystem: true
>     capabilities:
>       drop:
>       - ALL
> ```

Do you have any ETA on this? Would like to avoid having to keep a fork on our side

jessegoodier:
We are attempting to get this in 1.106. Would love your feedback on:

https://github.com/kubecost/cost-analyzer-helm-chart/pull/2493

and if this would fit your needs.

PR #2475, opened 2023-08-07
grafana request sizing dashboard

JesseGoodier:
## What does this PR change?

use max for cpu and memory queries
fix variable queries
update to latest grafana components

## Does this PR rely on any other PRs?

- https://github.com/kubecost/cost-analyzer-helm-chart/pull/2466

## How does this PR impact users? (This is the kind of thing that goes in release notes!)

Update dashboard for request-sizing to use max usage for cpu and memory

## How was this PR tested?
Tested against: 
- bundled prometheus (typical install)
- thanos

## Have you made an update to documentation?
NA

PR #2491, opened 2023-08-10
Grafana request sizing update

JesseGoodier:
## What does this PR change?

minor edits to https://github.com/kubecost/cost-analyzer-helm-chart/pull/2475:
1. fix queries that should have been using sum, not avg or max.
2. change default range to 2d to match Kubecost request-sizing UI default of 2d

- No

Update the request-sizing target grafana dashboard to use max usage for cpu and memory to match defaults

## How was this PR tested?

nightly and qa environments 

NA


kc-adawson (approved):
Looks good!

PR #2493, opened 2023-08-11
[Hold for 1.107] create global security contexts for nonRoot

JesseGoodier:
## What does this PR change?

Change global setting to run all pods as nonRoot and adds ability to make global changes to securityContexts.

Exceptions:
1. networkCosts requires root, is optional and not enabled by default
2. queryService initContainer requires root, is optional and not enabled by default
3. NodeExporter requires hostNetwork, is optional and is enabled by default
4. KSM should be disabled

New defaults (exceptions above):
```yaml
global:
  securityContext:
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
    fsGroup: 65534
    runAsGroup: 65534
    runAsUser: 65534
  containerSecurityContext:
    readOnlyRootFilesystem: true
    capabilities:
      drop:
      - ALL
```

no

change global setting to run all pods as nonRoot and adds ability to make global changes to securityContexts:

https://kubecost.atlassian.net/browse/CORE-274

## How was this PR tested?

test cluster with QSR and federator enabled

will be in values, but could use an official doc.

### DIFF
--- a/cost-analyzer/charts/prometheus/templates/server-deployment.yaml
+++ b/cost-analyzer/charts/prometheus/templates/server-deployment.yaml
@@ -59,10 +59,15 @@ spec:
             - --volume-dir={{ . }}
           {{- end }}
           resources:
-{{ toYaml .Values.configmapReload.prometheus.resources | indent 12 }}
-          {{- with .Values.configmapReload.prometheus.containerSecurityContext }}
+          {{- toYaml .Values.configmapReload.prometheus.resources | nindent 12 }}
           securityContext:
-            {{- toYaml . | nindent 12 }}
+          {{- if .Values.global.containerSecurityContext }}
+            {{- toYaml .Values.global.containerSecurityContext | nindent 12 }}
+          {{- else if .Values.global.containerSecurityContext }}
+            {{- toYaml .Values.global.containerSecurityContext | nindent 12 }}
+          {{- else }}
+          securityContext:
+            {{- toYaml .Values.configmapReload.prometheus.containerSecurityContext | nindent 12 }}
           {{- end }}
           volumeMounts:
             {{- if .Values.selfsignedCertConfigMapName }}
@@ -129,10 +134,12 @@ spec:
             failureThreshold: {{ .Values.server.livenessProbeFailureThreshold }}
             successThreshold: {{ .Values.server.livenessProbeSuccessThreshold }}
           resources:
-{{ toYaml .Values.server.resources | indent 12 }}
-          {{- with .Values.server.containerSecurityContext }}
+          {{- toYaml .Values.server.resources | nindent 12 }}
           securityContext:
-            {{- toYaml . | nindent 12 }}
+          {{- if .Values.global.containerSecurityContext }}
+            {{- toYaml .Values.global.containerSecurityContext | nindent 12 }}
+          {{- else }}
+            {{- toYaml .Values.server.prometheus.containerSecurityContext | nindent 12 }}
           {{- end }}
           volumeMounts:
             - name: config-volume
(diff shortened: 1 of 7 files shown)


chipzoller:
Since this PR gets almost all the way there, we may want to consider adding the additional two fields necessary to fully meet the [Pod Security Standards restricted profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted). These standards are increasingly important for users who require policy enforcement across the cluster, whether that be with Kubernetes' built-in [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) or dynamic admission controllers. The two missing fields, both under the `containers[]` array, are `privileged` and `allowPrivilegeEscalation`. Both should be set to a value of `false` for every container in the Pod.

jessegoodier:
> Since this PR gets almost all the way there, we may want to consider adding the additional two fields necessary to fully meet the [Pod Security Standards restricted profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted). These standards are increasingly important for users who require policy enforcement across the cluster, whether that be with Kubernetes' built-in [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) or dynamic admission controllers. The two missing fields, both under the `containers[]` array, are `privileged` and `allowPrivilegeEscalation`. Both should be set to a value of `false` for every container in the Pod.

Awesome!
can you add where appropriate? 

We aren't going to merge this until 107, so we have a couple weeks.

cost-analyzer/values.yaml
```diff
@@ -217,8 +217,19 @@ global:
...
-    # readOnlyRootFilesystem: true
+  securityContext:
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
+    fsGroup: 65534
+    runAsGroup: 65534
+    runAsUser: 65534
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
```
chipzoller, line 228: After looking at this more carefully, you've already got `allowPrivilegeEscalation`. It'd probably be easiest for you to just add a second line `privileged: false` beneath it.

chipzoller (changes requested):
Rather than me pushing one line, this looks like all that's needed for the cost-analyzer values.

cost-analyzer/values.yaml
jessegoodier: for sure, I was thinking it was more complex. thanks!

chipzoller (approved):

chipzoller:
This should fully succeed after PR #2512 is merged.

cost-analyzer/charts/prometheus/templates/server-deployment.yaml
```diff
@@ -59,10 +59,15 @@ spec:
...
           {{- end }}
           resources:
-{{ toYaml .Values.configmapReload.prometheus.resources | indent 12 }}
-          {{- with .Values.configmapReload.prometheus.containerSecurityContext }}
+          {{- toYaml .Values.configmapReload.prometheus.resources | nindent 12 }}
           securityContext:
-            {{- toYaml . | nindent 12 }}
+          {{- if .Values.global.containerSecurityContext }}
+            {{- toYaml .Values.global.containerSecurityContext | nindent 12 }}
+          {{- else if .Values.global.containerSecurityContext }}
```
thomasvn, line 66: Note, the `if` and `else if ` statements here look to be the same

cost-analyzer/values.yaml
```diff
@@ -217,8 +217,20 @@ global:
...
 
-  containerSecuritycontext: {}
-    # readOnlyRootFilesystem: true
+  securityContext:
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
+    fsGroup: 65534
+    runAsGroup: 65534
+    runAsUser: 65534
```
thomasvn, line 226: Any reason we've chosen `65534` here? Additionally, have we confirmed that all the containers in our chart can successfully run using a different user, with a `readonly` filesystem, and with all its capabilities dropped?

jessegoodier: yes, I did. 
the user works well for openshift and non-openshift tests I ran.

cost-analyzer/charts/prometheus/templates/server-deployment.yaml
jessegoodier: thanks for finding this!

cost-analyzer/values.yaml
jessegoodier: also, I stole the idea from: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L2286

thomasvn: Ah! According to stackexchange the significance of `65534` is that Linux reserves this for the `nobody` user/group. https://unix.stackexchange.com/questions/720557/what-is-the-id-of-nobody-user-and-nogroup-group

PR #2528, opened 2023-08-22
Allow disabling the kubecost frontend

abaguas:
A frontend is not necessary in remote Kubecosts, so disabling allows us to cut some resources. In our setup we have a central Kubecost receiving data from all the remote Kubecost instances.

## What does this PR change?
Add an additional `kubecostFrontend.enabled` value that allows disabling the Kubecost frontend.

No

No, the change is fully backwards compatible. By default the component is enabled.

None

Rendering the helm chart with the default `values.yaml` should return the same result before and after this PR.
Disabling `kubecostFrontend.enabled` must allow the Kubecost instance to work seamlessly but with one fewer container.

## How was this PR tested?
Disabled the frontend and applied it in Kubernetes.

I didn't find a relevant place where the documentation needs to be updated.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -76,25 +76,27 @@ spec:
       restartPolicy: Always
       serviceAccountName: {{ template "cost-analyzer.serviceAccountName" . }}
       volumes:
-      {{- if .Values.global.gcpstore.enabled }}
+        {{- if .Values.global.gcpstore.enabled }}
         - name: ubbagent-config
           configMap:
             name: ubbagent-config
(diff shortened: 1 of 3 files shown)


jessegoodier:
This looks good! 
Will test it out and likely get it into 1.107 in our September release.

jessegoodier:
testing:
```
helm install no-fe --namespace no-fe --create-namespace ./cost-analyzer
```
all pods/containers come up as expected
```
helm upgrade no-fe --namespace no-fe --create-namespace ./cost-analyzer --set kubecostFrontend.enabled=false
```
removes frontend container

JesseGoodier (approved):

abaguas:
Thanks!

PR #2536, opened 2023-08-24
add ingress for model access

leodag:
## What does this PR change?

This adds an optional kubecostModel.ingress option so that it can be configured separately from the frontend. This comes from a need in my company to expose the model through a basic auth ingress so that it can be accessed programatically, while the frontend is handled differently.

No.

Adds an additional option (defaulting off) to have an ingress that points directly to the model container.

I haven't created any issues for this, should I create one?

Should be a pretty simple change; one deployment with it enabled and disabled should be enough IMO.

## How was this PR tested?

Tested locally and backported with no changes o the version currently in use by my company (1.102.2), which is deployed in my company's test cluster.

Added `kubecostModel.ingress` to the README

This is my first contribution to the project, if there's any additional information or changes needed, do comment!

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -534,21 +534,6 @@ ingress:
   #    hosts:
   #      - cost-analyzer.local
 
-apiIngress:
-  enabled: false
-  # className: nginx
-  annotations:
-    # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: "true"
-  paths: ["/"] # There's no need to route specifically to the pods-- we have an nginx deployed that handles routing
-  pathType: ImplementationSpecific
-  hosts:
-    - cost-analyzer.local
-  tls: []
-  #  - secretName: cost-analyzer-tls
-  #    hosts:
-  #      - cost-analyzer.local
-
 nodeSelector: {}
 
 tolerations: []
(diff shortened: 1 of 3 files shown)


thomasvn:
Thanks for the PR @leodag! This seems reasonable to me! Could you test the PR using `helm template` and a variety of configurations for `.Values.kubecostModel.ingress`, so we know the template renders correctly?

cc @jessegoodier When you return, thoughts on the changes in this PR?

cost-analyzer/values.yaml
```diff
@@ -517,6 +534,21 @@ ingress:
   #    hosts:
   #      - cost-analyzer.local
 
+apiIngress:
```
thomasvn: Let's remove `apiIngress` since it's not used in the template anywhere?

leodag: That slipped in from how I was doing it before, I'll remove it, thanks!

leodag:
I've tested the template with `ingress: {}`, `ingress: { enabled: true }`, `ingress: { enabled: false }`, and some combinations of annotations and/or paths specified and it behaves as expected.

This is also basically just a copy of `cost-analyzer-ingress-template.yaml`, so it should behave the same. At first I tried to make the same template generate both, but it ended up being a bunch of `if`s in the middle of the template, and I thought it'd be better to have them separate.

This made me look at cost-analyzer-ingress template again and I noticed I was missing a small part of it (handling labels), I'll add that too.

JesseGoodier (approved):
Works for me. Thanks!

Issue #2409, opened 2023-07-03
Secret exposure through environment variables


than-pet:
**Describe the bug**  
1. `cost-analyzer` deployment has the following "sensitive" environmental variables defined as plain text:
  * `REMOTE_WRITE_PASSWORD`
  * `CLOUD_PROVIDER_API_KEY`

------

2. In addition, `REMOTE_WRITE_PASSWORD` environmental variable is always set even if the `remoteWrite.postgres.enabled` value is `false`.

**Expected behavior**  
1. These kind of values must be handled as kubernetes secrets in the generated charts. This approach is already implemented in the deployment yaml:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L379-L383
<img width="722" alt="image" src="https://github.com/kubecost/cost-analyzer-helm-chart/assets/108522410/6138a966-2447-43c6-bdc7-4994a9b1625a">

------

2. `REMOTE_WRITE_PASSWORD` shouldn't be set as env variable when the `remoteWrite.postgres` is disabled.
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L648-L649
<img width="592" alt="image" src="https://github.com/kubecost/cost-analyzer-helm-chart/assets/108522410/3d8c4c8b-4bda-45a9-8223-49938d854a62">

kwombach12:
@than-pet Thanks for logging this. @thomasvn when youre back, can you take a look at this?

thomasvn:
@than-pet I agree that `REMOTE_WRITE_PASSWORD` should only be set when `.Values.remoteWrite.postgres.enabled=true`. Therefore [this code block](https://github.com/kubecost/cost-analyzer-helm-chart/blob/53836363707d815c3846f9847cdc98be4ed70cc2/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L648-L652) should instead look something like the following:

```yaml
            {{- if .Values.remoteWrite.postgres.enabled }}
            - name: REMOTE_WRITE_ENABLED
              value: "true"
            - name: REMOTE_WRITE_PASSWORD
              value: {{ .Values.remoteWrite.postgres.auth.password }}
            {{- end }}
```

Are you interested in authoring the PR?

bt-macole:
[...]

PR #2578, opened 2023-09-11
fix(cost-analyzer): Set remoreWrite env variables only when enabled

than-pet:
## What does this PR change?
`REMOTE_WRITE_PASSWORD` environmental variable will be set only if `remoteWrite.postgres` is enabled.

There will be no impact since setting `REMOTE_WRITE_PASSWORD` is not really used if `remoteWrite.postgres` is 
not enabled.

https://github.com/kubecost/cost-analyzer-helm-chart/issues/2409

## How was this PR tested?
Only local rendering was tested.


JesseGoodier (approved):
Thank you!

PR #1914, opened 2023-01-23
add EBS-CSI warnings

AjayTripathy:
## What does this PR change?
Add warnings to notes on 

## Does this PR rely on any other PRs?
No

## How does this PR impact users? (This is the kind of thing that goes in release notes!)
Helps users debug EKS PV issues

## How was this PR tested?
Produces the message on an EKS cluster. 
```
atripathy@Ajays-MBP cost-analyzer-helm-chart % helm install kubecost ./cost-analyzer -n ajaytest
NAME: kubecost
LAST DEPLOYED: Sun Jan 22 16:13:03 2023
NAMESPACE: ajaytest
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
--------------------------------------------------Kubecost has been successfully installed.

WARNING: ON EKS v1.23+ INSTALLATION OF EBS-CSI DRIVER IS REQUIRED TO MANAGE PERSISTENT VOLUMES. LEARN MORE HERE: https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi.html

We detected a EBS-CSI Driver, but you should check that Kubecost's PersistentVolume has been installed correctly

Please allow 5-10 minutes for Kubecost to gather metrics.

If you have configured cloud-integrations, it can take up to 48 hours for cost reconciliation to occur.

When using Durable storage (Enterprise Edition), please allow up to 4 hours for data to be collected and the UI to be healthy.

When pods are Ready, you can enable port-forwarding with the following command:

    kubectl port-forward --namespace ajaytest deployment/kubecost-cost-analyzer 9090

Next, navigate to http://localhost:9090 in a web browser.

Having installation issues? View our Troubleshooting Guide at http://docs.kubecost.com/troubleshoot-install
```

## Have you made an update to documentation?
This is documentation, more or less.

PR #2586, opened 2023-09-19
remove unused check for Nodes

chipzoller:

## What does this PR change?

Removes what appears to be an unused check for Nodes from the NOTES.txt template.

No

It appears to have no effect.

N/A

Nodes won't be listed under the current user context after the chart is installed which appears to have no effect given this check isn't used.

## How was this PR tested?

Deployed chart.

N/A

### DIFF
--- a/cost-analyzer/templates/NOTES.txt
+++ b/cost-analyzer/templates/NOTES.txt
@@ -1,7 +1,6 @@
 
 
 --------------------------------------------------
-{{- $node := (lookup "v1" "Node" "" "") }}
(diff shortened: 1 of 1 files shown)


thomasvn:
It seems this variable was first introduced in PR https://github.com/kubecost/cost-analyzer-helm-chart/pull/1914, but the variable is not used. I think removal is fine!

@chipzoller Just double checking that you've successfully deployed this modified Helm chart to an EKS cluster?

chipzoller:
I have not because I could find no reference to this check being used.

JesseGoodier (approved):

PR #2599, opened 2023-09-22
add /savings/abandonedWorkloads proxy route for front end in waterfowl 

avrodrigues5:
## What does this PR change?
Adds the end point proxy route for waterfowl in front end config map

[1785](https://github.com/kubecost/kubecost-cost-model/pull/1785)

Get the end point details from waterfowl DB

-

## How was this PR tested?
-

-

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -480,6 +480,14 @@ data:
             proxy_set_header  X-Real-IP  $remote_addr;
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
+        location = /model/savings/abandonedWorkloads {
+            proxy_read_timeout          300;
+            proxy_pass http://waterfowl/savings/abandonedWorkloads;
+            proxy_redirect off;
+            proxy_set_header Connection "";
+            proxy_set_header  X-Real-IP  $remote_addr;
+            proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
+        }
     {{- end }}
     {{- if .Values.waterfowl.cloudCost.enabled }}
         location = /model/cloudCost/status {


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -480,6 +480,14 @@ data:
             proxy_set_header  X-Real-IP  $remote_addr;
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
+        location = /model/savings/abandonedWorkloads {
+            proxy_read_timeout          300;
+            proxy_pass http://waterfowl/savings/abandonedWorkloads;
```
dwbrown2, line 485: Know this is a broader question but should we be exposing `waterfowl` at all in the helm chart? Did we consider any other options, e.g. v2?

chipzoller: Using project code names typically come with a very long string attached. Unless we want to carry this forward in some capacity for potentially years, agree that we might want to come up with something that's more polished.

avrodrigues5: I acknowledge/appreciate your concerns however I'm going to defer this to question to @ameijer / @michaelmdresser.

The scope of this PR, is to setup a NGINX route when /savings/abandonedWorkloads is invoked to new end point logic that's off 9004 port. The `waterfowl` upstream server is prior setup and name change and others can be done in separate PR if they.

michaelmdresser (approved):

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
AjayTripathy: agree, let's not block/blow up the scope of this PR based on this.

michaelmdresser: @dwbrown2 @chipzoller Is there a better forum to continue this discussion? We can also stay here. I've approved this PR because @avrodrigues5 is correct that this discussion has a broader scope and shouldn't block this PR.

---

My thoughts:
Names are indeed sticky, as Chip pointed out. I chose Waterfowl as the name of the new software component (that includes: part of the codebase, container name, StatefulSet/Pod name, and associated Helm value names) for the following reasons:
- I didn't want to call it "DuckDB" because that might lead users down the unsupported and unstable path of trying to grab our DB files directly.
- I wanted to at least maintain a _connection_ to DuckDB so that our internal teams don't get too confused if someone interchangeably uses "DuckDB" and "\<name of software component\>". Ducks are waterfowl, hence a connection maintained.

What I don't want is to have multiple names for the same thing. I don't want to have references to "name A" in our code base and in internal documentation but then call it "name B" in everything public facing. That would be hard to maintain and error-prone.

> Did we consider any other options, e.g. v2?

In my opinion, Waterfowl alone is not "Kubecost v2". Kubecost v2 is whatever set of configuration, features, and documentation we decide it is -- that might be a new version of the Helm chart with a new set of default values like "Fed ETL + Waterfowl by default".

> Using project code names typically come with a very long string attached. Unless we want to carry this forward in some capacity for potentially years, agree that we might want to come up with something that's more polished.

Do you feel like "Waterfowl" is an unpolished name? It is almost (but not quite) too late to rename everything. I'd appreciate some suggestions of a better name; we aren't calling this "skunkworks" or "rideshare" any more because we felt those were code names unsuited for long-term naming.

PR #2600, opened 2023-09-22
skip online validation helm value

nickcurie:
## What does this PR change?
Adds helm value to skip online OIDC token validation

* https://github.com/kubecost/kubecost-cost-model/pull/1786

Users can pass in an environment variable that will allow OIDC tokens to skip online validation steps. These tokens will still be verified for authenticity.

* https://kubecost.atlassian.net/browse/SELFHOST-304

## How was this PR tested?
Tested on `aws-test`, confirmed that the env vars go through and affect the expected functions in KCM.

https://github.com/kubecost/docs/pull/740

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
@@ -19,7 +19,7 @@ data:
         "loginRedirectURL" : "{{ .Values.oidc.loginRedirectURL }}",
         "discoveryURL" : "{{ .Values.oidc.discoveryURL }}",
         "hostedDomain" : "{{ .Values.oidc.hostedDomain }}",
-        "skipOnlineTokenValidation" : "{{ .Values.oidc.skipOnlineTokenValidation }}",
+        "skipOnlineTokenValidation" : "{{ .Values.oidc.skipOnlineTokenValidation | default "false" }}",
         "rbac" : {
           "enabled" : {{ .Values.oidc.rbac.enabled }},
           "groups" : [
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
```diff
@@ -19,6 +19,7 @@ data:
         "loginRedirectURL" : "{{ .Values.oidc.loginRedirectURL }}",
         "discoveryURL" : "{{ .Values.oidc.discoveryURL }}",
         "hostedDomain" : "{{ .Values.oidc.hostedDomain }}",
+        "skipOnlineTokenValidation" : "{{ .Values.oidc.skipOnlineTokenValidation | default "false" }}",
```
kaelanspatel, line 22: This isn't technically needed, but it can stay.

cost-analyzer/values.yaml
```diff
@@ -289,6 +289,7 @@ oidc:
   authURL: "https://my.auth.server/authorize" # endpoint for login to auth server
   loginRedirectURL: "http://my.kubecost.url/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
   discoveryURL: "https://my.auth.server/.well-known/openid-configuration" # url for OIDC endpoint discovery
+  skipOnlineTokenValidation: false # allows oidc to validate the token offline
```
kaelanspatel: Could we have a clearer comment here?

Something like: "if true, will skip accessing OIDC introspection endpoint for online token verification, and instead try to locally validate JWT claims`.

kaelanspatel (commented):
One thing about the HC param comment, otherwise looks good.

kaelanspatel (approved):
Thanks, lgtm

PR #2601, opened 2023-09-25
Add nodeSelector to awsstore-deployment-template.yaml

george-paw:
## What does this PR change?

Add nodeSelector to awsstore-deployment-template.yaml

No.

Currently aweStore does not use nodeSelector. So you will run into issues where where you want to run awsStore on certain labels (e.g.`"kubernetes.io/arch"=amd64`) but you can't. 

Enable awsstore to use the same nodeSelector properties as the rest of the deployment

None. Test by running a deployment and describe the pod or deployment.

## How was this PR tested?

By doing a deployment


chipzoller:
Thanks for this. Does it address an issue?

cost-analyzer/templates/awsstore-deployment-template.yaml
```diff
@@ -29,6 +29,10 @@ spec:
       {{- if .Values.awsstore.priorityClassName }}
       priorityClassName: "{{ .Values.awsstore.priorityClassName }}"
       {{- end }}
+      {{- with .Values.nodeSelector }}
```
jessegoodier, line 32: I'm thinking we should give the option to use a specific node selector for this. 
`with .Values.awsstore.nodeSelector`
No?

george-paw:
> Thanks for this. Does it address an issue?

Currently aweStore does not use nodeSelector. So you will run into issues where where you want to run awsStore on certain labels (e.g."kubernetes.io/arch"=amd64) but you can't.

Enable awsstore to use the same nodeSelector properties as the rest of the deployment

cost-analyzer/templates/awsstore-deployment-template.yaml
george-paw: I have no strong opinion on this, either is fine to me

chipzoller:
Are comments resolved and is this ready for final review?

george-paw:
@chipzoller I think so

cost-analyzer/templates/awsstore-deployment-template.yaml
chipzoller: I agree with Jesse here. Because this is an esoteric component, I would rather pull this value from `awsstore.nodeSelector` than from the top-level key `nodeSelector`.

chipzoller:
In the interest of cleaning up these PRs, I will close this since I cannot push to your fork and re-open it myself with the changes.

PR #2609, opened 2023-09-27
Federator pod volume var config

JesseGoodier:
## What does this PR change?
adds PVC to federator pod

No

Create a PV for federator pod to reduce the likelihood of needing to pull 90d of data from s3 if the pod is rescheduled on a different node.

None that I can think of. This change follows the same deployment standard as cost-analyzer pod.

## How was this PR tested?
Helm installs with and without `--set persistentVolume.enabled=false`

NA, follows same model as cost-analyzer


chipzoller:
Anything need to be added/clarified in the values file? Perhaps comments?

PR #2611, opened 2023-09-30
Add /reports proxy to Waterfowl

nik-kc:
## What does this PR change?
* Adds a proxy for the various `/report` endpoints in Waterfowl.

* [Yup yup](https://github.com/kubecost/kubecost-cost-model/pull/1798).

* Scheduled reports will now query DuckDB instead of ETL.

* Addresses [this ticket](https://kubecost.atlassian.net/browse/SELFHOST-312).

* Any risks associated with this would relate to the underlying queries done to gather the report data. So, if we trust the underlying DuckDB queries, we can trust these reports.

## How was this PR tested?
* See the accompanying PR for testing notes.

* I have not.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -436,6 +436,14 @@ data:
             proxy_set_header  X-Real-IP  $remote_addr;
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
+        location = /model/reports {
+            proxy_read_timeout          300;
+            proxy_pass http://waterfowl/reports;
+            proxy_redirect off;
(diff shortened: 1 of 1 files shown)


michaelmdresser (approved):

michaelmdresser:
@chipzoller this PR wasn't supposed to be merged until its dependency PR https://github.com/kubecost/kubecost-cost-model/pull/1798 closes. I'm used to approving individual PRs and relying on the author to merge at the right time to handle situations like this. Are you following a different policy?

I don't think we have to revert, as the KCM PR will definitely merge eventually.

chipzoller:
Ok, I see that now. I'm glad to leave things alone from now forward if they go into any of the waterfowl branches, but in my opinion this process is a somewhat brittle one and relying upon authors to self-merge after approval is given is very unusual. For PRs with dependencies, I think these either need to be draft or have titles which reflect those dependencies.

PR #2613, opened 2023-10-02
feat: working oidc custom secret

### LINKED ISSUE #2612: [Feature] Support a Custom Secret for the OIDC Secret
### Problem Statement

In the current configuration the OIDC Client Secret has to be put into the values.yaml.

This shouldn't be required. Especially in a GitOps model where values.yaml is committed back to the repo. This exposes the secret improperly.

### Solution Description

A oidc.customSecret field should be supported to provide the k8s secret that should be loaded/used instead.  Then all secret sources can be used including:

- External Secrets
- Sealed Secrets
- SOPs
- Vault

### Alternatives

_No response_

### Additional Context

_No response_

knechtionscoding:
## What does this PR change?

Enables the use of a custom secret. There is an alternative implementation we could take which would use the oidc.secretName for both the custom and non-custom secret. Currently they are two distinct values.

Fixes: #2612

No

Allows the use of a variety of secret sources for the OIDC ClientSecret, including but not limited too:
- ExternalSecrets
- Vault
- SealedSecrets
- SOPs

The following iterations test it (though I recommend actually writing tests for the chart using helm-unittest).

Default OIDC behavior: 

```yaml
oidc:
  enabled: true
  clientID: "" # application/client client_id parameter obtained from provider, used to make requests to server
  clientSecret: "" # application/client client_secret parameter obtained from provider, used to make requests to server
  secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
  customSecret:
    enabled: false
    name: "" # name of the secret containing the client secret
```

Enable Custom Secret

```yaml
oidc:
  enabled: true
  clientID: "" # application/client client_id parameter obtained from provider, used to make requests to server
  clientSecret: "" # application/client client_secret parameter obtained from provider, used to make requests to server
  secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
  customSecret:
    enabled: true
    name: "testing-oidc" # name of the secret containing the client secret
```

## How was this PR tested?

With the above yamls

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -286,6 +286,10 @@ oidc:
   clientID: "" # application/client client_id parameter obtained from provider, used to make requests to server
   clientSecret: "" # application/client client_secret parameter obtained from provider, used to make requests to server
   secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
+  customSecret:
+    enabled: false
+    name: "" # name of the secret containing the client secret
+
   authURL: "https://my.auth.server/authorize" # endpoint for login to auth server
   loginRedirectURL: "http://my.kubecost.url/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
   discoveryURL: "https://my.auth.server/.well-known/openid-configuration" # url for OIDC endpoint discovery
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -286,6 +286,9 @@ oidc:
   clientID: "" # application/client client_id parameter obtained from provider, used to make requests to server
   clientSecret: "" # application/client client_secret parameter obtained from provider, used to make requests to server
   secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
+  # For use to provide a custom OIDC Secret. Overrides the usage of oidc.clientSecret and oidc.secretName.
+  # Should contain the field directly.
+  # Can be created using raw k8s secrets, external secrets, sealed secrets, or any other method.
   existingCustomSecret:
     enabled: false
     name: "" # name of the secret containing the client secret
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -286,7 +286,7 @@ oidc:
   clientID: "" # application/client client_id parameter obtained from provider, used to make requests to server
   clientSecret: "" # application/client client_secret parameter obtained from provider, used to make requests to server
   secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
-  customSecret:
+  existingCustomSecret:
     enabled: false
     name: "" # name of the secret containing the client secret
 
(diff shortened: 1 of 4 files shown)


cost-analyzer/values.yaml
```diff
@@ -286,6 +286,10 @@ oidc:
   clientID: "" # application/client client_id parameter obtained from provider, used to make requests to server
   clientSecret: "" # application/client client_secret parameter obtained from provider, used to make requests to server
   secretName: "kubecost-oidc-secret" # k8s secret where clientsecret will be stored
+  customSecret:
```
thomasvn: Minor comment. Thoughts on renaming this to `clientSecretName`? This way there is a clearer relationship between `clientSecret` and `clientSecretName`.

It may also be good to include an additional comment:

```yaml
clientSecretName: # optional
  enabled: false # if true, this config takes precedence over clientSecret and specifies a k8s secret which holds the clientSecret parameter obtained from provider
  name: "" # name of the secret containing the client secret
```

thomasvn:
Thanks for the PR @knechtionscoding! I just made a minor comment. Curious for your thoughts on it!

cost-analyzer/values.yaml
knechtionscoding: I didnt call it that because I felt like it would be confusing with the secret name field.

I can swap it to existingClientSecret if you prefer.

I don't have name as the suffix because then it becomes: clientSecretName.name which is pretty self-explanatory and felt unnecessarily duplicative. 

If you want me to remove the enabled section and only have name and then select based on if that field has a value I can do that. This felt clearer for intent however.

thomasvn: Good points! I like `existingClientSecret`.

thomasvn:
The rest of this PR looks good! All references to `.Values.oidc.clientSecret` and `.Values.oidc.secretName` have been accounted for, and `.Values.oidc.customSecret` correctly supersedes those previous two values.

knechtionscoding:
@thomasvn let me know if there's anything else needed.

thomasvn:
@knechtionscoding PR looks great. Could we add more comments in the `values.yaml` file stating that if this option is enabled, the Helm chart will look for an existing k8s secret which contains your client_secret parameter obtained from provider?

knechtionscoding:
@thomasvn done

thomasvn:
@knechtionscoding Great, thanks! I've additionally tested by running `helm template cost-analyzer -f values.yaml` against the following configs. It renders as expected.

```yaml
# Expect to see a k8s secret being generated and references to "kubecost-oidc-secret"
oidc:
  enabled: true
  clientSecret: "TEST"
  secretName: "kubecost-oidc-secret"
  existingCustomSecret:
    enabled: false
    name: "my-kubecost-oidc-secret"
```

```yaml
# Expect to see references to "my-kubecost-oidc-secret"
oidc:
  enabled: true
  clientSecret: "TEST"
  secretName: "kubecost-oidc-secret"
  existingCustomSecret:
    enabled: true
    name: "my-kubecost-oidc-secret"
```

thomasvn (approved):

Kavinraja-Ganesan:
@thomasvn @knechtionscoding 

What is the expected secretKey for these custom secret? Will the secret looks like `clientSecret=<redacted-secret>` ?

thomasvn:
Yes! When using the custom secret from this PR, the key should be `clientSecret`. It should look similar to this: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/kubecost-oidc-secret-template.yaml

PR #2620, opened 2023-10-03
Update Network Costs to 0.17.0

mbolt35:
## What does this PR change?
* This is the first production release of the Rust network-costs implementation, which has feature parity with the Go version, utilizes 50% less CPU, a much more efficient memory model, and can process connections about 2x-3x faster.


michaelmdresser (approved):
Let's GO

PR #2643, opened 2023-10-13
Fix incorrect trim-right whitespace in NOTES.txt

reidmv:
## What does this PR change?

This commit removes a trim-right-whitespace marker which was concatenating the NOTES.txt's header bar ("------") with its first line, "Kubecost has been successfully installed".

The trim-right-whitespace marker caused the template to print:

    ------Kubecost ...

While the desired format (I assume), and the behavior with this commit, is instead:

    ------
    Kubecost ...


chipzoller (approved):
Thank you!

PR #2650, opened 2023-10-16
implement new etl-utils pod

ameijer:
## What does this PR change?
adds new etl utils pod

none

Creates a pod that converts legacy Thanos ETL data to federated ETL data (alpha), supported by request only

users can write extra to s3 thanos buckets

## How was this PR tested?
tested in dev cluster


AjayTripathy:
Just chekcing-- this pod only needs to run duing a migration step right?

ameijer:
@AjayTripathy this actually is designed to run continuously, since the use case @jessegoodier describes is for users with many thanos installs that need to be migrated over to fed ETL slowly - and others that may never migrate at all. Once a certain cluster stops using thanos, this tool will stop placing new bingen files in its respective fed ETL directory... but any other clusters still using thanos will have their data processed by this. Eventually, there will be no more thanos data and this can be shut down. 

Full disclosure - I have not simulated a gradual migration like I describe, but the design of this hopefully puts that in sight.

AjayTripathy:
@ameijer but, to clarify, the user can turn off this pod once the thanos -> aggregator migration is complete?

ameijer:
correct - they can flip the 'enabled' switch to false to remove this pod and stop the conversion pipeline

chipzoller:
Since this is a one-time migration activity, I'm wondering if there's maybe a better way/place to do this rather than the Helm chart.

ameijer:
@chipzoller it would only be a migration activity for certain customers. this would be a permanent part of the app for customers using thanos and waterfowl

chipzoller:
Sorry, can you explain a bit more so I understand?

jessegoodier:
This utility could allow for many use cases. The most obvious being allowing users to continually run thanos and not have to update agents to etl-fed. There are benefits to Thanos, notably grafana dashboards.

In this use case, it will always run and is a part of our deployment mode.

It also can run as a one-time conversion utility. Though I can see other use cases for repair/diagnostics/etc.

jessegoodier:
This utility could allow for many use cases. The most obvious being allowing for users to continually run thanos and not have to update agents.
In which case, it will always run and is a part of our deployment mode.

It also can run as a one-time conversion utility. Though I can see other use cases for repair/diagnostics/etc.

JesseGoodier (approved):
nice work!

jessegoodier:
Feel free to merge and CP when you know the image is ready.

cost-analyzer/values.yaml
```diff
@@ -533,7 +533,15 @@ kubecostModel:
     #    hosts:
     #      - cost-analyzer-model.local
 
-
+etlUtils:
```
chipzoller, line 537: Please add some comment here as to what this does and why anyone would use it.

jessegoodier: good call, added.

PR #2651, opened 2023-10-16
option to hide diagnostics

JesseGoodier:
## What does this PR change?
add endpoint to disable certain diagnostics that are not accurate if the kubecost primary isn't monitoring itself.

No

add endpoint to disable certain diagnostics that are not accurate if the kubecost primary isn't monitoring itself.

KC-123

none

## How was this PR tested?
both with null kubecostFrontend.disableDiagnostics null and true

not needed.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -344,7 +344,7 @@ data:
         }
 
         # to get memory profile from query service need to prefix all request by queryservice/
-        # for example if you want heap dump from query service end point should be 
+        # for example if you want heap dump from query service end point should be
         # /model/queryservice/debug/pprof/heap to get queryservice heap dumps
         location ~ /model/queryservice/(.*)$ {
             proxy_connect_timeout       600;
@@ -580,6 +580,15 @@ data:
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
     {{- end }}
+        location = /model/hidediagnostics {
+            default_type text/html;
+            {{- if .Values.kubecostFrontend.hideDiagnostics }}
+            return 200 'true';
+            {{- else }}
+            return 200 'false';
+            {{- end }}
+        }
+
     {{- if .Values.kubecostAggregator.cloudCost.enabled }}
         location = /model/cloudCost/status {
             proxy_read_timeout          300;
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -580,7 +580,7 @@ data:
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
     {{- end }}
-        location = /model/hidediagnostics {
+        location = /model/hideDiagnostics {
             default_type text/html;
             {{- if .Values.kubecostFrontend.hideDiagnostics }}
             return 200 'true';
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -358,6 +358,7 @@ kubecostFrontend:
   #   proxy_buffers   4 512k;
   #   proxy_buffer_size   256k;
   #   large_client_header_buffers 4 64k;
+  # hideDiagnostics: false # used if the primary is not monitored. Supported in limited environments.
 
 #  api:
 #    fqdn: kubecost-api.kubecost.svc.cluster.local:9001


teevans:
@cliffcolvin / @nikovacevic / @michaelmdresser what are your thoughts on exposing via nginx these types of configuration changes? We don't need to do anything in the back-end, and allows the FE team to make changes via helm values.

teevans:
This would let us get rid of the /helmValues url.

chipzoller (commented):
This value must be defined in the values file.

michaelmdresser:
> what are your thoughts on exposing via nginx these types of configuration changes? We don't need to do anything in the back-end, and allows the FE team to make changes via helm values.

I'm not sure if I understand the question. Do you mean exposing the ability to edit the NGINX config file via Helm values?

teevans:
No, it would give us the ability to access configuration settings without having to create environment variables or expose endpoints in the cost-model pod.

michaelmdresser:
I'm sorry, I'm still confused. What is "it"? I think I'm struggling without an example.

chipzoller:
@michaelmdresser, it looks like the proposal here is to use Helm values to control a specific endpoint on the front end via the nginx config file rather using env vars or other mechanisms to control the back end. If `kubecostFrontend.hideDiagnostics` is set to `true` then the nginx config file will be configured with an endpoint at `/model/hideDiagnostics` which responds when hit. If set to `false`, that endpoint will not return a response thus not hiding the diagnostic page. @jessegoodier, please keep me honest 😄 

My question is day 2 changes. If this option is configured with a running Kubecost installation, the change will be propagated to the Pod mounting the ConfigMap, but will nginx live reload those changes or is it only read at start up?

teevans:
@chipzoller's explanation is correct.

nikovacevic:
I'm also not super clear on the idea. From what I can tell, the proposal is to include helm values under `.Values.kubecostFrontend` that disable certain endpoints/features from nginx? So is the idea to be able to use "frontend" helm values to circumvent the BE altogether and return different statuses from endpoints?

I think that idea is somewhat suspect, but do what you gotta do.

nikovacevic:
Looks like my previous message came in shortly after Thomas's and Chip's, but addresses the same point.

Let's just be aware of adding complexity (i.e. by fragmenting the "full story" of configuration between FE and BE) for the sake, presumably, of saving a few minutes of BE dev time. Not for or against. Let's just be prudent. If this particular thing is something BE doesn't need to know about, then there's no issue IMO. But let's not let it become a slippery slope where there's now an incentive to put BE logic in nginx, instead, in order to save teams from having to collaborate a little.

Does that make sense?

jessegoodier:
> Looks like my previous message came in shortly after Thomas's and Chip's, but addresses the same point.
> 
> Let's just be aware of adding complexity (i.e. by fragmenting the "full story" of configuration between FE and BE) for the sake, presumably, of saving a few minutes of BE dev time. Not for or against. Let's just be prudent. If this particular thing is something BE doesn't need to know about, then there's no issue IMO. But let's not let it become a slippery slope where there's now an incentive to put BE logic in nginx, instead, in order to save teams from having to collaborate a little.
> 
> Does that make sense?

100% agreed.

In this case, the concern is simply that the UI is checking endpoints that we know are going to fail because the primary is not running prometheus/monitoring itself (hosted).

There is a case to disable the primary from attempting to monitor itself, but this is a significant enhancement that doesn't impact the users and is largely being done anyway with aggregator. 

I have updated the PR details; sorry, this was done quickly not knowing how much attention it would get.

@chipzoller the change will always return a value. If the setting isn't explicitly true, it will return false. I do not think it needs to be published in values given that this is an internal flag that users would not interact with. But I'm happy to be corrected.

chipzoller:
> I do not think it needs to be published in values given that this is an internal flag that users would not interact with.

Is the intention that setting this value is only done at the recommendation of support? Seems like it _should_ be set any time there is a "primary" instance, is that correct?

I'm still curious about the answer to my earlier question:

> My question is day 2 changes. If this option is configured with a running Kubecost installation, the change will be propagated to the Pod mounting the ConfigMap, but will nginx live reload those changes or is it only read at start up?

jessegoodier:
This should only be set in situations where the primary itself would not be monitored, and this is not a common use case and would always involve support.

The nginx-conf is always updated during helm upgrades. The pod is always restarted during image updates- and most other helm value changes.
________________________________

chipzoller:
> The nginx-conf is always updated during helm upgrades. The pod is always restarted during image updates- and most other helm value changes.

Right, but I'm not talking about Helm upgrades or image updates, I'm talking about where a user has a running installation of Kubecost and they simply want to flip this switch. Since it's just a ConfigMap change, they could do this directly, but will nginx watch and live reload its config file if it changes?

What I usually see in other projects is there are no "hidden" Helm values; the `values.yaml` file is the authoritative "index" of all available values, even if they're commented out. My personal preference is to follow suit but I won't block this PR just because of that preference. I'll just also say we've got lots of "undocumented" values that really should be documented, and so adding to that list makes the 2.0 revamp a little more difficult.

jessegoodier:
> kubecostFrontend.hideDiagnostics

There's really no risk here, added to values.yaml

teevans (approved):

PR #2660, opened 2023-10-18
the aggregator should just use the cost analyzer's SA unless overridden

ameijer:
## What does this PR change?
has the aggregator use the cost analyzer's service account. it no longer ships with a default. Users can specify a different service account but they need to provide their own in that case 

None
 
should have no impact

low risk, aggregator is not GA

## How was this PR tested?
ensured helm template rendered and was validated with both default and override used


JesseGoodier (approved):
thanks!

PR #2666, opened 2023-10-18
add ability to override cc sa name

ameijer:
## What does this PR change?
add ability to default cloud costs to cost analyzer SA

none

makes setup easier. should not have any other impact

none

## How was this PR tested?
ensured passed helm template validate for both default and overridden cases


JesseGoodier (approved):
perfect.

PR #2667, opened 2023-10-18
setting to 50h to match etl retention time

JesseGoodier:
## What does this PR change?
update hosted agent default retention to 50h to match best practices for being at least etl hourly retention

No

Not needed

Confirmed that the settings have the intended impact:
```
    Args:
      --storage.tsdb.retention.time=50h
      --config.file=/etc/config/prometheus.yml
      --storage.tsdb.path=/data
      --web.console.libraries=/etc/prometheus/console_libraries
      --web.console.templates=/etc/prometheus/consoles
      --web.enable-lifecycle
      --query.max-concurrency=1
      --query.max-samples=1e+08
      --storage.tsdb.max-block-duration=2h
      --storage.tsdb.min-block-duration=2h
```

## How was this PR tested?
tested in several QA environments

NA

### DIFF
--- a/cost-analyzer/values-agent.yaml
+++ b/cost-analyzer/values-agent.yaml
@@ -72,10 +72,11 @@ prometheus:
           action: keep
           regex:  {{ template "cost-analyzer.networkCostsName" . }}
   server:
+    retention: 50h
     extraArgs:
       storage.tsdb.min-block-duration: 2h
       storage.tsdb.max-block-duration: 2h
-      storage.tsdb.retention: 10h
+      # storage.tsdb.retention: 10h
     securityContext:
       runAsNonRoot: true
       runAsUser: 1001
--- a/cost-analyzer/values-agent.yaml
+++ b/cost-analyzer/values-agent.yaml
@@ -76,7 +76,8 @@ prometheus:
     extraArgs:
       storage.tsdb.min-block-duration: 2h
       storage.tsdb.max-block-duration: 2h
-      # storage.tsdb.retention: 10h
+      storage.tsdb.retention.time: 50h
+      # storage.tsdb.retention.size: 1Gi
     securityContext:
       runAsNonRoot: true
       runAsUser: 1001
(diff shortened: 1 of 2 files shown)


chipzoller (commented):
1. Doesn't it make more sense for the templates to be adjusted so they're consistent with the other flags rather than fixing the values?
2. I see that `retentionSize` is an undocumented value. How about we add?

chipzoller:
Should `values-thanos.yaml` also be updated as well?

chipzoller:
Why not just delete [this block](https://github.com/kubecost/cost-analyzer-helm-chart/blob/5efcbdfff868d41a0970df93413f45c803fa1e28/cost-analyzer/charts/prometheus/templates/server-deployment.yaml#L98-L103) in the template (analogous block in the StatefulSet) since [this statement](https://github.com/kubecost/cost-analyzer-helm-chart/blob/5efcbdfff868d41a0970df93413f45c803fa1e28/cost-analyzer/charts/prometheus/templates/server-deployment.yaml#L115-L117) already exists. The values can then be simplified:

```yaml
prometheus:
  server:
    extraArgs:
      storage.tsdb.min-block-duration: 2h
      storage.tsdb.max-block-duration: 2h
      storage.tsdb.retention.time: 50h
      # storage.tsdb.retention.size: 1Gi
```

jessegoodier:
> Should `values-thanos.yaml` also be updated as well?

as long as it is > 50h it is good. I only want this change in our hosted-agent

jessegoodier:
Please see most recent commit. I do not want to make any major changes here.

chipzoller (commented):
What I suggested aren't major changes, they're pretty minor. Do you feel uncomfortable with them?

chipzoller:
@jessegoodier, check the changes. I just standardized on what was there rather than trying to improve it (right now).

jessegoodier:
Looks good. Feel free to merge.

chipzoller (approved):

PR #2674, opened 2023-10-19
Begin Helm testing

chipzoller:

## What does this PR change?

* Creates the first actual test as part of the Helm chart. In this test, basic health is determined which functions as an e2e test of all the components in the Kubecost stack (minus Grafana for the moment).
* Plumbs in Helm testing into CI.

No

Allows users to run Kubecost tests in their own environment should they wish, but otherwise is an internal change in CI.

None

## How was this PR tested?

Tested locally and in CI


JesseGoodier (approved):
great idea!

PR #2675, opened 2023-10-20
remove duplicate SA template

JesseGoodier:
remove dupe SA


ameijer (approved):

ameijer:
@jessegoodier are there roles/role bindings here too we can wipe out?

ameijer:
just checked - looks like only the SA is there

jessegoodier:
agree, this is the only extra template I see at the moment.

PR #2677, opened 2023-10-20
networkCosts service discovery

JesseGoodier:
## What does this PR change?
Allow users to run the network costs daemonset in any namespace without having to configure the target in Kubecost bundled prometheus.

1. add additional labels to network cost pods
2. adjust default scrapeConfig to find pods with these labels

No

Change networkCosts Prometheus service discovery to find the deamonset in any namespace

We are setting new default values:

```yaml
app.kubernetes.io/name: network-costs
```

IMO, we should add similar labels to all other pods too.

## How was this PR tested?
Tested in several different configs:
1. Default install with networkCost enabled
2. With serviceMonitor enabled in a BYO prometheus install
3. Upgrading an old kubecost install with new chart

TBD, I think we should add the service discovery how to in the current doc: https://docs.kubecost.com/install-and-configure/advanced-configuration/network-costs-configuration#prometheus

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -663,9 +663,12 @@ prometheus:
         - role: pod
       relabel_configs:
       # Scrape only the the targets matching the following metadata
-        - source_labels: [__meta_kubernetes_pod_label_app]
+        - source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_instance]
           action: keep
-          regex:  {{ template "cost-analyzer.networkCostsName" . }}
+          regex:  kubecost
+        - source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
+          action: keep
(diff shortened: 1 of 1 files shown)


dwbrown2:
I love the idea of this!

AjayTripathy (approved):

avrodrigues5:
@jessegoodier does this also address problem of network costs enabled in a different namespace in a shared cluster and you want network costs show up in your namespace too? 

does it mean the network cost running in a namespace other than yours has the two labels pointed out it should work? is that correct?

jessegoodier:
> @jessegoodier does this also address problem of network costs enabled in a different namespace in a shared cluster and you want network costs show up in your namespace too?
> 
> does it mean the network cost running in a namespace other than yours has the two labels pointed out it should work? is that correct?

yes, this will absolutely help with that use case.

PR #2687, opened 2023-10-24
add systemProxy env vars

JesseGoodier:
## What does this PR change?
Add systemProxy env variables to amp/gmp/federator containers

no

Added systemProxy env variables to amp/gmp/federator containers

Worst case would be spacing, but I verified each block was aligned, though I didn't deploy each.

## How was this PR tested?
changing gmp and amp, federator on:

```yaml
global:
  gmp:
    enabled: true
    prometheusServerEndpoint: https://localhost:8085/xxxxx/
    remoteWriteService: https://test.com
    sigv4:
      region: us-east-2
systemProxy:
  enabled: true
  httpProxyUrl: http://test
  httpsProxyUrl: http://test1
  noProxy: "0,localhost"
```

NA

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -321,6 +321,7 @@ oidc:
     #       - "editor"
 
 # Adds an httpProxy as an environment variable. systemProxy.enabled must be `true`to have any effect.
+# this is used in environments that have firewall rules that prevent kubecost from accessing cloud provider resources
 # Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
 systemProxy:
   enabled: false
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -320,9 +320,11 @@ oidc:
     #     claimValues:
     #       - "editor"
 
-# Adds an httpProxy as an environment variable. systemProxy.enabled must be `true`to have any effect.
-# this is used in environments that have firewall rules that prevent kubecost from accessing cloud provider resources
-# Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
+## Adds the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables to all
+## containers. Typically used in environments that have firewall rules which
+## prevent kubecost from accessing cloud provider resources.
+## Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
+##
 systemProxy:
   enabled: false
   httpProxyUrl: ""
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -335,6 +335,21 @@ spec:
           args:
           - "--web.listen-address=:{{ .Values.global.gmp.gmpProxy.port }}"
           - "--query.project-id={{ .Values.global.gmp.gmpProxy.projectId }}"
+          {{- if .Values.systemProxy.enabled }}
+          env:
+          - name: HTTP_PROXY
+            value: {{ .Values.systemProxy.httpProxyUrl }}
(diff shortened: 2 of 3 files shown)


chipzoller (commented):
Why are the env vars being duplicated except with different cases?

jessegoodier:
> Why are the env vars being duplicated except with different cases?

Best I could find is because linux doesn't have a governing body.

thomasvn:
Interesting that `.Values.systemProxy` already exists, but isn't currently referenced anywhere in the helm chart. 

https://github.com/kubecost/cost-analyzer-helm-chart/blob/374399c2a771a901712fa489f13abaf894b2779a/cost-analyzer/values.yaml#L300-L306

A few requests for this PR:

* Could we further document in the `values.yaml` specific situations when a user needs to set this value?
* We are only setting this env var for containers created in the `cost-analyzer-deployment-template.yaml` and `federator-deployment-template.yaml`. Do we need to set these env vars for other containers? (e.g. Prometheus, KSM, metrics emitter, etc)

jessegoodier:
> Interesting that `.Values.systemProxy` already exists, but isn't currently referenced anywhere in the helm chart.
> 
> https://github.com/kubecost/cost-analyzer-helm-chart/blob/374399c2a771a901712fa489f13abaf894b2779a/cost-analyzer/values.yaml#L300-L306
> 
> A few requests for this PR:
> 
> * Could we further document in the `values.yaml` specific situations when a user needs to set this value?
> * We are only setting this env var for containers created in the `cost-analyzer-deployment-template.yaml` and `federator-deployment-template.yaml`. Do we need to set these env vars for other containers? (e.g. Prometheus, KSM, metrics emitter, etc)

Thank you for the review!
we do use this, currently here: https://github.com/kubecost/cost-analyzer-helm-chart/blob/v1.106/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L850

It is only needed for things that would exit the cluster, so what I have here. Perhaps thanos stuff, but that's outside of scope for what we need here.

I'll add a comment. Suggesting that this problem is a good use-case for a service mesh. :)

chipzoller:
> > Why are the env vars being duplicated except with different cases?
> 
> Best I could find is because linux doesn't have a governing body.

Most env vars standardize on upper case, these proxy vars as well, so I suggest dropping the lower case duplicates.

jessegoodier:
> > > Why are the env vars being duplicated except with different cases?
> > 
> > 
> > Best I could find is because linux doesn't have a governing body.
> 
> Most env vars standardize on upper case, these proxy vars as well, so I suggest dropping the lower case duplicates.

I agree that it looks dumb and someone should have figured this out 20 years ago.

These systemProxy variables with the duplicate case are not new to our helm chart, or linux. 

If you want to test the possible permutations, I'd be happy to review that change. 

Otherwise, I'm simply looking to fix the problem for the small handful of users that depend on this setting.

thomasvn:
@jessegoodier Tested this PR with the following values. Validated that it's adding the system proxy env vars to the `cost-model` in `cost-analyzer` and `federator`.

```yaml
systemProxy:
  enabled: true
  httpProxyUrl: "http://TEST"
  httpsProxyUrl: "https://TEST"
  noProxy: "TEST"
federatedETL:
  federator:
    enabled: true
```

```sh
helm template ./cost-analyzer -f values.yaml > test1.yaml
```

cost-analyzer/values.yaml
```diff
@@ -321,6 +321,7 @@ oidc:
     #       - "editor"
 
 # Adds an httpProxy as an environment variable. systemProxy.enabled must be `true`to have any effect.
+# this is used in environments that have firewall rules that prevent kubecost from accessing cloud provider resources
 # Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
 systemProxy:
```
thomasvn: Minor style change suggestion for the comment!

```suggestion
## Adds the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables to all
## containers. Typically used in environments that have firewall rules which
## prevent kubecost from accessing cloud provider resources.
## Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
##
systemProxy:
```

thomasvn (approved):

PR #2696, opened 2023-10-26
update savedReports and advancedReports in values.yml to reflect curr…

saweber:
…ent filter schema

## What does this PR change?
The current examples in values.yaml no longer reflect the current report schema. This is because the back end has been updated to use v2 filters, but our values.yaml examples were never updated to reflect the schema change.

We(kubecost back end team) already broke by this feature by shipping v1.106. Any users that upgrade to v1.106 or newer and use this feature will have to update these report filters, otherwise reports from YAML will not load.

I would have preferred to make this change backwards compatible, however, the back end report change was not made in a backwards compatible way to support this, and I am not aware of a helm function that would let us convert to JSON using different keys. Example: values.yaml converts a 'filters2' yaml block to a 'filters' yaml block. (I would love to be wrong here)

https://helm.sh/docs/chart_template_guide/function_list/

No

Provides examples on how to use filters with saved (allocation) reports and advanced reports

https://kubecost.atlassian.net/browse/SELFHOST-903

can't make things worse; full testing is using these examples to successfully deploy saved (allocation) reports and advanced reports containing filters and no errors

## How was this PR tested?
spun up a new installation with helm, added reports, saw the filters are appearing

n/a, documentation links to values.yaml

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -149,11 +149,10 @@ global:
         idle: "separate"
         rate: "cumulative"
         accumulate: false # daily resolution
-        filters:
-          - property: "cluster"
-            value: "cluster-one,cluster*" # supports wildcard filtering and multiple comma separated values
-          - property: "namespace"
-            value: "kubecost"
+        filters: # Ref: https://docs.kubecost.com/apis/apis-overview/filters-api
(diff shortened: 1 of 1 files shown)


ameijer (approved):

chipzoller:
cc @bstuder99 to ensure the links in comments remain correct after your latest PR.

brstuder:
LGTM, the filter parameters doc wasn't moved or renamed so this should still work

PR #2705, opened 2023-10-30
fix missing usage for thanos converter

ameijer:
fixes missing usage to use new helm config item


jessegoodier:
This is superseded by `etlUtils.thanosSourceBucketSecret` 
correct?

ameijer:
hi @jessegoodier , good catch, yes, this variable reference was incorrect. I have updated it accordingly

PR #2717, opened 2023-10-31
Adopt global security contexts for Aggregator StatefulSet

chipzoller:
## What does this PR change?

Configures the Aggregator StatefulSet to be able to adopt the defaults from the `global` map if present. Also comments out the aggregator-specific security context settings allowing global values to take precedence.

No

Allows simpler Helm values for security contexts while still providing granular overrides.

N/A

Aggregator may fail to work properly.

## How was this PR tested?

Tested by local templating and deploying to an OCP cluster, ensuring the Pod was deployed.

N/A

### DIFF
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -53,7 +53,10 @@ spec:
       restartPolicy: Always
       {{- if .Values.kubecostAggregator.securityContext }}
       securityContext:
-          {{- toYaml .Values.kubecostAggregator.securityContext | nindent 10 }}
+        {{- toYaml .Values.kubecostAggregator.securityContext | nindent 8 }}
+      {{- else if .Values.global.securityContext }}
+      securityContext:
+        {{- toYaml .Values.global.securityContext | nindent 8 }}
       {{ end }}
       serviceAccountName: {{ template "aggregator.serviceAccountName" . }}
       volumes:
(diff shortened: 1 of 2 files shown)


cost-analyzer/templates/aggregator-statefulset.yaml
```diff
@@ -79,6 +81,9 @@ spec:
         {{- if .Values.kubecostAggregator.containerSecurityContext }}
           securityContext:
             {{- toYaml .Values.kubecostAggregator.containerSecurityContext | nindent 12 }}
+        {{- else if .Values.global.containerSecurityContext }}
+          securityContext:
+            {{- toYaml .Values.global.containerSecurityContext | nindent 12 -}}
```
ameijer: are you good with the trailing hyphen here? Will that pull the next key in to the back of this?

chipzoller: Thanks for calling that out. It won't have a rendering effect but it shouldn't be there. This was a copy-and-paste artifact from another template (that should have also been removed). Pushed that now.

ameijer (approved):
LGTM

chipzoller:
@jessegoodier, @thomasvn: Are you guys ok with this?

thomasvn:
For context, this is the current `.Values.global.securityContext`:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/8b463fdd72fe6dc191ae53b7bba0267c5f54a077/cost-analyzer/values.yaml#L221-L236

This is the current `.Values.kubecostAggregator.securityContext`:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/8b463fdd72fe6dc191ae53b7bba0267c5f54a077/cost-analyzer/values.yaml#L1004-L1020

Looks like there are some minor differences with the `securityContext`. And this PR would now be taking the global `securityContext` and commenting out the kubecostAggregator `securityContext`. I'm happy to get this into v1.107 so long as we have a couple days to run the nightly build with this change? Make sure no permission issues here?

chipzoller:
If you look carefully, you'll see the only differences are in the container-level security context. These values are already present at the `spec` level so they take precedence if the same fields are not defined at the container level. The net effect is the same. Still, I definitely would rather us test to be certain and not guess.

jessegoodier:
I'll CP quick

PR #2723, opened 2023-11-01
add SA

JesseGoodier:
## What does this PR change?
Add service account to etl-util pod

No release-notes requried.

### DIFF
--- a/cost-analyzer/templates/etl-utils-deployment.yaml
+++ b/cost-analyzer/templates/etl-utils-deployment.yaml
@@ -35,7 +35,7 @@ spec:
             defaultMode: 420
             secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret }}
         {{- end }}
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/etl-utils-deployment.yaml
```diff
@@ -35,6 +35,7 @@ spec:
             defaultMode: 420
             secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret }}
         {{- end }}
+      serviceAccountName: {{ template "aggregator.serviceAccountName" . }}
```
ameijer: there could be circumstances where the etl utils could be deployed and the aggregator could not, right? maybe the cost analyzer SA template, and also, we provide a means to give this its own SA?

jessegoodier: good point!
one moment.

ameijer (approved):

PR #2734, opened 2023-11-03
CI testing improvements

### LINKED ISSUE #2515: [CI] Use official Helm chart-testing
Helm has its own official chart testing action, based upon the eponymous CLI tool, called [chart-testing](https://github.com/helm/chart-testing-action) which may be good to standardize on.

Try and adopt chart-testing in CI tests.

Use existing which may have gaps.

Internal only.

### LINKED ISSUE #2708: Helm tests for aggregator
We need to start building a matrix of tests for the Helm chart. Since the aggregator is a new component and an important configuration, we could start with below as suggested.

```yaml
kubecostAggregator:
  enabled: true
  cloudCost:
    enabled: true
kubecostModel:
kubecostProductConfigs:
  cloudIntegrationSecret: cloud-integration
```

_Originally posted by @thomasvn in https://github.com/kubecost/cost-analyzer-helm-chart/issues/2699#issuecomment-1786201083_

chipzoller:
## What does this PR change?

Closes #2708 
Closes #2515 

This PR dramatically expands and improves chart testing in this repository by setting up an extensible framework, documenting it, and adding 3 files which cover most of the common Kubecost deployment configurations. For further details on this PR, see the DEVELOPMENT.md guide contained herein.

No

No user impact. This is internal only.

CI could break.

## How was this PR tested?

Extensive testing in fork.

Developer docs updated in DEVELOPMENT.md


.github/workflows/chart.yaml
```diff
@@ -60,19 +72,65 @@ jobs:
...
+
+      - name: Uninstall chart
+        run: helm uninstall kubecost -n kubecost --no-hooks --wait
+
+      - name: Cleanup all Pods in Kubecost Namespace
+        run: kubectl -n kubecost delete deployments,daemonsets,statefulsets,pods --all --force
+
+      # Loops over all the other values files in the ci directory and installs the chart, runs Helm tests, and uninstalls.
+      # Additional sleeps seem necessary for the main Kubecost Pod to be completely available, both frontend and backend.
+      - name: Install, test, remove chart for other values
```
thomasvn, line 113: Could you help clarify how this test, which installs all yaml files in `cost-analyzer/ci` one by one, differs from the tests run above? Are the tests run above for a default install of Kubecost?

.github/workflows/chart.yaml
```diff
@@ -60,19 +72,65 @@ jobs:
...
+            echo "### Performing installation with values from $file ###"
+            helm install --wait --wait-for-jobs kubecost cost-analyzer/ -n kubecost -f "$file"
+            echo "### Waiting for cost-analyzer Pod readiness. ###"
+            kubectl wait -n kubecost --for=condition=ready pod --selector app.kubernetes.io/name=cost-analyzer --timeout=120s
+            echo "### Sleeping for 60 seconds longer. ###"
+            sleep 60
+            echo "### [DEBUG] Listing Pods in Kubecost Namespace. ###"
+            kubectl -n kubecost get pods
+            echo "### Performing helm tests ###"
+            helm test -n kubecost $(helm ls -n kubecost --all --short)
```
thomasvn, line 130: Could you help clarify what exactly `helm test` does once Kubecost is installed?

.github/workflows/chart.yaml
```diff
@@ -14,22 +14,34 @@ jobs:
...
+      # Run `helm template` on the main values file plus all those in the ci directory.
       - name: Helm template
-        working-directory: ./cost-analyzer
-        run: helm template . --set global.prometheus.enabled=false --set global.grafana.enabled=false > full.yaml
+        run: |
+          helm template cost-analyzer/ -f cost-analyzer/values.yaml > full.yaml
+          directory="cost-analyzer/ci"
+          for file in "$directory"/*; do
+          if [ -f "$file" ]; then
+              helm template cost-analyzer/ -f "$file" >> full.yaml
```
thomasvn, line 31: If we are looping over the entire `cost-analzyer/ci` directory. Then rendering the templates and appending output to `full.yaml`

Is the expected behavior for `Kubeconform` to later fail if at least one of these `values.yaml` files have rendered incorrectly?

thomasvn:
@chipzoller Really awesome! Glad that we can now test deployments of these common configurations upon each PR. Only a few minor comments here, mostly questions for clarity.

Also, is there any reason the `aggregator-values.yaml` file is so large? Could we remove all the excess and just include what's pertinent to aggregator?

.github/workflows/chart.yaml
chipzoller: The tests here, as opposed to the previous one, run across the values in the `ci` directory. So the total tests include the one with the default `values.yaml` file plus the values files in the `ci` directory.

.github/workflows/chart.yaml
chipzoller: Sure, although [here](https://helm.sh/docs/topics/chart_tests/) are the docs. At this point in time, `helm test` will only execute the one test case I've written. After a chart is deployed, running `helm test` will run the test which is included as part of the chart. It is invoked only when the `test` command is used.

.github/workflows/chart.yaml
chipzoller: No, Kubeconform is only used to assess conformance of the final, rendered output against Kubernetes resource specs. For example, let's say you had `spec.foo: bar` in a Pod. While this may template correctly, it's not conformant to the Pod specification. A `helm template` would not catch this because the templating may occur successfully, but the final resource does not conform to the known Pod spec. Kubeconform is used to ensure that the (successfully) templated output are actually valid K8s resources.

chipzoller:
Great questions, as always.

> Also, is there any reason the `aggregator-values.yaml` file is so large? Could we remove all the excess and just include what's pertinent to aggregator?

No particular reason, I suppose. I used an existing config in the PoC repo plus your suggestions (plus a couple of my own, ex., enabling network costs) to come up with this manifest. If we can slim it down that'd be ideal. Do you have any recommendations?

thomasvn:
How about using these values for the Aggregator? ([`aggregator-minimal.yaml` internal repo](https://github.com/kubecost/clusters/blob/main/how-to/aggregator-minimum.yaml))

.github/workflows/chart.yaml
thomasvn: Oh I see now, thanks for linking to docs!

chipzoller:
Can I opt out of using an existing ServiceAccount? This will reduce the dependencies by one. It doesn't impact the ability to determine if aggregator works properly, correct?

thomasvn:
Is this the ServiceAccount you're referring to? Aggregator shouldn't need it!

```
# when using managed identity/irsa, set the service account accordingly:
serviceAccount:
  create: false
  name: kubecost-irsa-sa
```

chipzoller:
Can `kubecostProductConfigs.productKey.key` just remain as `YOUR_KEY` or does it have to be something valid?

chipzoller:
Hey @thomasvn, can you help me refine this so we can get it in?

thomasvn:
@chipzoller Thoughts on omitting `kubecostProductConfigs.productKey.key` from any CI tests?

Even if we put a dummy string here, it wouldn't be testing much functionality. And we also wouldn't want to be putting a valid key in these CI tests unless it's possible to store this secret outside git.

chipzoller:
Omitting that sounds good, I just don't know what happens if the value isn't provided. Can you summarize the changes you'd like to see in this PR?

jessegoodier:
by default, no value is needed. The majority of installs are free.

chipzoller:
But can it be null or does it have to be `""`?

thomasvn:
@chipzoller Happy to get this PR in ASAP! No major concerns with it. I think as a minor improvement we can use a much more minimal `aggregator-values.yaml`.

When I configured a minimal aggregator, these were the only configs I needed:

```yaml
# kubectl create secret generic federated-store --from-file=federated-store.yaml
# kubectl create secret generic cloud-integration --from-file=cloud-integration.json
kubecostAggregator:
  enabled: true
  cloudCost:
    enabled: true
  aggregatorStorage:
    storageRequest: 5Gi
  aggregatorDbStorage:
    storageRequest: 10Gi
kubecostModel:
kubecostProductConfigs:
  cloudIntegrationSecret: cloud-integration
```

chipzoller:
@thomasvn, made the changes. Please review.

thomasvn (approved):

PR #2735, opened 2023-11-04
Minor template fixes

chipzoller:
## What does this PR change?

Makes minor adjustments to templates to fix invalid outputs addressing things like null/empty string fields, empty structs, or resources which are (basically) empty. These fixes make it so that the fully rendered output is conformant for Kubernetes linters. Two of the most visible changes:

1. Role and RoleBinding for Grafana will not be created when PSP is not in use. 
2. UTC_OFFSET now has a value of `+00:00` (adopted from EKS values) whereas previously the env var was set with an empty value.

No

No user impact

Change has an unintended consequence.

## How was this PR tested?

Templated locally for each change.

### DIFF
--- a/cost-analyzer/charts/grafana/templates/role.yaml
+++ b/cost-analyzer/charts/grafana/templates/role.yaml
@@ -1,5 +1,4 @@
-{{ if .Values.global.grafana.enabled }}
-{{- if .Values.rbac.create }}
+{{ if and .Values.global.grafana.enabled .Values.rbac.create .Values.rbac.pspEnabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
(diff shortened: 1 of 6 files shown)


cost-analyzer/charts/grafana/templates/role.yaml
```diff
@@ -1,5 +1,4 @@
-{{ if .Values.global.grafana.enabled }}
-{{- if .Values.rbac.create }}
+{{ if and .Values.global.grafana.enabled .Values.rbac.create .Values.rbac.pspEnabled }}
```
thomasvn, line 1: Is there any situation where we would want to create this role even when `.Values.rbac.pspEnabled=false`?

thomasvn:
@chipzoller Thanks for these improvements! Was there a tool you used to spot these issues?

cost-analyzer/charts/grafana/templates/role.yaml
chipzoller: I can't see how. Given that the Role appears to exist to service a single permission, and that permission seems to be PSP, why create an empty role if that isn't enabled?

chipzoller:
Some of them I find when templating the chart and piping the resulting YAML into VS Code to be validated by the K8s extension, and some are just old-fashion eagle eyes.

cost-analyzer/charts/grafana/templates/role.yaml
thomasvn: Yeah, I'm thinking the same. Let's move forward with this change!

thomasvn (approved):

PR #2754, opened 2023-11-15
Hackathon. Cronjob to collect data for PowerBI integration.

thomasvn:
## What does this PR change?

Currently testing using the following `values-integrations.yaml`:

```yaml
# helm template ./cost-analyzer -f values-integrations.yaml -s templates/data-export-postgres-cronjob-template.yaml > integrations.yaml

integrations:
  postgres:
    enabled: true
    imagePullSecrets: my-docker-key
    fullImageName: docker.io/mclean0328/kube-postgres-cron:latest
    kubecostEndpoint: "https://kubecost.me.com/model/allocation"
    databaseName: "postgres"
    databaseUser: "test"
    databasePassword: "test"
    databaseHost: "my.postgres.database.azure.com"
    databasePort: "5432"
    schedule: '0 * * * *'

```

TODO:
- Template needs sane defaults. Or template needs to throw error if user does not provide specific configs.
- Template needs to include metadata + labels similar to other templates in this repo.
- Update script to parse arguments. Based on the argument, run a different data export job.


cost-analyzer/values.yaml
```diff
@@ -1191,6 +1191,21 @@ costEventsAudit:
 ##
 # readonly: false
 
+## Kubecost Integrations
+## 
+integrations:
+  powerbi:
```
mclean0328: Wondering if the integration should just be called "postgres". The example we are working in terms of end to end pipeline is getting the data somewhere that powerbi can connect to pull it, but ultimately the job that runs is just going to be sending data to postgres and has nothing to do with powerbi

cost-analyzer/values.yaml
```diff
@@ -1191,6 +1191,21 @@ costEventsAudit:
...
+    enabled: true
+    imagePullSecrets: my-docker-key
+    fullImageName: docker.io/mclean0328/test-cron:latest
+    kubecostEndpoint: "https://kubecost.me.com/model/allocation"
```
mclean0328: The example kubecost endpoint should not include /model/allocation. The script takes that into account. So https://kubecost.me.com/ works just fine for the example.

cost-analyzer/templates/powerbi-cronjob-template.yaml
```diff
@@ -0,0 +1,54 @@
+{{ if .Values.integrations.powerbi.enabled }}
```
mclean0328: The template overall looks good. Just potentially changing all the 'powerbi' references in the values to 'postgres' if we think calling the integration 'postgres' is better

cost-analyzer/values.yaml
```diff
@@ -1191,6 +1191,21 @@ costEventsAudit:
 ##
 # readonly: false
 
+## Kubecost Integrations
+## 
+integrations:
+  powerbi:
+    enabled: true
+    imagePullSecrets: my-docker-key
+    fullImageName: docker.io/mclean0328/test-cron:latest
```
mclean0328, line 1301: Update the docker image to this "docker.io/mclean0328/kube-postgres-cron:latest "

cost-analyzer/templates/powerbi-cronjob-template.yaml
thomasvn: Updated!

cost-analyzer/values.yaml
thomasvn: Updated!

cost-analyzer/values.yaml
thomasvn: Updated!

As a TODO, we should try to get this image into the `kubecost1` or `guestbook` container registry.

cost-analyzer/values.yaml
thomasvn: Updated!

PR #2766, opened 2023-11-16
fix(cost-analyzer): queryService ServiceAccountName

### LINKED ISSUE #2764: [Bug] ServiceAccount ClusterRole wrong assumption
### Kubecost Helm Chart Version

1.107.1

### Kubernetes Version

1.27.7

### Kubernetes Platform

GKE

### Description

If `queryService` is used and `serviceAccount.create` is set to `false` or `true`, the same `serviceAccount.name` is used as for kubecost-cost-analyzer. This also leads to the rendering of ClusterRoles with the same `name`, which isn't allowed and so the Helm post-render step fails.

### Steps to reproduce

Own serviceAccount.name set and want to create serviceAccounts on our own.

`helm template --set="serviceAccount.create=false" --set="serviceAccount.name=foobar" --set="kubecostDeployment.queryServiceReplicas=1" --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer --version v1.107.1 | yq '. | select(.kind == "ClusterRole") | .metadata.name'`

Own serviceAccount.name set and want to create serviceAccounts **NOT** on our own.

`helm template --set="serviceAccount.create=true" --set="serviceAccount.name=foobar" --set="kubecostDeployment.queryServiceReplicas=1" --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer --version v1.107.1 | yq '. | select(.kind == "ClusterRole") | .metadata.name'`

### Expected behavior

Different ClusterRoles and ServiceAccounts are present and used, one for kubecost (cost-analyzer) and one for queryService

### Impact

Helm install not possible with queryService being enabled.

see also https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/_helpers.tpl#L125-L138

### Screenshots

_No response_

### Logs

_No response_

### Slack discussion

_No response_

### Troubleshooting

- [X] I have read and followed the [issue guidelines](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/ISSUE_GUIDELINES.md) and this is a bug impacting only the Helm chart.

sebbrandt87:
resolves #2764

## What does this PR change?

We now can set a value for the queryService ServiceAccount name and can decide if we want to create it, so that we can also reuse our own existing ones.

no

- fix of query-service ServiceAccount name and all related resources consuming this name

resolves #2764

none, as the old behaviour is still respected but it now uses the correct generated ServiceAccount name

## How was this PR tested?

`helm template . --set="kubecostDeployment.queryService.serviceAccount.create=false" --set="kubecostDeployment.queryService.serviceAccount.name=kc-qs-test"  --set="kubecostDeployment.queryServiceReplicas=1" | yq '. | select(.kind == "ClusterRole") | .metadata.name'`

`helm template . --set="kubecostDeployment.queryService.serviceAccount.name=kc-qs-test"  --set="kubecostDeployment.queryServiceReplicas=1" | yq '. | select(.kind == "ClusterRole") | .metadata.name'`

`helm template . --set="kubecostDeployment.queryServiceReplicas=1" | yq '. | select(.kind == "ClusterRole") | .metadata.name'`

no


cost-analyzer/templates/_helpers.tpl
```diff
@@ -130,10 +130,10 @@ Create the name of the service account
 {{- end -}}
 {{- end -}}
 {{- define "query-service.serviceAccountName" -}}
-{{- if .Values.serviceAccount.create -}}
-    {{ default (include "query-service.fullname" .) .Values.serviceAccount.name }}
+{{- if .Values.kubecostDeployment.queryService.serviceAccount.create -}}
```
jessegoodier, line 133: If I'm reading this correctly- 
unfortunately, this will change the account defaults and there are users that will break if they don't make a change before upgrading.
Perhaps the .Values.queryService.serviceAccount.create needs to be false by default and will still allow your override to work here?

chipzoller (commented):
Cursory looks ok

nikovacevic:
This also looks fine to me -- @jessegoodier know of any reason this would be a problem?

jessegoodier:
> This also looks fine to me -- @jessegoodier know of any reason this would be a problem?

My concern is that the cost-analyzer SA is being used by default today. That SA could be using a CSP managed identity (irsa).
This PR would, by default, use a different account without privs. Any QSR user would need to update their config before upgrading.
So, not ideal.

Please correct me if I'm not reading this correctly.

sebbrandt87:
> > This also looks fine to me -- @jessegoodier know of any reason this would be a problem?
> 
> My concern is that the cost-analyzer SA is being used by default today. That SA could be using a CSP managed identity (irsa). This PR would, by default, use a different account without privs. So, any QSR user would need to update their config before upgrading. So, not ideal.
> 
> Please correct me if I'm not reading this correctly.

If right now, the cost-analyzer SA is created by the Helm Chart, which is the default, the SA name is already `{{ default (include "query-service.fullname" .)` if  `.Values.serviceAccount.name` is not set. Sure it is always set by default Helm Value, but also can be set to an empty string. Maybe there are already customers out there using it this way.
Should we "fallback" here to a logical wrong set SA name for this case, but we would not be able to render the Helm Chart with non duplicate ClusterRole names?

But if you want to create your own ServiceAccount you render / template the Helm Chart right now as well with a duplicate ClusterRole with the same name which is controlled by the SA name defined. As I tried to explain here https://github.com/kubecost/cost-analyzer-helm-chart/issues/2764 

You concern might can be ruled out as of, if an IRSA (AWS) or WorkloadIdentity (GCP) is added to a ServiceAccount, it would be managed by the `annotations` via which we could implement a fallback here(https://github.com/kubecost/cost-analyzer-helm-chart/pull/2766/files#diff-4ecbebc5ba2d33ed58610641848ef7de912e0c31b96c8bc1bb934aaac492e2b5R11) to act still the same way as before if not a dedicated annotation for the query-service ServiceAccount is defined and fallback to the cost-model service account annotation?

JesseGoodier (approved):
Thanks for your help here. I finally read your issue and agree. 
I see little risk to this change.

PR #2783, opened 2023-11-22
Autocomplete for allocation, assets, and cloud costs

nikovacevic:
## What does this PR change?
* Autocomplete for allocation, assets, and cloud costs on aggregator

* Requires https://github.com/kubecost/kubecost-cost-model/pull/1928

* Enables autocomplete for filters and other fields

* Risks involve performance at scale

## How was this PR tested?
* Tested for correctness by me; see PR notes on the KCM PR
* Tested for performance at scale by Michael; see same PR notes


michaelmdresser (approved):
I'm okay with this shipping without cache first, then we can make a separate PR for that to avoid too much happening at once.

PR #2788, opened 2023-11-24
Remove frontend port from service when it is disabled

abaguas:
## What does this PR change?
When the kubecost frontend is disabled, `kubecostFrontend.enabled: false`, the service should not expose the frontend port. Exposing the port has no impact on the kubecost application but has the undesired effect that the port is still discovered by the ServiceMonitor which causes `TargetDown` alerts.
This is a follow up of #2528 where the frontend container and volumes were removed but the service was forgotten.

No

No, it doesn't require any user configuration changes.

No risk. To fully test the PR take `master` and edit values.yaml to set `kubecostFrontend.enabled: false`. Generate the manifest:
```
helm dependency update
helm package . && helm template cost-analyzer-1.107.1.tgz -n kubecost > manifests_after.yaml
```
Then, take this branch, set again `kubecostFrontend.enabled: false`, and generate the manifests. The diff is the removal of the following lines from `# Source: cost-analyzer/templates/cost-analyzer-service-template.yaml`:
```
    - name: tcp-frontend
      port: 9090
      targetPort: 9090
```


chipzoller:
This LGTM and is simple but would like @jessegoodier and @thomasvn to have a look.

chipzoller (approved):
LGTM

thomasvn (approved):
Thanks!

PR #2795, opened 2023-11-29
add extraPorts config for development environments

saweber:
## What does this PR change?
Adds options for extra ports to a running installation. The only current use for this is attaching a debugger to a back end that is running through tilt (for development purposes only).

Note: I am open to alternatives here (especially if there is a cleaner way to do this!); the key thing is I need to open up a port to the back end specifically for delve.

This functionality is identical to opencost - see https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/values.yaml#L51

No

No impact - only current use is for internal development only.

n/a

To test the PR, see that an extra port can be configured to reach the back-end
Risks are very low - no change occurs for current users.

## How was this PR tested?
With `hackathon-tilt` branch of KCM

not documenting this until we have a user-facing use for this

### DIFF
--- a/cost-analyzer/templates/aggregator-service.yaml
+++ b/cost-analyzer/templates/aggregator-service.yaml
@@ -16,5 +16,8 @@ spec:
     - name: tcp-api
       port: 9004
       targetPort: 9004
+  {{- with .Values.kubecostAggregator.extraPorts }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 {{- end }}
 {{- end }}
--- a/cost-analyzer/templates/query-service-deployment-template.yaml
+++ b/cost-analyzer/templates/query-service-deployment-template.yaml
@@ -118,6 +118,9 @@ spec:
             - name: tcp-model
               containerPort: 9003
               protocol: TCP
+          {{- with .Values.kubecostDeployment.queryService.extraPorts }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           resources:
             {{- toYaml .Values.kubecostDeployment.queryService.resources | nindent 12 }}
           volumeMounts:
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -113,6 +113,9 @@ spec:
             - name: tcp-api
               containerPort: 9004
               protocol: TCP
+          {{- with.Values.kubecostAggregator.extraPorts }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           resources:
             {{ toYaml .Values.kubecostAggregator.resources | nindent 12 }}
           volumeMounts:
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -494,6 +494,9 @@ spec:
           - name: tcp-frontend
             containerPort: 9090
             protocol: TCP
+        {{- with .Values.kubecostModel.extraPorts }}
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
           resources:
 {{ toYaml .Values.kubecostModel.resources | indent 12 }}
           readinessProbe:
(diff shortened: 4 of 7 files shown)


cost-analyzer/templates/aggregator-service.yaml
```diff
@@ -16,5 +16,8 @@ spec:
     - name: tcp-api
       port: 9004
       targetPort: 9004
+      {{- with .Values.kubecostModel.extraPorts }}
```
michaelmdresser: `.Values.kubecostAggregator.extraPorts` I think makes more sense. Same deal for other aggregator files.

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -118,6 +118,9 @@ spec:
             - name: tcp-model
               containerPort: 9003
               protocol: TCP
+             {{- with .Values.kubecostModel.extraPorts }}
```
michaelmdresser: Does query service have a separate heading other than `kubecostModel`?

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -118,6 +118,9 @@ spec:
...
+             {{- toYaml . | nindent 12 }}
```
michaelmdresser: Does this line have to be indented two more spaces to match the others or does it not matter?

kubecost.yaml
```diff
@@ -19942,6 +19942,9 @@ spec:
     - name: tcp-model
       port: 9003
       targetPort: 9003
+    - name: debug
```
michaelmdresser: I don't think we should be updating `kubecost.yaml` in this PR

chipzoller (changes requested):
Please add this new field into the main `values.yaml` file. We must not update templates with fields which have not been documented, even if deactivated by default.

cost-analyzer/templates/aggregator-service.yaml
saweber: Agreed. The original thought was to minimize the number of additional configurations to manage on the tilt side, but in retrospect this feels like a minor footgun.

cost-analyzer/templates/query-service-deployment-template.yaml
saweber: From what I can tell - it seems to be purely style. The nindent value is what matters. I changed the indentation to better match elsewhere and add clarity - ensure that the toYaml is appropriately indented to its parent.

cost-analyzer/templates/query-service-deployment-template.yaml
saweber: Yes. moving to kubecostDeployment.queryService.extraPorts, same reasoning as the aggregator comment.

cost-analyzer/templates/aggregator-service.yaml
```diff
@@ -16,5 +16,8 @@ spec:
     - name: tcp-api
       port: 9004
       targetPort: 9004
+  {{- with .Values.kubecostAggregator.extraPorts }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
michaelmdresser, line 21: Is this indentation off by one space? (it may not matter)

saweber: I'm seeing in my IDE:
indentation of 2 spaces for the conditional
indentation of 4 spaces for the yaml insertion

The intention is to put the toYaml statement at the same indentation of the output yaml files. That seems to be the more common style in this repository, though I see examples of zero indentation as well.

The actual indentation here does not matter, just a matter of us defining a style and sticking to it.

chipzoller: Best practice here is to place the template at the same level of indentation as the values being inserted. We're very inconsistent here which is something I'd like to work on.

saweber: I would like to see that, especially if that can be auto-linted. :)

chipzoller:
Conflicts to resolve.

michaelmdresser (approved):

chipzoller (approved):

PR #2796, opened 2023-11-30
Sean/account mapping

Sean-Holcomb:
## What does this PR change?
This PR adds a template for a new ConfigMap which holds a mapping from Account ID and Account Name to allow the human readable Account Name to be displayed by the UI in place of the Account ID. While adding a new ConfigMap for each new config from ideal it is the current president and isn't easy to undo for this task. I have created this ConfigMap in such a way that it will be easier to join with other Config Maps at a later date.

Additionally it fixes some faulty templates for other ConfigMaps 

https://github.com/kubecost/kubecost-cost-model/pull/1944

## How was this PR tested?
Tested via template and deployment


ameijer (approved):

nikovacevic (approved):

cost-analyzer/values.yaml
```diff
@@ -1290,6 +1290,12 @@ costEventsAudit:
 #    daemonset_external_label: "kubernetes_daemonset"
 #    pod_external_label: "kubernetes_pod"
 #  grafanaURL: ""
+#  # Provide a mapping from Account ID to a readable account name which will be displayed in the UI for CloudCosts
+#  # Account ID refers to different entities for different entities for each provider:
+#  # - AWS: Account ID
+#  # - Azure: Subscription ID
+#  # - GCP: Project ID
+#  cloudAccountMapping: {}
```
dwbrown2: Worth adding an example for users here? It's not entirely clear what format I should pass to this newb user.

chipzoller:
Moving to draft state because there is a dependency which has not been merged. Please mark as ready for review when the dependency has been resolved.

cost-analyzer/values.yaml
chipzoller: Agreed. The comment structure is also confusing here.

Sean-Holcomb: updated

cost-analyzer/values.yaml
```diff
@@ -1290,6 +1290,10 @@ costEventsAudit:
 #    daemonset_external_label: "kubernetes_daemonset"
 #    pod_external_label: "kubernetes_pod"
 #  grafanaURL: ""
+#  # Provide a mapping from Account ID to a readable Account Name in a key/value object. Provide Account IDs as they are displayed in CloudCost
```
dwbrown2, line 1293: Might be helpful to indicate that these values would be shown in both Kubecost UI + APIs?

dwbrown2:
cc @bstuder99 as we may want to document this behavior...

Sean-Holcomb:
> Moving to draft state because there is a dependency which has not been merged. Please mark as ready for review when the dependency has been resolved.

How are you determining that the related PR is a dependency of this PR and not the other way around?

cost-analyzer/values.yaml
Sean-Holcomb: The related PR exposes them 
https://github.com/kubecost/kubecost-cost-model/pull/1944

chipzoller:
> > Moving to draft state because there is a dependency which has not been merged. Please mark as ready for review when the dependency has been resolved.
> 
> How are you determining that the related PR is a dependency of this PR and not the other way around?

Because you stated in the initial post that this PR relied upon an external PR, which I see has now been merged.

Sean-Holcomb:
> > > Moving to draft state because there is a dependency which has not been merged. Please mark as ready for review when the dependency has been resolved.
> > 
> > 
> > How are you determining that the related PR is a dependency of this PR and not the other way around?
> 
> Because you stated in the initial post that this PR relied upon an external PR, which I see has now been merged.

Ah I see, that section is titled "Does this PR relate to any other PRs?" in KCM and OpenCost. We generally use it as a bi-directional link between related work .

Strictly speaking the Helm chart has no dependencies on KCM or OpenCost

chipzoller:
Ok, I see. It has just come up a few times at this point that some of these PRs do, in fact, rely (i.e., hard incoming dependency) upon code PRs in sister repositories and that merging a PR here prior to one being merged elsewhere causes breaking changes in CI or tests. I'm just trying to help us avoid that by suggesting some ways on how to make this more clear.

Sean-Holcomb:
From your comment it sounds like we need a mechanism to ensure that related PR's are all merged before our Nightly CI/CD pipeline runs. Maybe we could do something with tagging?

chipzoller:
> From your comment it sounds like we need a mechanism to ensure that related PR's are all merged before our Nightly CI/CD pipeline runs. Maybe we could do something with tagging?

That might be possible, I admit this is really something I haven't spent a lot of time doing. One of the challenges here is when you have two separate repos, one being private, in a single org.

brstuder:
Is there a reason this feature is Aggregator only?

dwbrown2:
> Is there a reason this feature is Aggregator only?

cc @teevans @kwombach12

Sean-Holcomb:
@bstuder99 it is not, or is not limited to be here

brstuder:
Gotcha, I just got the draft release notes and this was labeled as Aggregator only so I just wanted to confirm

Sean-Holcomb:
> That might be possible, I admit this is really something I haven't spent a lot of time doing. One of the challenges here is when you have two separate repos, one being private, in a single org.

Agreed, and automated solution might be too challenging. A cursory search doesn't bring up anything promising.

PR #2809, opened 2023-12-04
Update Network Costs to v0.17.2

mbolt35:
* Update Network Costs to v0.17.2 which uses the latest Alpine image


cliffcolvin (approved):
LGTM

PR #2815, opened 2023-12-05
[108] OpenShift support

chipzoller:
## What does this PR change?

* Adds formal OpenShift support via a simple abstraction layer by setting securityContext values properly and enables creation of an OpenShift Route resource. Although this new `platforms` abstraction layer shouldn't have been necessary, due to Helm's mistreatment of being able to delete values from subcharts, this was the only option.
* Adjusts precedence order of Pod-level securityContext for all Pod controllers which used `global.securityContext`. Now, uniformly the most granular level applies first followed by OpenShift (if enabled) and then global. Values files in subcharts have been adjusted accordingly. For non-OpenShift deployments, this change should have no impact on the resulting securityContext values.
* Allows creation of SecurityContextConstraints for both Prometheus node exporter and Kubecost network costs.
* Fixes a couple errors and lints
* Updates CI pipeline to test OCP values on OCP clusters

This PR obviates the need for the [kubecost/openshift-helm-chart](https://github.com/kubecost/openshift-helm-chart) repository and it will be marked for deprecation.

No

Adds formal support for OpenShift

Templating could break for non-OCP installs. OCP installs might not fully function. An actual OCP cluster is required to test this.

## How was this PR tested?

Changes were tested by enabling all the components where templates were updated and deploying to an OKD cluster in various configurations.

https://github.com/kubecost/docs/issues/834

### DIFF
--- a/cost-analyzer/charts/prometheus/templates/node-exporter-ocp-scc.yaml
+++ b/cost-analyzer/charts/prometheus/templates/node-exporter-ocp-scc.yaml
@@ -1,0 +1,29 @@
+{{- if and (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") (.Values.global.platforms.openshift.scc.nodeExporter) (.Values.nodeExporter.enabled) }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{ template "prometheus.nodeExporter.fullname" . }}
+priority: 10
+allowPrivilegedContainer: true
+allowHostDirVolumePlugin: true
+allowHostNetwork: true
+allowHostPorts: true
+allowHostPID: true
+allowHostIPC: false
+readOnlyRootFilesystem: false
+runAsUser:
+  type: RunAsAny
+fsGroup:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+seccompProfiles:
+- runtime/default
+volumes:
+  - hostPath
+  - projected
+users:
+  - system:serviceaccount:{{ .Release.Namespace }}:{{ template "prometheus.serviceAccountName.nodeExporter" . }}
+{{- end }}
--- a/cost-analyzer/charts/prometheus/values.yaml
+++ b/cost-analyzer/charts/prometheus/values.yaml
@@ -865,11 +865,11 @@ server:
 
   ## Security context to be added to server pods
   ##
-  securityContext:
-    runAsUser: 1001
-    runAsNonRoot: true
-    runAsGroup: 1001
-    fsGroup: 1001
+  securityContext: {}
+    # runAsUser: 1001
+    # runAsNonRoot: true
+    # runAsGroup: 1001
+    # fsGroup: 1001
 
   containerSecurityContext: {}
 
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -116,6 +116,9 @@ spec:
             - name: tcp-api
               containerPort: 9004
               protocol: TCP
+          {{- with.Values.kubecostAggregator.extraPorts }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           resources:
             {{ toYaml .Values.kubecostAggregator.resources | nindent 12 }}
           volumeMounts:
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -51,13 +51,16 @@ spec:
         app: aggregator
     spec:
       restartPolicy: Always
-      {{- if .Values.kubecostAggregator.securityContext }}
+    {{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.securityContext }}
       securityContext:
-        {{- toYaml .Values.kubecostAggregator.securityContext | nindent 8 }}
-      {{- else if .Values.global.securityContext }}
+    {{- toYaml .Values.global.platforms.openshift.securityContext | nindent 8 }}
+    {{- else if .Values.global.securityContext }}
       securityContext:
-        {{- toYaml .Values.global.securityContext | nindent 8 }}
-      {{ end }}
+    {{- toYaml .Values.global.securityContext | nindent 8 }}
+    {{- else if .Values.kubecostAggregator.securityContext }}
+      securityContext:
+    {{- toYaml .Values.kubecostAggregator.securityContext | nindent 8 }}
+    {{- end }}
       serviceAccountName: {{ template "aggregator.serviceAccountName" . }}
       volumes:
         {{- $etlBackupBucketSecret := "" }}
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -72,6 +72,8 @@ spec:
           secret:
             defaultMode: 420
             secretName: {{ $etlBackupBucketSecret }}
+        {{- else }}
+        {{- fail "Kubecost Aggregator requires .Values.kubecostModel.federatedStorageConfigSecret" }}
         {{- end }}
       containers:
       {{- if .Values.kubecostAggregator.jaeger.enabled }}
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -49,6 +49,10 @@ spec:
         app.kubernetes.io/name: aggregator
         app.kubernetes.io/instance: {{ .Release.Name }}
         app: aggregator
+      {{- with .Values.global.podAnnotations}}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
       restartPolicy: Always
     {{- if .Values.kubecostAggregator.securityContext }}
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -51,15 +51,15 @@ spec:
         app: aggregator
     spec:
       restartPolicy: Always
-    {{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.securityContext }}
+    {{- if .Values.kubecostAggregator.securityContext }}
+      securityContext:
+    {{- toYaml .Values.kubecostAggregator.securityContext | nindent 8 }}
+    {{- else if and (.Values.global.platforms.openshift.enabled) (.Values.global.platforms.openshift.securityContext) }}
       securityContext:
     {{- toYaml .Values.global.platforms.openshift.securityContext | nindent 8 }}
     {{- else if .Values.global.securityContext }}
       securityContext:
     {{- toYaml .Values.global.securityContext | nindent 8 }}
-    {{- else if .Values.kubecostAggregator.securityContext }}
-      securityContext:
-    {{- toYaml .Values.kubecostAggregator.securityContext | nindent 8 }}
     {{- end }}
       serviceAccountName: {{ template "aggregator.serviceAccountName" . }}
       volumes:
--- a/cost-analyzer/templates/query-service-deployment-template.yaml
+++ b/cost-analyzer/templates/query-service-deployment-template.yaml
@@ -44,12 +44,15 @@ spec:
       {{- end }}
     spec:
       restartPolicy: Always
-      {{- if .Values.kubecostDeployment.queryService.securityContext }}
+      {{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.securityContext }}
       securityContext:
-        {{- toYaml .Values.kubecostDeployment.queryService.securityContext | nindent 8 }}
+        {{- toYaml .Values.global.platforms.openshift.securityContext | nindent 8 }}
       {{- else if .Values.global.securityContext }}
       securityContext:
         {{- toYaml .Values.global.securityContext | nindent 8 }}
+      {{- else if .Values.kubecostDeployment.queryService.securityContext }}
+      securityContext:
+        {{- toYaml .Values.kubecostDeployment.queryService.securityContext | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ template "query-service.serviceAccountName" . }}
       volumes:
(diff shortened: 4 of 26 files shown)


jessegoodier:
This looks great!
we should get this in the RC and ask some users to try it out.

cost-analyzer/templates/aggregator-statefulset.yaml
```diff
@@ -51,13 +51,16 @@ spec:
...
       securityContext:
-        {{- toYaml .Values.kubecostAggregator.securityContext | nindent 8 }}
-      {{- else if .Values.global.securityContext }}
+    {{- toYaml .Values.global.platforms.openshift.securityContext | nindent 8 }}
+    {{- else if .Values.global.securityContext }}
       securityContext:
-        {{- toYaml .Values.global.securityContext | nindent 8 }}
-      {{ end }}
+    {{- toYaml .Values.global.securityContext | nindent 8 }}
+    {{- else if .Values.kubecostAggregator.securityContext }}
```
thomasvn: In this logic, `.Values.global.securityContext` takes precedence over `.Values.kubecostAggregator.securityContext`. Is that the desired behavior?

Maybe the best approach would be to take the securityContext defined from both configurations?

cost-analyzer/templates/query-service-deployment-template.yaml
```diff
@@ -44,12 +44,15 @@ spec:
...
       restartPolicy: Always
-      {{- if .Values.kubecostDeployment.queryService.securityContext }}
+      {{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.securityContext }}
       securityContext:
-        {{- toYaml .Values.kubecostDeployment.queryService.securityContext | nindent 8 }}
+        {{- toYaml .Values.global.platforms.openshift.securityContext | nindent 8 }}
       {{- else if .Values.global.securityContext }}
       securityContext:
         {{- toYaml .Values.global.securityContext | nindent 8 }}
+      {{- else if .Values.kubecostDeployment.queryService.securityContext }}
```
thomasvn: Same as above

cost-analyzer/templates/aggregator-statefulset.yaml
chipzoller: Yes, this was the desired behavior and an attempt to make it consistent. It's basically the same behavior (albeit a little inconsistent) with other changes I made. In any case, to make this work with OpenShift, the OpenShift values have to have the highest precedence order.

> Maybe the best approach would be to take the securityContext defined from both configurations?

That probably wouldn't work well and we aren't doing that anywhere else in the chart. The securityContext should be taken from one discrete instruction set so it's easier to reason about the final result without having to deal with merge hell.

cost-analyzer/templates/query-service-deployment-template.yaml
chipzoller: Posted comment above.

cost-analyzer/templates/aggregator-statefulset.yaml
jessegoodier: Seems like the right approach, given the legacy. but will defer to Thomas here.

chipzoller:
Adding of SecurityContextConstraints has been added for both Prometheus node exporter and network costs, so I think this is ready for review.

jessegoodier:
@thomasvn I'm good with this. You make the call.

chipzoller:
I discovered something here and am making final changes.

cost-analyzer/templates/aggregator-statefulset.yaml
chipzoller: Thinking more about this, we're actually doing this the inverted way. The most granular config option should win followed by increasingly higher levels of granularity. I've made that change here and am making it throughout all templates which name the `global.securityContext` struct to be consistent. The order of precedence should be Deployment => OpenShift => global where "Deployment" is the lowest-level Pod controller.

cost-analyzer/templates/query-service-deployment-template.yaml
chipzoller: Posted reply and making changes throughout for consistency sake.

cost-analyzer/charts/prometheus/templates/node-exporter-ocp-scc.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if and (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") (.Values.global.platforms.openshift.scc.nodeExporter) (.Values.nodeExporter.enabled) }}
```
thomasvn, line 1: Nice catch here. If I'm understanding correctly, OCP users who want to run Node Exporter need this `SecurityContextConstraints` to define extra permissions?

Is this how it currently works in the OpenShift Helm Chart as well?

cost-analyzer/charts/prometheus/values.yaml
```diff
@@ -865,11 +865,11 @@ server:
 
   ## Security context to be added to server pods
   ##
-  securityContext:
-    runAsUser: 1001
-    runAsNonRoot: true
-    runAsGroup: 1001
-    fsGroup: 1001
+  securityContext: {}
```
thomasvn, line 868: Any reason you're commenting out the `securityContext` in these subcharts? Will this affect existing users currently requiring these permission?

thomasvn:
@chipzoller Really like all the OpenShift changes here so far. Good catch on the `SecurityContextConstraints`!

I also agree with your design decision that the most granularly defined SecurityContext should probably be taking precedence over the globally defined SecurityContext. However I think we should address this in a separate PR, to keep this OpenShift PR tightly scoped? Particularly because inverting the existing order of precedence could cause breaking changes to users.

cost-analyzer/charts/prometheus/values.yaml
chipzoller: The reason these are commented out is because with the move in order of precedence, if they were not then they would apply. Currently, they do not, so these changes are basically no-ops so long as the global level securityContext is defined. In other words, these changes produce the same effect as what users experience today. It's only *IF* they decide to use more granular settings (which requires explicit action) will that result change.

cost-analyzer/charts/prometheus/templates/node-exporter-ocp-scc.yaml
chipzoller: The current fork doesn't offer this ability at all. I added it as a measure of convenience for users but disabled by default. Many users prefer to use their own SCCs because of corporate policy. They are free to do so, but one non-default SCC must apply. If it does not, the Pods will be unable to start because OCP does not ship with an SCC which allows these types of permissive Pods to run.

chipzoller:
See comments. With the combination of order inversion and commenting out the most granular controls in subchart values, this is essentially a no-op for users.

cost-analyzer/charts/prometheus/values.yaml
thomasvn: Oh I see! Currently, this prometheus-server is already taking the global security context:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/f481c59dc79e7eaf79dc6e607511d9eb52b1d67c/cost-analyzer/charts/prometheus/templates/server-deployment.yaml#L63-L65

However when changing the order of precedence across all securitycontexts, we need to comment out highest granularity (and currently unused) securitycontext for prometheus-server so that it continues to take the global security context.

thomasvn:
Just for extra assurance, could we do a `helm template` before and after the changes in this PR? Specifically want to make sure there are no diffs in securitycontext configs.

^^ This is also assuming default configs for securitycontext. What if a current user had mistakenly set a highly granular securitycontext. But was overriding it anyway with the global securitycontext? This PR may incur breaking changes? I'm ok if it does, but we may want to make users aware in release notes.

cost-analyzer/charts/prometheus/values.yaml
chipzoller: Yes, and those changes have been made in this PR.

chipzoller:
> Just for extra assurance, could we do a `helm template` before and after the changes in this PR? Specifically want to make sure there are no diffs in securitycontext configs. 

Sure, see terminal output below with comments interspersed. The first section marked "Current" is templating from `develop` as it currently exists. "After" is after the changes in this PR as it stands now.

> What if a current user had mistakenly set a highly granular securitycontext. But was overriding it anyway with the global securitycontext? This PR may incur breaking changes? I'm ok if it does, but we may want to make users aware in release notes.

Yes, in a case where a user is doing an upgrade (not an install) where they aren't pulling in the values set in the chart they might experience a failure in Prometheus or Grafana. Rolling that out again anew should fix that issue assuming their custom securityContext works anyway (it may not depending on platform and config). I think that's a pretty small concern, but I agree we should list this as a change in the release notes for 1.108.0 if this makes it in.

```sh
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Current
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s charts/grafana/templates/deployment.yaml | yq .spec.template.spec.securityContext
fsGroup: 1001
fsGroupChangePolicy: OnRootMismatch
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
seccompProfile:
  type: RuntimeDefault
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s charts/prometheus/templates/server-deployment.yaml | yq .spec.template.spec.securityContext
fsGroup: 1001
fsGroupChangePolicy: OnRootMismatch
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
seccompProfile:
  type: RuntimeDefault
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/aggregator-statefulset.yaml | yq .spec.template.spec.securityContext
Error: could not find template templates/aggregator-statefulset.yaml in chart
null
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Aggregator not enabled in previous template attempt.
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/cost-analyzer-deployment-template.yaml | yq .spec.template.spec.securityContext
fsGroup: 1001
fsGroupChangePolicy: OnRootMismatch
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
seccompProfile:
  type: RuntimeDefault
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/federator-deployment-template.yaml | yq .spec.template.spec.securityContext
Error: could not find template templates/federator-deployment-template.yaml in chart
null
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Federator not enabled in previous template attempt.
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/query-service-deployment-template.yaml | yq .spec.template.spec.securityContext
Error: could not find template templates/query-service-deployment-template.yaml in chart
null
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Query service not enabled in previous template attempt.
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ 
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ 
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # After
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s charts/grafana/templates/deployment.yaml | yq .spec.template.spec.securityContext
fsGroup: 1001
fsGroupChangePolicy: OnRootMismatch
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
seccompProfile:
  type: RuntimeDefault
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s charts/prometheus/templates/server-deployment.yaml | yq .spec.template.spec.securityContext
fsGroup: 1001
fsGroupChangePolicy: OnRootMismatch
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
seccompProfile:
  type: RuntimeDefault
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/aggregator-statefulset.yaml | yq .spec.template.spec.securityContext
Error: could not find template templates/aggregator-statefulset.yaml in chart
null
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Aggregator not enabled in previous template attempt.
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/cost-analyzer-deployment-template.yaml | yq .spec.template.spec.securityContext
fsGroup: 1001
fsGroupChangePolicy: OnRootMismatch
runAsGroup: 1001
runAsNonRoot: true
runAsUser: 1001
seccompProfile:
  type: RuntimeDefault
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/federator-deployment-template.yaml | yq .spec.template.spec.securityContext
Error: could not find template templates/federator-deployment-template.yaml in chart
null
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Federator not enabled in previous template attempt.
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ helm template . -s templates/query-service-deployment-template.yaml | yq .spec.template.spec.securityContext
Error: could not find template templates/query-service-deployment-template.yaml in chart
null
(⎈|k3d-kyv1111:N/A)chip@Schmitt:~/git/kubecost-cost-analyzer-helm-chart/cost-analyzer$ # Query service not enabled in previous template attempt.
```

JesseGoodier (approved):
templated with and without the values for openshift, compared to develop branch and all looks good.

thomasvn (approved):

chipzoller:
Cherry pick to 108 complete.

PR #2846, opened 2023-12-12
Updates defaults for Multi-cluster Diagnostics feature.

thomasvn:
## What does this PR change?

- `.Values.diagnostics.keepDiagnosticHistory=false` by default. Reduces load placed on bucket & diagnostics service.
- `.Values.diagnostics.isDiagnosticsPrimary.enabled=false` by default. Only one cluster needs this, and we shouldn't be enabling the service on all clusters by default.
- Fixes logic in configmap

No

We plan to default enable the Multi-cluster Diagnostics in v1.108. If users are running a Thanos or FedETL setup, they will see the following files uploaded to their storage buckets `/diagnostics/CLUSTER_ID.json`. One per cluster.

None

This has been internally used and under test for ~1 month.

The risk is that the new feature does not work for users, or causes unnecessary load. However based on the default configurations of this feature, I believe load should be minimal. Additionally, because all this logic is contained in a separate`deployment/kubecost-diagnostics`, there should be no impact to the main `deployment/kubecost-cost-analyzer`.

TODO

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -729,14 +729,12 @@ data:
 
         location /model/multi-cluster-diagnostics-enabled {
             default_type 'application/json';
-            {{- if .Values.diagnostics.isDiagnosticsPrimary.enabled }}
-            {{- if .Values.diagnostics.enabled }}
+            {{- if and .Values.diagnostics.enabled .Values.diagnostics.isDiagnosticsPrimary.enabled }}
             {{- if or .Values.global.thanos.enabled (not (empty .Values.kubecostModel.federatedStorageConfigSecret )) }}
             return 200 '{"multi-cluster-diagnostics-enabled": "true"}';
+            {{- end }}
             {{- else }}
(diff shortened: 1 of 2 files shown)


JesseGoodier (approved):

thomasvn:
CI tests are showing the following error:

```
Error: execution error at (cost-analyzer/templates/diagnostics-deployment.yaml:5:4): Error: The 'cluster_id' is set to default 'cluster-one'. Please update so that the diagnostics service can uniquely identify data coming from this cluster.
```

I believe because it's currently trying to template the configurations available at `./cost-analyzer/ci`, and [`aggregator-values.yaml`](https://github.com/kubecost/cost-analyzer-helm-chart/blob/ec23beb7ec42ecb31f0975c273fd53324fa8228d/cost-analyzer/ci/aggregator-values.yaml) still needs to set a `cluster_id`. I will open a PR to fix that. cc @chipzoller

thomasvn:
I've updated the CI test in the above PR. I'd like to get that merged, then rerun CI tests in this PR. Then merge & cherry-pick this PR.

thomasvn:
The primary test `test-chart` was successful, and I believe the other tests which setup environments got rate limited. I think this change is safe to merge!

PR #2887, opened 2023-12-20
change default nginx TLS port to 9090

JesseGoodier:
## What does this PR change?
Prior to this PR, any frontend configuration using TLS would require nginx to run on port 443. Using ports under 1025 requires that the container run as root. 

There is no need to run nginx with a different port, though this is still configurable via 
```yaml
service:
  type: ClusterIP
  port: 9090
  targetPort: 9090
```
If a user wants to use port 443, they will need to modify the SCC to allow root. 
We do not recommend running any container as root.

No

Allow users to use TLS without requiring nginx run as root. The default nginx frontend port for all configurations is 9090

Unknown configurations requiring ports under 443.

## How was this PR tested?
```
helm upgrade -i kubecost-eks10 \
. \
--namespace kubecost-eks10 \
--create-namespace \
--set kubecostFrontend.tls.enabled=true \
--set kubecostFrontend.tls.secretName=kubecost-eks10 \
```
and
```
helm upgrade -i kubecost-eks10 \
. \
--namespace kubecost-eks10 \
--create-namespace \
--set kubecostFrontend.tls.enabled=false
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -1,5 +1,4 @@
 {{- if and (not .Values.agent) (not .Values.cloudAgent) }}
-{{- $nginxPort := int .Values.service.port | default 9090 -}}
 apiVersion: apps/v1
 {{- if and .Values.kubecostDeployment.statefulSet.enabled .Values.kubecostDeployment.leaderFollower.enabled }}
 kind: StatefulSet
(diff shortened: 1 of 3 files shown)


chipzoller (commented):
Is the `nginxPort` template still needed after this change?

jessegoodier:
> Is the `nginxPort` template still needed after this change?

I can't think of a reason why it would be needed. 

You want to take a pass at removing that?

chipzoller:
It's locally defined only. L2

jessegoodier:
> It's locally defined only. L2

done.

chipzoller (approved):

PR #2889, opened 2023-12-21
clean up agent values - fix diagnostic pod logic

JesseGoodier:
## What does this PR change?
- update hosted agent to run diagnostics with thanos
- clean up agent values 

No

Fixed logic in diagnostic pod deployment

Unknown configs, though I believe this risk is low.

## How was this PR tested?
Local install, will be sharing with a customer asap.

NA

### DIFF
--- a/cost-analyzer/templates/diagnostics-deployment.yaml
+++ b/cost-analyzer/templates/diagnostics-deployment.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.diagnostics.enabled }}
-{{- if or .Values.global.thanos.enabled (not (empty .Values.kubecostModel.federatedStorageConfigSecret )) -}}
+{{- if or ( not (empty .Values.prometheus.server.extraSecretMounts )) (not (empty .Values.kubecostModel.federatedStorageConfigSecret )) -}}
 
 {{- if eq .Values.prometheus.server.global.external_labels.cluster_id "cluster-one" }}
 {{- fail "Error: The 'cluster_id' is set to default 'cluster-one'. Please update so that the diagnostics service can uniquely identify data coming from this cluster." }}
@@ -46,7 +46,7 @@ spec:
           secret:
            defaultMode: 420
            secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret }}
-        {{- else if .Values.global.thanos.enabled }}
+        {{- else }}
         - name: federated-storage-config
           secret:
             defaultMode: 420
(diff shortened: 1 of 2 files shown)


JesseGoodier (commented):
/cherry-pick v1.108

chipzoller (approved):

cost-analyzer/templates/diagnostics-deployment.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.diagnostics.enabled }}
-{{- if or .Values.global.thanos.enabled (not (empty .Values.kubecostModel.federatedStorageConfigSecret )) -}}
+{{- if or ( not (empty .Values.prometheus.server.extraSecretMounts )) (not (empty .Values.kubecostModel.federatedStorageConfigSecret )) -}}
```
thomasvn, line 2: I don't fully understand this change. Are we assuming that when users are running Thanos, they will always have the `.Values.prometheus.server.extraSecretMounts` set? If so we should leave a comment here.

jessegoodier: That is correct.

No doubt there's a better way to do this. Probably check if the list of sidecarContainers has one with an image containing thanos?

Though I'd also consider forcing thanos configs to add a flag to enable diagnostics.

I was just following the logic of the federated store. 

Here are the values for a thanos agent:

```yaml
prometheus:
  server:
    enableAdminApi: true
    extraArgs:
      storage.tsdb.max-block-duration: 2h
      storage.tsdb.min-block-duration: 2h
      storage.tsdb.retention: 2w
    extraVolumes:
    - name: object-store-volume
      secret:
        secretName: kubecost-thanos
    global:
      external_labels:
        cluster_id: kc-demo-dev
    sidecarContainers:
    - args:
      - sidecar
      - --log.level=debug
      - --tsdb.path=/data/
      - --prometheus.url=http://127.0.0.1:9090
      - --objstore.config-file=/etc/config/object-store.yaml
      - --min-time=-3h
      image: thanosio/thanos:v0.29.0
      name: thanos-sidecar
```

chipzoller:
This doesn't look like it got cherry picked.

cost-analyzer/templates/diagnostics-deployment.yaml
thomasvn: Gotcha! Was there any issue with `.Values.global.thanos.enabled` though?

jessegoodier: yes, it isn't required for agents, so they wouldn't ship diagnostics by default.
https://github.com/kubecost/poc-common-configurations/blob/main/aws/values-amazon-secondary.yaml

PR #2895, opened 2024-01-02
diagnostics logic to allow non-monitored primary

JesseGoodier:


thomasvn:
This capability has been built into the backend, and is configurable via the `DIAGNOSTICS_PRIMARY_READONLY` env var. Currently available in v2.0.x. More docs & helm values coming soon.

https://github.com/kubecost/kubecost-cost-model/pull/2066

PR #2902, opened 2024-01-04
add PSP failure message

chipzoller:

## What does this PR change?

Adds a failure message when PSPs are enabled.

2.0 changes only

Fails the installation/upgrade if PSPs are enabled.

## How was this PR tested?

In a local 1.24 cluster with PSPs enabled as well as a standard upgrade where no PSPs are enabled.


chipzoller:
Note that CI will fail because the state of the chart on this branch requires changes past 1.108 which is what it references.

JesseGoodier (approved):
I like this for two reasons.
1. simplification
2. makes it obvious that there is a need to consider the impact of leaving PSP enabled now that it isn't supported in most current k8s versions.

chipzoller:
Thanks, @jessegoodier. Although this is minor, @thomasvn and @AjayTripathy, are you satisfied with this approach?

thomasvn:
Looks good! This should cover cases where users are deploying Kubecost-related PSPs. What if users are deploying PSPs in the subcharts? Do we need to capture that in the `if` statement here too?

PR #2907, opened 2024-01-04
Fix default Cloud Cost config and make sure new CC container has access to configs from values.yaml

michaelmdresser:


JesseGoodier (approved):
works!

PR #2908, opened 2024-01-04
remove prom configmap reloader

AjayTripathy:
## What does this PR change?
Removes the configmap reloader from prometheus

It's based off of 2.0-helm-changes

Reduce number of images run for simplicity.

Breaking the install step. I have tested as below.

## How was this PR tested?
helm install, then 
```
atripathy@ajay:~$ kubectl get pods -n kubecost | grep prometheus-server
kubecost-prometheus-server-75cc744cb5-zmrpx                 1/1     Running            0            104s
```

No; this should happen under the hood


chipzoller (changes requested):
You must also excise references to the reloader from all other templates as well as any other values files. There are four total templates which reference the ConfigMap reloader and two values files (the other being `values-eks-cost-monitoring`).

chipzoller:
As the Prometheus ConfigMap reloader was a by-default component, we also need to test a simple upgrade from 1.10x to the changes reflected in this PR to ensure presence of those residual values don't cause any unforeseen consequences.

chipzoller:
Bump @AjayTripathy

chipzoller:
Not removing Prometheus ConfigMap reloader but rather disabling by default. See change in #2994.

PR #2918, opened 2024-01-09
allow upgrades from existing aggregator configs

JesseGoodier:
## What does this PR change?
Do we want to remove aggregator-storage pv? if so, it will break upgrades. 

This PR allow upgrades. Though I still need to test singlePod installs.


ameijer:
jesse, when you say 'break upgrades', what do you mean? Someone who upgrades to this will need to re-ingest from scratch? if so, it may be a moot point - we have introduced a database schema versioning change where we start from scratch every time a DB version changes. Duck is not backwards/forwards compatible. This change will land in 2.0, so we will be starting from scratch anyway

jessegoodier:
Spoke with Alex. We are using 2.0 as the chance to simplify the aggregator config. very minor impact to existing users.

PR #2920, opened 2024-01-09
fix aggregator 2.0 storage

JesseGoodier:
## What does this PR change?
2.0 PR was missing 
1. -db- from the PVC mount
2. federated-storage secret.

NA

## How was this PR tested?
helm install using aggregator sts


ameijer (approved):
LGTM

PR #2921, opened 2024-01-09
Only render the k8s Service if Aggregator is enabled

thomasvn:
## What does this PR change?

* For non-primary Federated clusters which do not have Aggregator enabled, do not create a Service for it.

* No

* 

None

* There _should_ be no risk as non-primary clusters should not be querying the aggregator. @michaelmdresser and @ameijer for confirmation?

## How was this PR tested?

```yaml
kubecostAggregator:
  deployMethod: "disabled"
```

Before

```sh
$ helm template ~/kubecost/cost-analyzer-helm-chart/cost-analyzer/ -f values-agent.yaml
Error: execution error at (cost-analyzer/templates/aggregator-service.yaml:12:3): Failed to set aggregator.selectorLabels
```

After

```sh
# No errors, success!
$ helm template ~/kubecost/cost-analyzer-helm-chart/cost-analyzer/ -f values-agent.yaml
```

* Not needed

### DIFF
--- a/cost-analyzer/templates/aggregator-service.yaml
+++ b/cost-analyzer/templates/aggregator-service.yaml
@@ -1,4 +1,5 @@
 {{- if and (not .Values.agent) (not .Values.cloudAgent) }}
+{{- if not (eq .Values.kubecostAggregator.deployMethod "disabled") }}
 
 kind: Service
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


ameijer (approved):
this LGTM. If aggregator is disabled, then having the service be rendered can only cause confusion

PR #2939, opened 2024-01-18
Forecasting and anomaly detection

ghost:
## What does this PR change?
Adds deployment information for forecasting and anomaly detection. For the two env vars it defines, I also want to make sure my understanding that they can be overwritten by a values.yaml is correct, because right now they're defaulted to the same dev values used prior, and we'll want to change that. Also KCM_BASE_URL probably should get a different default.

Doesn't rely on #2890 , but is an extension of it.

Allows enterprise users to enable forecasting and anomaly detection!

It shouldn't affect any other important KCM components, so the main risk is the embarrassment of non-functional forecasting/anomaly detection if those break.

## How was this PR tested?
In one of the QA clusters.

Nope, still need to work with Brett on that


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -102,6 +102,19 @@ data:
 {{- end }}
 
   {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
+  {{- if and (.Values.kubecostDeployment) (.Values.kubecostDeployment.queryServiceReplicas) (gt (.Values.kubecostDeployment.queryServiceReplicas | toString | atoi) 0) }}
+    upstream queryservice {
+        server {{ .Release.Name }}-query-service-load-balancer.{{ .Release.Namespace }}:9003;
+    }
+  {{- end }}
```
thomasvn: The diff shows that you're adding in some lines here referencing `queryServiceReplicas`. Was that intentional? My understanding was that QSR is no more?

cost-analyzer/templates/forecasting-deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+              containerPort: 5000
+              protocol: TCP
+          resources:
+            {{- toYaml .Values.forecasting.resources | nindent 12 }}
+          volumeMounts:
+          env:
+            - name: CONFIG_PATH
+              value: /var/configs/
+            - name: KCM_BASE_URL
+              value: "https://aggregator.nightly.kubecost.xyz/model"
```
thomasvn:
```suggestion
              value: {{ template "cost-analyzer.serviceName" . }}
```

This _should_ work. This will point to the k8s DNS name of the `cost-analyzer` service. You may need to add `:9003/model` as well.

cost-analyzer/templates/forecasting-deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+            - name: MODEL_STORAGE_PATH
+              value: "/tmp/localrun/models"
```
thomasvn, line 73: I assume you're writing your models to disk at this location? How large are these models? If they're sufficiently large we may want to consider creating a PersistentVolume for them.

cost-analyzer/templates/forecasting-service-account.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.forecasting.enabled }}
+{{- if and .Values.serviceAccount.create .Values.forecasting.serviceAccountName }}
+apiVersion: v1
+kind: ServiceAccount
```
thomasvn: I don't think we need this file? @jessegoodier Any reason you created it in the initial draft?

cost-analyzer/values.yaml
```diff
@@ -2235,6 +2235,52 @@ kubecostDeployment:
   labels: {}
   annotations: {}
 
+  ## QueryServiceReplicas
```
thomasvn: Again, not sure if you meant to add back in QueryServiceReplicas here?

thomasvn:
@CampfireRemnants Looking really good! I'm noticing this CI failure though:

```
Error:  templates/: parse error at (cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml:940): unexpected EOF
```

As you make changes to this draft, it may help to continuously run `helm template ./cost-analyzer --set forecasting.enabled=true` to make sure that helm is templating all these variables as you expect it to.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: Yes, this is the right thing to do. As Thomas said, you may need to add a port and route -- _not_ 9003 though! `:9090/model`, or `:9004` if you want to hit aggregator directly. 9003 is the cost-model container.

thomasvn: Oops! Great catch thank you

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
ghost: Nope, unintentional, that was a mistake when I was resolving merge conflicts, thanks for the catch

chipzoller (changes requested):
Let's please not start another new root-level values tree here. Let's start moving towards some semblance of organizing these values into logical structures. Forecasting is a product configuration so how about we start something like an `appConfig` at the root instead?

Also, please ensure all new values are defined (even if commented out) in the main values file. There shouldn't be any new values unaccounted for.

thomasvn:
@chipzoller Agree that root-level values are getting pretty crowded. But since this forecasting feature has its own `Deployment` and `Service` I think an argument can be made that it does belong at the root-level? It wouldn't seem to belong under `kubecostModel` or `kubecostFrontend` since this feature is more than just an environment variable configuration.

chipzoller:
I'm not opposed to putting more things at root level but I think there's a better organizational strategy here. Perhaps some other abstraction layer?

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
michaelmdresser: Addressed

cost-analyzer/templates/forecasting-service-account.yaml
michaelmdresser: Good point Thomas, I removed this. If it causes problems it can be reintroduced.

cost-analyzer/values.yaml
michaelmdresser: Removed, nice catch.

cost-analyzer/values.yaml
```diff
@@ -2235,6 +2235,10 @@ kubecostDeployment:
   labels: {}
   annotations: {}
 
+## Kubecost Forecasting
```
chipzoller: Please provide more description for what this feature is and what it does.

michaelmdresser: Done!

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: They are starting out relatively simple, and we're focusing on functionality first over perfect stability of saved models, so this isn't a major initial concern. This is an excellent general callout, though! We have to follow up here when the models get more sophisticated and the component has been proven out more.

cost-analyzer/templates/forecasting-deployment.yaml
chipzoller, line 1: Needs to incorporate the following common elements from the chart:

* Global Pod annotations
* Global labels
* Global securityContext and containerSecurityContext
* Fixed probes and, preferably, templatization similar to cost-model and frontend

cost-analyzer/templates/forecasting-deployment.yaml
chipzoller: Ditto. Also note because this does not yet adopt global securityContext, writing to `/tmp` may be disallowed with `readOnlyRootFilesystem` set to `true`.

cost-analyzer/templates/forecasting-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: v1
+metadata:
+  name: {{ template "forecasting.serviceName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "forecasting.commonLabels" . | nindent 4 }}
+spec:
+  selector:
+{{ include "forecasting.selectorLabels" . | nindent 4 }}
+  type: "ClusterIP"
```
chipzoller:
```suggestion
  type:ClusterI
```

cost-analyzer/templates/forecasting-service.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.forecasting.enabled }}
+
+kind: Service
+apiVersion: v1
+metadata:
+  name: {{ template "forecasting.serviceName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "forecasting.commonLabels" . | nindent 4 }}
```
chipzoller:
```suggestion
    {{ include "forecasting.commonLabels" . | nindent 4 }}
```

cost-analyzer/templates/forecasting-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  selector:
+{{ include "forecasting.selectorLabels" . | nindent 4 }}
```
chipzoller:
```suggestion
    {{ include "forecasting.selectorLabels" . | nindent 4 }}
```

cost-analyzer/templates/forecasting-service.yaml
michaelmdresser: This is correctly indented when I open the file in my local editor.

cost-analyzer/templates/forecasting-service.yaml
michaelmdresser: This is correctly indented when I open the file in my local editor.

cost-analyzer/templates/forecasting-service.yaml
michaelmdresser: Juust kidding, different file, will fix

cost-analyzer/templates/forecasting-service.yaml
michaelmdresser: Fixed

chipzoller (commented):
This needs to be tested with global level securityContext entries populated for both Pod- and container-level securityContext settings. All Deployments/Pods need to be able to pass the Pod Security Standards (PSS) restricted profile.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: With r/o root fs, I added an empty dir volume to make this work correctly. In the future we'll want to consider the PV question more closely and ideally make the path configurable.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: First three should be done now. The probes are an in-code requirement and out of my control on this PR at the moment.

What do you mean by "templatization similar to cost-model and frontend"?

cost-analyzer/templates/forecasting-deployment.yaml
chipzoller: That works, but there should probably be a size limit set for this. emptyDir's default backing medium is memory. Do we know how large these models are, per Thomas' question?

cost-analyzer/templates/forecasting-deployment.yaml
chipzoller: Liveness and readiness probes are templated for cost-model and frontend, per a more recent change. We should strive for consistency to do the same across all our other Pod controllers.

chipzoller: Ref: https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L549-L560

cost-analyzer/templates/forecasting-deployment.yaml
```diff
@@ -0,0 +1,124 @@
...
+          fix readinessProbe
+
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: 5000
+            initialDelaySeconds: 10
+            periodSeconds: 5
+            failureThreshold: 200
+          */}}
```
chipzoller: Commented out. Need to fix.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: Got it, thank you. I don't believe Ashleigh has the necessary endpoints implemented on this container yet.

michaelmdresser:
Security contexts have been added and this chart is now installed in a live cluster; the container appears to be working correctly.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: @CampfireRemnants can you make flask serve `/healthz` and always return a 200? That's a prerequisite

cost-analyzer/templates/forecasting-deployment.yaml
ghost: Currently, the models are very small json files, generally 1-2 Mb for all of them

cost-analyzer/templates/forecasting-deployment.yaml
ghost: Yes, easily.

cost-analyzer/templates/forecasting-deployment.yaml
ghost: @chipzoller  For the liveness and readiness probes, any endpoints other than just `/healthz` required?

michaelmdresser: Chart updated for the upcoming /healthz -- we will use that for both liveness and readiness Ashleigh, so we just need a 200 from that.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: Probes fixed. Will test once an image with /healthz is available.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: Thanks Chip, I didn't know that -- set sizeLimit on the emptydir to 500Mi to start

michaelmdresser:
@chipzoller I think I've responded to all of the feedback on Ashleigh's PR here. I know we've talked about moving values out of the root in the past; I agree with the general idea but I don't feel like I have a good concept of the structure we should be pursuing, so I haven't made a change there.

chipzoller:
Is this ready for final review or there still outstanding items?

michaelmdresser:
Ready for final review @chipzoller

cost-analyzer/templates/forecasting-deployment.yaml
chipzoller: `/healthz` is pretty standard, but the health check should really not have any external dependencies in order to return a successful code. We don't do that for other containers, but this is one of the best practices.

chipzoller: Resolving this conversation since the current state of this PR does not have probes. My understanding from this discussion is that probes will be implemented in a follow-up PR once the code is incorporated and a new image has been published. If that's not right, please correct me.

cost-analyzer/values.yaml
```diff
@@ -2235,6 +2235,46 @@ kubecostDeployment:
...
+  enabled: false
+
+  # fullImageName overrides the default image construction logic. The exact
+  # image provided (registry, image, tag) will be used for the forecasting
+  # container.
+  # Example: fullImageName: gcr.io/kubecost1/forecasting:v0.0.1
+  fullImageName: gcr.io/kubecost1/kubecost-modeling:6ff46def0575545eb37c5dae179e5035a1de279f
+
+  # Resource specification block for the forecasting container.
+  resources: {}
```
chipzoller: Are there truly no resources we can offer as defaults here?

cost-analyzer/values.yaml
```diff
@@ -2235,6 +2235,46 @@ kubecostDeployment:
...
+## Kubecost Forecasting forecasts future cost patterns based on historical
+## patterns observed by Kubecost.
+forecasting:
+  enabled: false
+
+  # fullImageName overrides the default image construction logic. The exact
+  # image provided (registry, image, tag) will be used for the forecasting
+  # container.
+  # Example: fullImageName: gcr.io/kubecost1/forecasting:v0.0.1
+  fullImageName: gcr.io/kubecost1/kubecost-modeling:6ff46def0575545eb37c5dae179e5035a1de279f
```
chipzoller: What is this tag?

cost-analyzer/templates/forecasting-deployment.yaml
```diff
@@ -0,0 +1,134 @@
...
+    {{- with .Values.global.additionalLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{ include "forecasting.selectorLabels" . | nindent 6 }}
+  strategy:
+    type: Recreate
```
chipzoller: cost-analyzer defaults to `RollingUpdate`. Are we sure we want `Recreate` here?

chipzoller (commented):
See comments thus far. Please also confirm this Deployment does not need a ServiceAccount of any sort. If it does not, we should disable the automounting of one (`spec.template.spec.automountServiceAccountToken: false`).

cost-analyzer/values.yaml
michaelmdresser: A good point. I took a stab at some, let me know if they seem reasonable

cost-analyzer/values.yaml
michaelmdresser: We plan to release forecasting like Cluster Controller and Network Costs, with a fixed image in the Helm chart (instead of the branching/tagging used for higher-traffic repos like KCM). The particular images being set here are development images that are very close to the final release state.

cost-analyzer/templates/forecasting-deployment.yaml
michaelmdresser: Switched to RollingUpdate for consistency with other Deployments, I'm not sure why Recreate was chosen initially.

michaelmdresser:
Responded to all -- AFAIK we don't need a service account (runs fine without) so that is gone and I set the value you suggested be set in that case.

michaelmdresser (approved):

cliffcolvin:
I discussed with Ajay and pinged Chip. I'll setup time to discuss the action items after this and get tasks entered for us to do the work on moving things out of root. Squash+Merging now

PR #2940, opened 2024-01-18
Collections caching

nikovacevic:
## What does this PR change?
Adds caching for collections query endpoints

* https://github.com/kubecost/kubecost-core/pull/7
* https://github.com/kubecost/kubecost-cost-model/pull/2058

## How was this PR tested?
See https://github.com/kubecost/kubecost-cost-model/pull/2058

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -905,6 +905,8 @@ spec:
             - name: ADVANCED_NETWORK_STATS
               value: "false"
             {{- end}}
+            - name: COLLECTIONS_MEMORY_CACHE_ENABLED
+              value: {{ (quote .Values.kubecostModel.collections.cache.enabled) | default (quote true) }}
             {{- end }}
             {{- end }}
             {{- end }}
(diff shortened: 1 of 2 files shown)


cliffcolvin (approved):
Lgtm

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -905,6 +905,8 @@ spec:
             - name: ADVANCED_NETWORK_STATS
               value: "false"
             {{- end}}
+            - name: COLLECTIONS_MEMORY_CACHE_ENABLED
+              value: {{ (quote .Values.kubecostModel.collections.cache.enabled) | default (quote true) }}
```
chipzoller, line 909: I don't see this present in the main values file. Did I miss it?

chipzoller:
Further, this PR was merged when linting failed which revealed a null pointer because the value was not defined. Why?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
nikovacevic: No, the idea is that it doesn't need to be there, but maybe that's worng.

chipzoller: We still need to have it accounted for in the values file. The values file should be the comprehensive dictionary of all settable values, even if it isn't technically required. Can you please address in a follow-up PR?

PR #2941, opened 2024-01-19
Fix nesting of collection cache enabling flag

nikovacevic:
## What does this PR change?
Fixes some bad nesting for a new env var for collections.

No

## How was this PR tested?
Manually

### Without the values, the default works, and is "on"
Screenshot from 2024-01-19 11-21-10
Screenshot from 2024-01-19 11-22-03

### Without the values, the cache can be turned off
Screenshot from 2024-01-19 11-16-47
Screenshot from 2024-01-19 11-17-16

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -905,8 +905,6 @@ spec:
             - name: ADVANCED_NETWORK_STATS
               value: "false"
             {{- end}}
-            - name: COLLECTIONS_MEMORY_CACHE_ENABLED
-              value: {{ (quote .Values.kubecostModel.collections.cache.enabled) | default (quote true) }}
             {{- end }}
             {{- end }}
             {{- end }}
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -812,6 +812,14 @@ Aggregator config reconciliation and common config
     - name: OIDC_SKIP_ONLINE_VALIDATION
       value: {{ (quote .Values.oidc.skipOnlineTokenValidation) | default (quote false) }}
     {{- end}}
+    {{- if .Values.kubecostAggregator }}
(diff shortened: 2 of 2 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -905,6 +905,12 @@ spec:
             - name: ADVANCED_NETWORK_STATS
               value: "false"
             {{- end}}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- if .Values.kubecostModel }}
+            {{- if .Values.kubecostModel.collections }}
+            {{- if .Values.kubecostModel.collections.cache }}
```
chipzoller:
```suggestion
            {{- if (((.Values.kubecostModel).collections).cache) }}
```

mbolt35 (approved):
LGTM

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
nikovacevic: Seems to work without the parens. Are they required? Just following existing patterns.

chipzoller: Yes it works but this is a more efficient pattern. Makes it simpler.

nikovacevic: Added the parens, still works.

mbolt35: > Yes it works but this is a more efficient pattern. Makes it simpler.

I can say with 100% certainty that this is not conventional go templating and there is 0 added efficiency from adding parens like this. There's no ambiguous access anywhere in this chain.

chipzoller: This is common Helm syntax for nested `if` statements which reduces the necessary templating by two thirds. Several instances have been written already and we'll convert more in this chart over time.

PR #2945, opened 2024-01-22
Change chart on `develop` branch to use nightly images

chipzoller:

## What does this PR change?

Modifies the state of the Helm chart on the `develop` branch so the `appVersion` is listed as `development`, causing deployment of the chart (again, only intended from the `develop` branch) pulls the nightly images for `cost-model` and `frontend`. This change is required because chart testing would always fail if modifications to the chart depended on modifications only found in these nightly images. This problem was apparent even before the development work on 2.0 began. In order for CI to work properly on the `develop` branch, this change is required.

**Note:** This change will likely require some modifications to the release process in that new minor versions, after the release branch is cut from `develop` MUST modify the `appVersion` field in `Chart.yaml` to point to the then-released version.

No

No impact to end users as this change only affects the state on the `develop` branch which no users deploy.

N/A

As noted above, these changes probably necessitate changes to the release process and potentially other areas. I'm relying on those tagged for review here to understand those requirements and make the necessary changes.

## How was this PR tested?

Tested locally and in CI. Note that CI is currently failing on when it comes to deploying the Replicated clusters. These steps will be removed in a forthcoming PR. The tests are otherwise green for local KinD clusters.

N/A

### DIFF
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "1.108.0"
+appVersion: "development"
 description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to monitor cloud costs.
 name: cost-analyzer
-version: "1.108.0"
+version: "2.0.0"
 annotations:
   "artifacthub.io/links": |
     - name: Homepage
--- a/cost-analyzer/templates/etl-utils-deployment.yaml
+++ b/cost-analyzer/templates/etl-utils-deployment.yaml
@@ -47,8 +47,8 @@ spec:
           image: {{ .Values.kubecostModel.fullImageName }}
           {{- else if .Values.imageVersion }}
           image: {{ .Values.kubecostModel.image }}:{{ .Values.imageVersion }}
-        {{- else if eq "development" .Chart.AppVersion }}
-        - image: gcr.io/kubecost1/cost-model-nightly:latest
+          {{- else if eq "development" .Chart.AppVersion }}
+          image: gcr.io/kubecost1/cost-model-nightly:latest
           {{- else }}
           image: {{ .Values.kubecostModel.image }}:prod-{{ $.Chart.AppVersion }}
           {{ end }}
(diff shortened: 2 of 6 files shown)


cost-analyzer/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "1.108.0"
+appVersion: "development"
```
ameijer, line 2: What I like about this change is that it defaults to the nightly images when you are installing the chart from your filesystem, which will help when we are doing local chart dev. In contrast, releases are always built via CI so it is possible to have logic there that alters this and feel confident about the coverage there

ameijer (commented):
chip this LGTM but going to let someone more involved with release scripts give the final approval

cost-analyzer/Chart.yaml
chipzoller: Yes, that is my suggestion for modifying the automated release process.

chipzoller: Looks like, based on past PRs, no modifications may be needed (but I'll leave that to you all). The release automation is bumping the two version related fields in the release branch.

cost-analyzer/templates/etl-utils-deployment.yaml
```diff
@@ -47,6 +47,8 @@ spec:
           image: {{ .Values.kubecostModel.fullImageName }}
           {{- else if .Values.imageVersion }}
           image: {{ .Values.kubecostModel.image }}:{{ .Values.imageVersion }}
+        {{- else if eq "development" .Chart.AppVersion }}
+        - image: gcr.io/kubecost1/cost-model-nightly:latest
```
michaelmdresser: I think this might be in error
```suggestion
        {{- else if eq "development" .Chart.AppVersion }}
          image: gcr.io/kubecost1/cost-model-nightly:latest
```

michaelmdresser (commented):
Release scripts should be fine, they already modify appVersion: https://github.com/kubecost/release-scripts/blob/80092b589a0b9967476d37f96ec2a8fa9ef9dc4d/create_release_tags.py#L56

cost-analyzer/templates/etl-utils-deployment.yaml
chipzoller: Thanks very much. Fixed.

thomasvn (approved):

ameijer (approved):

michaelmdresser (approved):

cliffcolvin (approved):
Thank you Chip, looks good to me

PR #2954, opened 2024-01-23
Unset DB memory limit

michaelmdresser:
## What does this PR change?
This is a little too aggressive (it can cause OOM on nightly size data), and we've noticed that DDB obeys the K8s limits better than this limit (for some reason).

Removed the default env var `DB_MEMORY_LIMIT` setting for `kubecostAggregator`. If limiting the Aggregator memory, use the K8s limit first.


ameijer (approved):
⚡

PR #2961, opened 2024-01-24
[Proposal] Users must set flag `upgrade.toV2=true` if upgrading from Kubecostv1

thomasvn:
## What does this PR change?

- Users must set flag `upgrade.toV2=true` if upgrading from Kubecostv1
- Intended to act as a safeguard in environments that frequently/continuously upgrade their deployments. Need to ensure that they're aware of the architectural requirements for Kubecost 2.0.
- Open to feedback on the idea, the wording, or the flag.

No

Flag required to be set when upgrading to KubecostV2.

None

Risk is creating added friction to the user experience when going to Kubecost 2.0.

## How was this PR tested?

```yaml
# Chart.yaml
version: "1.108.1"
```

```sh
# Error thrown
$ helm template ./cost-analyzer
Error: execution error at (cost-analyzer/templates/NOTES.txt:4:4):

You are attempting to upgrade to Kubecost 2.0. Please refer to the following documentation and talk to your Kubecost representative before upgrading, as there are potential breaking changes:
https://docs.kubecost.com/install-and-configure/install/kubecostv2

When ready to upgrade, set `.Values.upgrade.toV2=true`.

Use --debug flag to render out invalid YAML
```

```sh
# Successfully upgrade
$ helm template ./cost-analyzer --set upgrade.toV2=true
```

TODO: Need to generate documentation for users upgrading up to Kubecost 2.0. (e.g. Thanos users, FedETL users, single-cluster users).

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -2822,6 +2822,12 @@ kubecostAdmissionController:
 costEventsAudit:
   enabled: false
 
+## This flag is only required for users upgrading from 1.x to 2.x. Ensure users
+## are aware of any changes that might need to be made to their architecture
+## before continuing with the upgrade.
+## 
+upgradeToKubecostV2: false
+
 ## Disable updates to kubecost from the frontend UI and via POST request
 ##
 # readonly: false
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -2825,7 +2825,7 @@ costEventsAudit:
 ## This flag is only required for users upgrading from 1.x to 2.x. Ensure users
 ## are aware of any changes that might need to be made to their architecture
 ## before continuing with the upgrade.
-## 
+##
 upgradeToKubecostV2: false
 
 ## Disable updates to kubecost from the frontend UI and via POST request
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -2822,11 +2822,12 @@ kubecostAdmissionController:
 costEventsAudit:
   enabled: false
 
-## This flag is only required for users upgrading from 1.x to 2.x. Ensure users
-## are aware of any changes that might need to be made to their architecture
-## before continuing with the upgrade.
+## This flag is only required for users upgrading to a new version of Kubecost
+## which require a flag set. The flag is used to ensure users are aware of
+## important (potentially breaking) changes included in the new version.
 ##
-upgradeToKubecostV2: false
+upgrade:
+  toV2: false
 
 ## Disable updates to kubecost from the frontend UI and via POST request
 ##
--- a/cost-analyzer/templates/NOTES.txt
+++ b/cost-analyzer/templates/NOTES.txt
@@ -1,6 +1,8 @@
 
 
 --------------------------------------------------
+{{- include "kubecostV2-preconditions" . }}
+
 {{/*
 https://github.com/helm/helm/issues/8026#issuecomment-881216078
 */}}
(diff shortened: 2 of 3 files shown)


cost-analyzer/values.yaml
```diff
@@ -2822,6 +2822,12 @@ kubecostAdmissionController:
 costEventsAudit:
   enabled: false
 
+## This flag is only required for users upgrading from 1.x to 2.x. Ensure users
+## are aware of any changes that might need to be made to their architecture
+## before continuing with the upgrade.
+##
+upgradeToKubecostV2: false
```
chipzoller: I suggest making this a map so we can repurpose the struct for future major versions if the need should arise (and, let's be honest, it probably will):

```yaml
upgrade:
  fromV1: true
  (fromV2: true)
```

thomasvn: Great idea, doing that now!

thomasvn: @chipzoller What do you think about this modification

```yaml
upgradeKubecost:
  toV2: true
```

chipzoller: Because this chart is not called "Kubecost" but rather "Cost Analyzer" I think this creates a small amount of confusion. Should we implement the complimentary changes that have been proposed but not accepted for the [2.0 Helm chart](https://docs.google.com/document/d/17GJkIS2ce5w0u9qmaBJYIi99NTs_dEs6Iid83aOZw2w/) (Kubecost internal document), this works perfectly. Otherwise, I would drop the `Kubecost` portion of the key but that is a minor detail.

thomasvn: Right, I see what you're saying. Will do!

teevans:
@dwbrown2 - Can you weigh in here? 

We're wanting to safeguard the upgrade - because we're reducing the number of PVs, Kubernetes will not let us remove a PV from a stateful set - so we have to tell the user to delete the stateful set, then install.

dwbrown2:
Just want to make sure I really understand the impact of this one... 

First questions: 
* any specific scenarios that are driving the recommendation? 
* what would be the impact to EKS Optimized upgrades?

teevans:
So - Specifically, if anyone is currently running aggregator on versions 1.107+, the users will need to delete the stateful set that currently exists prior to upgrading to 2.0. 

For EKS - I'm actually not sure - @thomasvn / @chipzoller can you weigh in?

thomasvn:
Sorry for any miscommunication here! I've got a few similar ideas in flight:
* (not in this PR) When using Aggregator in v1.107 or v1.108 then upgrading to v2.0, Helm fails due to a templating issue. This requires users to manually delete a resource on the cluster, then run the upgrade again. To get ahead of this issue, we should generate our own error message when detecting Aggregator in v1.107 or v1.108, then provide helpful instructions on how to resolve the issue.
* (in this PR) In a scenario where users are continuously upgrading their Kubecost environment but are unaware of the 2.0 release and what it entails, we want to generate this warning message stopping them from the install and providing some documentation. We want to clearly document what needs to be done in a Thanos/FedETL/singlecluster environment before going up to 2.0. Then when they're ready to upgrade they can set the flag `upgrade.toV2=true` and monitor the situation. The tradeoff here is extra info/awareness in exchange for some added upgrade friction.

Regarding the question on EKS Optimized upgrades, I don't think there's any impact. However, I don't know what the effect of this would be, if the user installed Kubecost via a marketplace instead of Helm.

thomasvn:
Closing this proposal. Instead of requiring users to set `.Values.upgrade.toV2=true`, we will build in tests to the Helm Chart which throw errors if users are running incompatible architectures. Examples [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/f34a9e00ce713480369fb1eb597f1d3d019d5c5d/cost-analyzer/templates/_helpers.tpl#L24-L64).

PR #2965, opened 2024-01-25
disable node exporter by default

JesseGoodier:
## What does this PR change?
disable node exporter by default

no

Disable node exporter by default, it can be reenabled as needed.

Reserved instances will display a message that node-exporter is required, but this report is only useful for single cluster users.

## How was this PR tested?
helm installs

@bstuder99 can you look for our node exporter docs and add a message about the default?

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1514,9 +1514,10 @@ prometheus:
 
   # node-export must be disabled if there is an existing daemonset: https://guide.kubecost.com/hc/en-us/articles/4407601830679-Troubleshoot-Install#a-name-node-exporter-a-issue-failedscheduling-kubecost-prometheus-node-exporter
   nodeExporter:
-    ## If false, node-exporter will not be installed
+    ## If false, node-exporter will not be installed.
+    ## This is disabled by default in Kubecost 2.0, though it can be enabled as needed.
     ##
-    enabled: true
+    enabled: false
 
     ## If true, node-exporter pods share the host network namespace
     ##


AjayTripathy (approved):

brstuder:
@jessegoodier Our existing docs only mention node exporter in the context of disabling it on secondary clusters, will it now be automatically disabled on all primary/secondary clusters? Also, the sample Helm config used in [this article](https://docs.kubecost.com/install-and-configure/install/secondary-clusters#node-exporter) does not seem to actually enable/disable the node exporter because of the #, is this supposed to be correct?

```
kubecostProductConfigs:
  clusterName: kubecostProductConfigs_clusterName
  # productKey not needed on secondary clusters
kubecostModel:
  warmCache: false
  warmSavingsCache: false
  etl: false
  etlCloudAsset: false
  maxQueryConcurrency: 1
global:
  grafana:
    enabled: false
    proxy: false
prometheus:
  server:
    global:
      external_labels:
        # cluster_id should be unique for all clusters and the same value as .kubecostProductConfigs.clusterName
        cluster_id: kubecostProductConfigs_clusterName
    retention: 2d
  # nodeExporter:
  #   enabled: false
  # serviceAccounts:
  #   nodeExporter:
  #     create: false
thanos:
  compact:
    enabled: false
  bucket:
    enabled: false
  query:
    enabled: false
  queryFrontend:
    enabled: false
  store:
    enabled: false
   ```

jessegoodier:
> @jessegoodier Our existing docs only mention node exporter in the context of disabling it on secondary clusters, will it now be automatically disabled on all primary/secondary clusters? Also, the sample Helm config used in [this article](https://docs.kubecost.com/install-and-configure/install/secondary-clusters#node-exporter) does not seem to actually enable/disable the node exporter because of the #, is this supposed to be correct?
> 
> ```
> kubecostProductConfigs:
>   clusterName: kubecostProductConfigs_clusterName
>   # productKey not needed on secondary clusters
> kubecostModel:
>   warmCache: false
>   warmSavingsCache: false
>   etl: false
>   etlCloudAsset: false
>   maxQueryConcurrency: 1
> global:
>   grafana:
>     enabled: false
>     proxy: false
> prometheus:
>   server:
>     global:
>       external_labels:
>         # cluster_id should be unique for all clusters and the same value as .kubecostProductConfigs.clusterName
>         cluster_id: kubecostProductConfigs_clusterName
>     retention: 2d
>   # nodeExporter:
>   #   enabled: false
>   # serviceAccounts:
>   #   nodeExporter:
>   #     create: false
> thanos:
>   compact:
>     enabled: false
>   bucket:
>     enabled: false
>   query:
>     enabled: false
>   queryFrontend:
>     enabled: false
>   store:
>     enabled: false
> ```

Thanks Brett. If we could just update the https://docs.kubecost.com/install-and-configure/install/secondary-clusters#node-exporter verbiage, I think we're good.

PR #2976, opened 2024-01-26
Sensible error on /forecasting if forecasting is disabled (instead of index.html)

michaelmdresser:
## What does this PR change?
Causes all `/forecasting` routes to 404 if forecasting is disabled. This means we shouldn't hit the fallback to index.html, which was odd behavior because it caused a 200 code response but with the default index.html, not the expected /forecasting/... JSON response.

N/A, unreleased

## How was this PR tested?
```
→ helm template kubecost ./cost-analyzer --set forecasting.enabled='false' -s 'templates/cost-analyzer-frontend-config-map-template.yaml' | yq '.data["nginx.conf"]' | grep -B 1 -A 5 'forecasting'
    }
    location /forecasting {
        return 404 '{"forecastingEnabled": "false"}';
    }

}
```
```
→ helm template kubecost ./cost-analyzer --set forecasting.enabled='true' -s 'templates/cost-analyzer-frontend-config-map-template.yaml' | yq '.data["nginx.conf"]' | grep -B 1 -A 5 'forecasting'
}
upstream forecasting {
    server kubecost-forecasting.default:5000;
}
upstream aggregator {
    server kubecost-aggregator.default:9004;
}

--
    }
    location /forecasting {
        add_header 'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;
        proxy_read_timeout          300;
        proxy_pass http://forecasting/;
        proxy_redirect off;
        proxy_set_header Connection "";
        proxy_set_header  X-Real-IP  $remote_addr;
    }
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -936,7 +936,7 @@ data:
         {{- else }}
         location /forecasting {
             default_type 'application/json';
-            return 404 '{"forecastingEnabled": "false"}';
+            return 405 '{"forecastingEnabled": "false"}';
         }
         {{- end }}
 
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -933,7 +933,12 @@ data:
             proxy_set_header  X-Real-IP  $remote_addr;
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
-        {{ end }}
+        {{- else }}
+        location /forecasting {
+            default_type 'application/json';
+            return 404 '{"forecastingEnabled": "false"}';
+        }
+        {{- end }}
 
     }
 {{- end }}


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -933,7 +933,12 @@ data:
             proxy_set_header  X-Real-IP  $remote_addr;
             proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
         }
-        {{ end }}
+        {{- else }}
+        location /forecasting {
+            default_type 'application/json';
+            return 404 '{"forecastingEnabled": "false"}';
```
jessegoodier: how about 501, not implemented? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

michaelmdresser: Maybe 405? That link says
> If the server does recognize the method, but intentionally does not support it, the appropriate response is [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405).

jessegoodier: actually, https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204
not an error, but no content.

JesseGoodier (approved):

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
nealormsbee: Hey y'all. Sorry I'm so late to the party here 😅 

I have opinions, rationale below.

`204 NO CONTENT` is typically used to confirm the success of a write-style operation that does not return meaningful feedback data. Example: this status code is often used in response to DELETE-ing an entity in a restful API.
It would feel strange to me to return a 2XX response on a GET request an endpoint that does not and cannot return data. But then again....it's like "endpoint was reached, and there's no content", so it's not _totally_ crazy.

I believe `405 METHOD NOT ALLOWED` generally refers to HTTP verbs (GET POST PUT PATCH DELETE). Trying to POST to a read-only endpoint results in a 405 in most webserver frameworks, for example. You could argue that we have an endpoint that allows _no methods_, but that feels to me like it should just be `501 NOT IMPLEMENTED`.

The other thing to help make a determination here is that `2XX` statues are "success", `4XX` statuses are "client issue" and `5XX` statuses are "server issue". In a strict sense, I think a 501 is correct here, because the server is telling the client "that capability is Not Implemented (given the current configuration with which the server is deployed)", paren text mine. There's also `503 Service Unavailable`, but that implies a temporary state of affairs (ex. pod tipped over).

So yeah, I think 501 is most http-semantics correct, but 405 will work just fine given that we really just care about getting a >400 (error) response.

PR #2981, opened 2024-01-26
rbac upgrade check

JesseGoodier:
## What does this PR change?
Force RBAC users to accept risk before upgrading. 

Message:
```
Error: UPGRADE FAILED: execution error at (cost-analyzer/templates/NOTES.txt:4:4): Kubecost 2.x has significant architectural changes that may impact RBAC.
This should be tested before giving end-users access to the UI.
Kubecost has tested various configurations and believe that 2.x will be 100% compatible with existing configurations.
Refer to the following documentation for more information: https://docs.kubecost.com/install-and-configure/install/kubecostv2

When ready to upgrade, add `--set upgrade.toV2=true`.
```

No

RBAC users: Kubecost 2.x has significant architectural changes that may impact RBAC. Please test the configuration before allowing end-user access to Kubecost. 

NA

## How was this PR tested?
Tested against a 1.x SAML environment. 

NA

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -40,7 +40,7 @@ Kubecost 2.0 preconditions
             {{- if gt (len $chartNameAndVersion) 2 -}}
               {{- $chartVersion := $chartNameAndVersion._2 -}}        {{/* 1.108.1 */}}
               {{- if semverCompare "<2.0.0-0" $chartVersion -}}
-                {{- fail "Detected an existing Aggregator StatefulSet in your namespace. Before upgrading to Kubecost 2.0, please `kubectl delete` this Statefulset. Refer to the following documentation for more information: https://docs.kubecost.com/install-and-configure/install/kubecostv2" -}}
+                {{- fail "\n\nAn existing Aggregator StatefulSet was found in your namespace.\nBefore upgrading to Kubecost 2.x, please `kubectl delete` this Statefulset.\nRefer to the following documentation for more information: https://docs.kubecost.com/install-and-configure/install/kubecostv2" -}}
               {{- end -}}
             {{- end -}}
           {{- end -}}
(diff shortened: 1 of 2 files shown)


cost-analyzer/templates/_helpers.tpl
```diff
@@ -51,15 +51,21 @@ Kubecost 2.0 preconditions
...
-    {{- fail "In Kubecost 2.0, there is no such thing as a federated primary. If you are a Federated ETL user, this setting has been removed. Make sure you have kubecostAggregator.deployMethod set to 'statefulset' and federatedETL.federatedCluster set to 'true'." -}}
+    {{- fail "\n\nIn Kubecost 2.0, there is no such thing as a federated primary. If you are a Federated ETL user, this setting has been removed. Make sure you have kubecostAggregator.deployMethod set to 'statefulset' and federatedETL.federatedCluster set to 'true'." -}}
+  {{- end -}}
+
+  {{- if or (.Values.saml).enabled (.Values.oidc).enabled -}}
+    {{- if (not (.Values.upgrade).toV2) -}}
+      {{- fail "\n\nSSO is enabled.\nNote that Kubecost 2.x has significant architectural changes that may impact RBAC.\nThis should be tested before giving end-users access to the UI.\nKubecost has tested various configurations and believe that 2.x will be 100% compatible with existing configurations.\nRefer to the following documentation for more information: https://docs.kubecost.com/install-and-configure/install/kubecostv2\n\nWhen ready to upgrade, add `--set upgrade.toV2=true`." -}}
```
thomasvn: Should we only deliver this message if RBAC is enabled?

```
.Values.saml.rbac.enabled=true
.Values.oidc.rbac.enabled=true
```

cost-analyzer/templates/_helpers.tpl
```diff
@@ -51,15 +51,21 @@ Kubecost 2.0 preconditions
...
-  {{- fail "Kubecost no longer includes Thanos by default. Please see https://docs.kubecost.com/install-and-configure/install/thanos for more information." -}}
+  {{- fail "\n\nKubecost no longer includes Thanos by default. Please see https://docs.kubecost.com/install-and-configure/install/thanos for more information." -}}
   {{- end -}}
 
   {{- if (.Values.federatedETL).primaryCluster -}}
-    {{- fail "In Kubecost 2.0, there is no such thing as a federated primary. If you are a Federated ETL user, this setting has been removed. Make sure you have kubecostAggregator.deployMethod set to 'statefulset' and federatedETL.federatedCluster set to 'true'." -}}
+    {{- fail "\n\nIn Kubecost 2.0, there is no such thing as a federated primary. If you are a Federated ETL user, this setting has been removed. Make sure you have kubecostAggregator.deployMethod set to 'statefulset' and federatedETL.federatedCluster set to 'true'." -}}
+  {{- end -}}
+
+  {{- if or (.Values.saml.rbac).enabled (.Values.oidc.rbac).enabled -}}
```
thomasvn:
```suggestion
  {{- if or ((.Values.saml).rbac).enabled ((.Values.oidc).rbac).enabled -}}
```

I don't think the `()` are needed in this case, because all these values are defined in the values.yaml file. But if we do want to perform these safety checks, I _think_ we have to modify slightly

thomasvn (approved):

PR #2994, opened 2024-01-27
disable configmapReloader by default

JesseGoodier:
## What does this PR change?
Disable Prometheus configMap reloader by default

No

Disable Prometheus configMap reloader by default to reduce containers/images. There is no impact to typical Kubecost installs

Users that don't see the release notes and utilize the functionality provided by the bundle prometheus server.

## How was this PR tested?
Helm installs on several environments.

TBD

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1443,7 +1443,7 @@ prometheus:
     prometheus:
       ## If false, the configmap-reload container will not be deployed
       ##
-      enabled: true
+      enabled: false
 
       ## configmap-reload container name
       ##


chipzoller:
Can we not just yank it out entirely or do you feel that's not safe at this point? Leaving it disabled is still one more image we have to maintain.

jessegoodier:
> Can we not just yank it out entirely or do you feel that's not safe at this point? Leaving it disabled is still one more image we have to maintain.

I'd like to give it a version or two. much easier to rollback this way.

chipzoller:
As discussed, can we also disable the Grafana sidecar as part of this?

jessegoodier:
> As discussed, can we also disable the Grafana sidecar as part of this?

That one needs more work. dashboards don't load without it. I would need to research how to make that happen.

My biggest priority right now is agent config in 2.0

chipzoller (approved):

PR #3048, opened 2024-01-31
Add sanity check for cloudIntegrationSecret

### LINKED ISSUE #3046: [Bug] When setting cloudIntegrationSecret, if it is not exactly `cloud-integration` some containers fail to start
### Kubecost Helm Chart Version

2.0.0

### Kubernetes Version

1.24

### Kubernetes Platform

Bare metal

### Description

I was testing some changes and set `kubecostProductConfigs.cloudIntegrationSecret` to a value other than `cloud-integration`, the cost analyzer pod failed to start.

### Steps to reproduce

1. set `kubecostProductConfigs.cloudIntegrationSecret` to a value other than `cloud-integration`,
2. fail

### Expected behavior

Not fail.

### Impact

Nothing at the moment, since I renamed the secret `cloud-integration`

### Screenshots

_No response_

### Logs

_No response_

### Slack discussion

_No response_

### Troubleshooting

- [X] I have read and followed the [issue guidelines](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/ISSUE_GUIDELINES.md) and this is a bug impacting only the Helm chart.

chipzoller:

## What does this PR change?

Adds a sanity check that the cloudIntegrationSecret, when defined, both exists and has a key with the name `cloud-integration.json` or else fails the chart installation. This is done to prevent users from naming Secrets which don't yet exist and ensure that those Secrets have a key the chart expects. Without both, Kubecost will not be able to come into a healthy, running state.

This also fixes a bug with the name of the volume holding said cloudIntegration secret enabling secrets with a custom name to be properly mounted.

No

Provides a better user experience in that the chart fails and provides feedback to users rather than succeeding in installation only to fail when running.

Fixes #3046

The check could be wrong or it could somehow impact installations where the cloudIntegrationSecret was not specified. It could potentially impact users who use tools or generate manifests from `helm template` although this behavior was tested locally and in CI.

## How was this PR tested?

Various Helm templating and installation scenarios. This check should only be invoked when a "real" connection to a K8s cluster exists and not when using `helm template` with or without `--dry-run=server`.

### Attempting to install when the Secret does not exist

```sh
$ yq e .kubecostProductConfigs.cloudIntegrationSecret cost-analyzer/values.yaml 
mycustomsecret
```

```sh
$ k -n default get secret
No resources found in default namespace.
```

```sh
$ helm install kubecost -n default cost-analyzer/
Error: execution error at (cost-analyzer/templates/NOTES.txt:4:4): The cloud integration secret 'mycustomsecret' does not exist or does not contain the expected key 'cloud-integration.json'
```

### Attempting to install when the Secret exists but the cloud-integration.json key does not

```sh
$ yq e .kubecostProductConfigs.cloudIntegrationSecret cost-analyzer/values.yaml 
mycustomsecret
```

```sh
$ k -n default get secret mycustomsecret -o yaml | yq .data
foo: YmFy
```

```sh
$ helm install kubecost -n default cost-analyzer/
Error: execution error at (cost-analyzer/templates/NOTES.txt:4:4): The cloud integration secret 'mycustomsecret' does not exist or does not contain the expected key 'cloud-integration.json'
```

### Attempting to install when the Secret exists and does contain the key cloud-integration.json

```sh
$ yq e .kubecostProductConfigs.cloudIntegrationSecret cost-analyzer/values.yaml 
mycustomsecret
```

```sh
$ k -n default get secret mycustomsecret -o yaml | yq .data
cloud-integration.json: YmFy
```

```sh
$ helm install kubecost -n default cost-analyzer/
NAME: kubecost
LAST DEPLOYED: Thu Feb  1 10:28:03 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:
--------------------------------------------------
Kubecost 2.0.0 has been successfully installed.

Welcome to Kubecost 2.0!

Kubecost 2.0 is a major upgrade from previous versions and includes major new features including a brand new API Backend. Please review the following documentation to ensure a smooth transition: https://docs.kubecost.com/install-and-configure/install/kubecostv2

For the full list of enhancements, please see our release notes: https://github.com/kubecost/cost-analyzer-helm-chart/releases/tag/v2.0.0

When pods are Ready, you can enable port-forwarding with the following command:

    kubectl port-forward --namespace default deployment/kubecost-cost-analyzer 9090

Then, navigate to http://localhost:9090 in a web browser.

Please allow 25 minutes for Kubecost to gather metrics. A progress indicator will appear at the top of the UI.

Having installation issues? View our Troubleshooting Guide at http://docs.kubecost.com/troubleshoot-install
```

N/A

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -115,7 +115,7 @@ Verify the cloud integration secret exists with the expected key when cloud inte
 {{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
   {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.kubecostProductConfigs.cloudIntegrationSecret }}
   {{- if or (not $secret) (not (index $secret.data "cloud-integration.json")) }}
-    {{- fail "The cloud integration secret does not exist or does not contain the expected key 'cloud-integration.json'" }}
+    {{- fail (printf "The cloud integration secret '%s' does not exist or does not contain the expected key 'cloud-integration.json'" .Values.kubecostProductConfigs.cloudIntegrationSecret) }}
   {{- end }}
 {{- end -}}
 {{- end -}}
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1004,7 +1004,7 @@ Begin Kubecost 2.0 templates
       readOnly: true
   {{- end }}
   {{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
-    - name: {{ .Values.kubecostProductConfigs.cloudIntegrationSecret }}
+    - name: cloud-integration
       mountPath: /var/configs/cloud-integration
   {{- end }}
   env:
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -109,6 +109,18 @@ ERROR: MISSING EBS-CSI DRIVER WHICH IS REQUIRED ON EKS v1.23+ TO MANAGE PERSISTE
 {{- end -}}
 
 {{/*
+Verify the cloud integration secret exists with the expected key when cloud integration is enabled.
+*/}}
+{{- define "cloudIntegrationSecretCheck" -}}
+{{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.kubecostProductConfigs.cloudIntegrationSecret }}
+  {{- if or (not $secret) (not (index $secret.data "cloud-integration.json")) }}
+    {{- fail "The cloud integration secret does not exist or does not contain the expected key 'cloud-integration.json'" }}
+  {{- end }}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Expand the name of the chart.
 */}}
 {{- define "cost-analyzer.name" -}}
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -113,10 +113,12 @@ Verify the cloud integration secret exists with the expected key when cloud inte
 */}}
 {{- define "cloudIntegrationSecretCheck" -}}
 {{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
+{{-  if .Capabilities.APIVersions.Has "v1/Secret" }}
   {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.kubecostProductConfigs.cloudIntegrationSecret }}
   {{- if or (not $secret) (not (index $secret.data "cloud-integration.json")) }}
     {{- fail (printf "The cloud integration secret '%s' does not exist or does not contain the expected key 'cloud-integration.json'" .Values.kubecostProductConfigs.cloudIntegrationSecret) }}
   {{- end }}
+{{- end -}}
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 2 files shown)


thomasvn (approved):
Looks great, thanks for the UX improvement here!

thomasvn:
In order to get CI working, should we remove the cloud integration secret from the following files? https://github.com/kubecost/cost-analyzer-helm-chart/tree/develop/cost-analyzer/ci

chipzoller:
The problem here is that the templatization will always fail if a cluster connection cannot be made (`--dry-run=server`) AND if the Secret is not present. This poses a couple problems in that CI will have to spin up a cluster first (not a big deal), but users who rely on templating the chart to render manifests will also have to adopt a similar practice.

Open to suggestions here but by attempting to improve UX comes with a cost for those users who define this value.

chipzoller:
I think might have found a neat trick/workaround. By using the `Capabilities` call, this will implicitly only work when connected to a real cluster; otherwise it uses the local client's stored API versions.

JesseGoodier (approved):

chipzoller:
Updated initial post with new results.

michaelmdresser (approved):
This is awesome, thanks Chip

cost-analyzer/templates/_helpers.tpl
```diff
@@ -992,7 +1006,7 @@ Begin Kubecost 2.0 templates
       readOnly: true
   {{- end }}
   {{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
-    - name: {{ .Values.kubecostProductConfigs.cloudIntegrationSecret }}
+    - name: cloud-integration
```
thomasvn, line 1009: Oh wait. Are we sure we want to be hardcoding this to `cloud-integration`? I think the primary requirement is that the key is named `cloud-integration.json`, but the secret can have any name?

ameijer: this looks like the name of the volume that he is mounting the secret as - is internal to the pod spec

cost-analyzer/templates/_helpers.tpl
```diff
@@ -108,6 +108,20 @@ ERROR: MISSING EBS-CSI DRIVER WHICH IS REQUIRED ON EKS v1.23+ TO MANAGE PERSISTE
...
 
+{{/*
+Verify the cloud integration secret exists with the expected key when cloud integration is enabled.
+*/}}
+{{- define "cloudIntegrationSecretCheck" -}}
+{{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
+{{-  if .Capabilities.APIVersions.Has "v1/Secret" }}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.kubecostProductConfigs.cloudIntegrationSecret }}
+  {{- if or (not $secret) (not (index $secret.data "cloud-integration.json")) }}
+    {{- fail (printf "The cloud integration secret '%s' does not exist or does not contain the expected key 'cloud-integration.json'" .Values.kubecostProductConfigs.cloudIntegrationSecret) }}
```
ameijer, line 119: not for now, but could this be expanded in the future to include a schema check of the json contents?

ameijer (approved):

cost-analyzer/templates/_helpers.tpl
thomasvn: Ah! I see. It's the name of the volume:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/456f992b65d2652b625a1996aaa3dedf9ca00e9b/cost-analyzer/templates/aggregator-cloud-cost-deployment.yaml#L54

Thanks for clarifying.

cost-analyzer/templates/_helpers.tpl
chipzoller: I thought about doing that, yes, but think it's best to perhaps leave that outside of the chart or at least address later.

cost-analyzer/templates/_helpers.tpl
chipzoller: Correct, there is no reason the name of the volume needs to be influenced by the value. What matters is the value be pointing to the right Secret.

chipzoller:
I'm going to go ahead and merge this since it seems there have been enough eyes on it. Thanks to everyone.

PR #3051, opened 2024-02-01
EKS values for 2.0

chipzoller:

## What does this PR change?

Updates EKS values file for 2.0 images

No

Allows EKS users referencing this values file to get updated images including the new forecasting image.

N/A

Maybe some images can't be fetched.

## How was this PR tested?

Untested but verified all images have been mirrored into ECR.

N/A

### DIFF
--- a/cost-analyzer/values-eks-cost-monitoring.yaml
+++ b/cost-analyzer/values-eks-cost-monitoring.yaml
@@ -66,6 +66,9 @@ kubecostModel:
     #  cpu: "800m"
     #  memory: "256Mi"
 
+forecasting:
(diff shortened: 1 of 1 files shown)


thomasvn (approved):
Thanks! Just curious, do you know if Dependabot is capable of bumping these for us?

chipzoller:
It's capable, yes, but these images are in a different registry and so the images have to be available there (not the ones referenced by non-EKS values). The new automated image synchronization workflows I've put into place elsewhere will do this, but what we have to be careful of is not accepting Dependabot's proposals for this values file if they haven't been made in the main values file, or else an image discrepancy will occur.

PR #3062, opened 2024-02-02
Add some missing Helm values

chipzoller:

## What does this PR change?

* Adds (some) values into the main values file which were unaccounted for.

No

No impact other than good documentation.

N/A

No functional changes made.

## How was this PR tested?

`helm template` with current state of the values file.

N/A


cost-analyzer/values.yaml
```diff
@@ -2914,6 +2983,9 @@ federatedETL:
   ## Prometheus)
   useMultiClusterDB: false
 
+  ## Define if this is the primary cluster
+  primaryCluster: false
```
thomasvn: I'd recommend removing this, as this is only a value used in legacy FedETL installations. As of 2.x there is no concept of a primary FedETL cluster. Instead, the Aggregator should be run on the cluster which was previously designated the primary. Overall, these changes look great though!

thomasvn (approved):

cost-analyzer/values.yaml
chipzoller: Ah thank you, that shouldn't have been there. I think when I resolved conflicts it got included somehow. Fixed.

PR #3076, opened 2024-02-06
[CI] ignore missing schemas with kubeconform

chipzoller:

## What does this PR change?

Sets the `-ignore-missing-schemas` flag to Kubeconform to allow not checking of CRDs.

Outgoing dependency to #3075 

CI only.

We may actually need CRD schema checking at some point which won't happen by this change.

## How was this PR tested?

Tested in CI.

N/A

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -44,7 +44,7 @@ jobs:
         uses: docker://ghcr.io/yannh/kubeconform:latest
         with:
           entrypoint: /kubeconform
-          args: "-summary -output text full.yaml"
+          args: "-summary -ignore-missing-schemas -output text full.yaml"
 
 
   # Test cluster versions.


ameijer (approved):
lgtm. Chip, do you have a jira board to track these outstanding follow ups like this 'someday add CRD checking' callout?

chipzoller:
I do not. In this specific case, I don't think it's necessary to log something like this, but in general I would have created a GitHub issue or Jira. I'm still not clear which we prefer. Do you have advice here?

ameijer:
Thinking about your specific domain here, I guess I would recommend Jira for a few reasons. Having your own 'helm chart' or 'integrations' Jira board would let you estimate the work, and you would have complete control over the ordered backlog, which you don't have here since this is a public issue stream. Even if you aren't doing sprints or biting down the backlog yourself, if it is already in jira, it can get moved over to another teams sprint when it is decided that does need to get built. Just my 0.02

chipzoller:
Haven't forgot about this and appreciate the feedback. My concern is creating something that no one else sees and/or cares about other than me. I will check with Thomas and folks to see what path they'd suggest.

PR #3082, opened 2024-02-06
thanos-converter

JesseGoodier:


cost-analyzer/values.yaml
```diff
@@ -610,6 +611,14 @@ etlUtils:
   tolerations: {}
   affinity: {}
 
+  thanosConverter:
```
chipzoller, line 634: Can you add a top-level comment here, please?

chipzoller (changes requested):

PR #3083, opened 2024-02-06
Add CICD option to platforms with skipSanityChecks

### LINKED ISSUE #3078: [Feature] Allow the 'cloudIntegrationSecret' sanity check to be disabled
### Problem Statement

The addition of the sanity check (that cannot be disabled presently) prevents the deployment of cost-analyzer by argocd, which runs a 'helm template' command to generate k8s manifests and then separately applies the rendered manifests. The helm template command is unable to process the 'lookup' function call included in the sanity check due to the fact that helm does not contact the k8s API during the process (as long as the '--dry-run=server' arg isn't provided), and so the templating fails due to the sanity check. 

See here the helm documentation for the 'lookup' function which explains this behaviour at the bottom: https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function

For cost-analyzer to be deployable with argocd, the sanity check would need to either be bypassed or made able to recognise when a k8s api connection is not present. 

As stated in the PR by @chipzoller here that added the check - https://github.com/kubecost/cost-analyzer-helm-chart/pull/3048
"It could potentially impact users who use tools or generate manifests from helm template".
Argocd is one CD tool which is impacted in that predicted way. 

### Solution Description

A nice solution would be to allow the sanity check to be toggled on/off with an '.enabled' flag. The flag would be set 'true' by default, allowing affected users to be able to actively disable it if required. 

### Alternatives

_No response_

### Additional Context

_No response_

### Troubleshooting

- [X] I have read and followed the [issue guidelines](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/ISSUE_GUIDELINES.md) and this is a feature request only for the Helm chart.

chipzoller:

## What does this PR change?

Adds a new map under `global.platforms` called `cicd{}` which is largely a placeholder for future settings but presently contains a single option called `skipSanityChecks`. Enabling this option (along with `enabled: true`) will result in Helm (or a CD tool, mainly Argo CD) _not_ attempting to perform the sanity/existence check when a cloud integration Secret is used. This is needed because Argo CD does not presently support the Helm `lookup()` function and has been a well-known issue documented [here](https://github.com/argoproj/argo-cd/issues/5202).

No

Allows users who are deploying the chart via Argo CD or some other affected CD tool to continue deploying by setting these new options with the understanding that no sanity/existence check will be performed.

Closes #3078

The sanity check could still be broken for Argo CD users although this was tested locally.

## How was this PR tested?

`helm install` using the necessary switches to disable sanity checks.

No

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -143,12 +143,12 @@ ERROR: MISSING EBS-CSI DRIVER WHICH IS REQUIRED ON EKS v1.23+ TO MANAGE PERSISTE
 
 {{/*
 Verify the cloud integration secret exists with the expected key when cloud integration is enabled.
-Skip the check if Argo CD is enabled and skipSanityChecks is set since Argo CD presently does not
+Skip the check if CI/CD is enabled and skipSanityChecks is set. Argo CD, for example, does not
 support templating a chart which uses the lookup function.
 */}}
 {{- define "cloudIntegrationSecretCheck" -}}
 {{- if (.Values.kubecostProductConfigs).cloudIntegrationSecret }}
-{{- if not (and .Values.global.platforms.argocd.enabled .Values.global.platforms.argocd.skipSanityChecks) }}
+{{- if not (and .Values.global.platforms.cicd.enabled .Values.global.platforms.cicd.skipSanityChecks) }}
 {{-  if .Capabilities.APIVersions.Has "v1/Secret" }}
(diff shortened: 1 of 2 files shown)


thomasvn (approved):
Thanks! Lgtm. This is something I'd think we want to patch in v2.0.x as well?

chipzoller:
It's technically a feature although it's low risk.

chipzoller:
@jessegoodier, please review as you're able.

chipzoller:
I made this PR more generic in nature `argocd` => `cicd`

JesseGoodier (approved):

PR #3090, opened 2024-02-07
cloud-cost-volume-logic

JesseGoodier:
## What does this PR change?
This gives preference to federated-store over object-store for cloud costs. Should only fix issues and be backwards compatible.


jessegoodier:
Not going to merge yet, but want to get your thoughts @Sean-Holcomb. Does this make sense to you? 
I found an environment where cloud-cost was looking for `object-store.yaml` but the file was `federated-store.yaml`

jessegoodier:
Not going to merge yet, but want to get your thoughts @Sean-Holcomb. Does this make sense to you? 
I found an envrironment where cloud-cost was looking for `object-store.yaml` but the file was `federated-store.yaml`

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1040,8 +1040,7 @@ Begin Kubecost 2.0 templates
     - name: federated-storage-config
       mountPath: /var/configs/etl/federated
       readOnly: true
-  {{- end }}
-  {{- if .Values.kubecostModel.etlBucketConfigSecret }}
+  {{- else }}
```
Sean-Holcomb: I am not sure about this. when running as a container, `cloud-cost` doesn't depend on a secret

Sean-Holcomb: maybe an else if

Sean-Holcomb:
Proposal for a clean up effort, making it so there is only one bucket configuration

PR #3099, opened 2024-02-08
provide reasonable default for db copy chunks

ameijer:
## What does this PR change?
Provides a 'middle ground' setting in the time vs resources tradeoff for the db copy operation 

no

it will make new data show up in kubecost UI sooner, while keeping ram usage reasonable

## How was this PR tested?
tested via chart template rendering


chipzoller:
CI is going to fail here because `cost-model-nightly` issues, FYI.

michaelmdresser (approved):
Perfect. Great documentation.

michaelmdresser:
This is a major UX improvement, I think its worth picking into 2.1

PR #3105, opened 2024-02-09
adding helm configurable cluster turndown

ameijer:
## What does this PR change?
adds ability to configure cluster turndown from kubecost helm chart yaml values

no

allows them to configure cluster turndown from their yaml 

## How was this PR tested?
ensured it passed helm template --validate, and that generated CR matched spec


ameijer:
@teevans as per your request

nickcurie (approved):
Think this needs a rebase, but otherwise LGTM

ameijer:
bad bot

cost-analyzer/templates/kubecost-cluster-controller-actions-config.yaml
```diff
@@ -38,4 +38,18 @@ metadata:
...
+apiVersion: kubecost.com/v1alpha1
+kind: TurndownSchedule
+metadata:
+  name: {{ .name }}
+  finalizers:
+  - "finalizer.kubecost.com"
+spec:
+  start: {{ .start }}
+  end: {{ .end }}
+  repeat: {{ .repeat }}
```
chipzoller: Have you tested this to ensure the CR is properly created? The CRD is deployed as part of the `kubecost-cluster-controller-template.yaml` template.

ameijer: I ran it through helm template --validate. Was figuring we can rely on helm to install CRD before the CRs that use them

chipzoller: Can you do an actual installation to ensure that happens? Generally speaking, it's not a great idea to start mixing standard and custom resources in the same template file.

ameijer: do we have a separate ci-values.yaml file? would be nice to deploy cluster controller to test stuff like this in general in the kind matrix

chipzoller: Yes, it's being done [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/ci/statefulsets-cc.yaml#L22-L23) already. Action run for it is [here](https://github.com/kubecost/cost-analyzer-helm-chart/actions/runs/7872700021/job/21478482827#step:11:171). I guess it's fine.

ameijer: we can add an enhancement to those CI values that instantiate that pre-defined config just to confirm this case is true. I will hold the action item to add something like that

chipzoller: Would you be logging a Jira issue or a GitHub issue? It would indeed be nice to supply some values here and ensure they are laid down by the CR.

ameijer: I was just going to do it, perhaps in like 30-60 mins?

chipzoller: I would still create a GitHub enhancement to track.

cost-analyzer/crds/turndown-schedule.yaml
chipzoller, line 1: We may not want to do this but have the CRD as a separate resource outside the `crds` folder for reasons listed [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations). This is a classic problem with Helm, FYI, and many a headache are caused over it.

ameijer: @chipzoller I couldn't get this to work with the CRD anywhere outside the crds directory

chipzoller: Does it install the CRD if it's part of another template, in a multi-document file?

ameijer: I have tried both. https://github.com/kubecost/cost-analyzer-helm-chart/pull/3105/commits/e4b1737991463af951658ee45b89e0da92f0ccb3 tried as another document in the file. https://github.com/kubecost/cost-analyzer-helm-chart/pull/3105/commits/57d1847597a11bc465642f31db57cf62e58c2e95 had it in its own file. both times failed with 
```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "my-schedule2" namespace: "" from "": no matches for kind "TurndownSchedule" in version "kubecost.com/v1alpha1"
ensure CRDs are installed first
```

chipzoller: It's looking like uninstallation is where it's failing. If you test this locally, what's happening?

ameijer: the issue here is that the cluster turndown logic uses a finalizer to unregister from its internal scheduler. However, in an uninstall situation, the cluster controller might go before the TurndownSchedule, leaving nothing to remove the finalizer, leading to a hung uninstall. You will note that on the most recent build, I add a CI specific value to omit the finalizer when using in CI matrices

chipzoller: Typically finalizers are not part of the CR; they get added by the controller when reconciling the CR. You may need to resort to using a Helm pre- or post-delete hook in these cases. Or just simply don't use a finalizer and use an informer to watch and internally deregister any CRs from the reconciliation.

ameijer: ok, let me blow the finalizer away

ameijer: @chipzoller that worked, CI passing now. Are we good to give the crds dir solution a try?

chipzoller: You understand what the consequences are, right? Namely, you will not be able to use Helm to upgrade/modify this CRD when future changes are made. I don't think we want to commit to that.

ameijer: No worries, chip. I guess where I am coming from is that, as I understand it, there business value in allowing folks to configure objects that this CRD defines via yaml, and I'm trying to meet the goals set out for me here. I'd love your input in how we can accomplish that with incurring the least amount of technical debt possible. Here are the options, from my view:
1. use the crds file. pros are that it is easy for users to install and use, cons are that we need to either define a new CRD for the next version or do an upgrade process
2. install the CRDs via a separate helm chart. Pros are it is helm and configurable. Cons are that is is additional friction to the user
3. Install the CRDs via a kubectl apply from some online manifest. Same con as above.

WIth 2 and 3, we will likely need handler code to catch the user and guide them to the solution, but it is still friction any way you cut it.

Do you see any other options? how would you like to proceed?

chipzoller: Absolutely, we definitely want to use the CR approach to define turndown schedules, we just want to make sure we don't hang ourselves later when folks update it.

My suggestion to eliminate the use for finalizers is handle this in the controller, otherwise we will probably have to implement a Helm pre- or post-delete hook. This is going to be a problem for Argo CD users as the brand new release (2.10) just now added support for post-delete hooks. Pre-delete are still not supported.

So can we make this work without using finalizers?

ameijer: yes - I wiped out the finalizers. I think that will be fine. The question is around installing the CRD - if you wouldn't like to use the CRDs directory in the chart, when how would you suggest that we install the CRD?

chipzoller: Put the CRD as a regular template along with everything else. Maybe it's fine to leave it in the main cluster controller file.

ameijer: unfortunately, when I do that, I get the following error: 
```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "my-schedule2" namespace: "" from "": no matches for kind "TurndownSchedule" in version "kubecost.com/v1alpha1"
ensure CRDs are installed first
```
I got this same error when trying to place it in multiple templated locations: https://github.com/kubecost/cost-analyzer-helm-chart/commit/e4b1737991463af951658ee45b89e0da92f0ccb3 and https://github.com/kubecost/cost-analyzer-helm-chart/commit/57d1847597a11bc465642f31db57cf62e58c2e95

chipzoller: Let me take a look.

chipzoller: It looks like use of the `/crds` folder might be best for right now given the choices. The alternative is to place the one lone CRD in a subchart and make a dependency for this chart, which we really don't want to do if we can avoid it.

chipzoller: Good info: https://handbook.giantswarm.io/docs/dev-and-releng/app-developer-processes/handle_crds_with_helm_3/

ameijer: interesting, so they define a whole job structure to manage those CRDs in the context of hooks. Do you want to pursue that? The CRDs directory seems a lot easier, especially for this CR which has not seen any changes in 2 years

chipzoller: The Job strategy is more used for much more complex scenarios and needs than what we have. I think starting with `/crds` is probably fine--and maybe totally fine. One consideration which some people may complain about is lack of CRD removal, which is one of the side effects of this strategy. Some folks want/expect removal of a chart to remove all of the elements it installed, and the CRD will be left behind.

Another strategy is to have your controller create and manage the CRDs. The challenge here is when you're installing a CR, like we are, it has a strict dependency on its CRD which cannot be created until the controller starts. More complex than what our simple needs are at the moment.

ameijer: thanks, going to resolve this then as we will use CRDs dir for initial implementation

PR #3122, opened 2024-02-12
Update MultiCluster Diagnostics endpoint to be `/model/diagnostics/multicluster`

thomasvn:
## What does this PR change?

Updates MultiCluster Diagnostics endpoint from `/diagnostics/api` to `/model/diagnostics/multicluster`. This way regardless of whether this service is run within the `cost-model` container ([code ref](https://github.com/kubecost/kubecost-cost-model/blob/fa9ec60b1a33e45f568bac182d5cd65c3fafa007/pkg/cmd/costmodel/costmodel.go#L1694)) or as a separate `diagnostics` container, its API is always served at the same endpoint.

This abstracts away the need for the frontend/user to determine where the API is located based on how it's deployed.

N/A

N/A

N/A

N/A

## How was this PR tested?

Templated using the following `values.yaml`. Correctly saw the `/model/diagnostics/multicluster` endpoint in the nginx conf.

```sh
helm template ./cost-analyzer -f values.yaml
```

```yaml
# values.yaml
diagnostics:
  enabled: true
  primary:
    enabled: true
  deployment:
    enabled: true
kubecostModel:
prometheus:
  server:
    global:
      external_labels:
        cluster_id: thomasn-diagnostics
```

N/A

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -993,14 +993,14 @@ data:
         }
 
         {{- if and .Values.diagnostics.enabled .Values.diagnostics.primary.enabled .Values.diagnostics.deployment.enabled }}
-        # When the Multi-cluster Diagnostics Service is run within the
-        # cost-model container, its endpoint is available at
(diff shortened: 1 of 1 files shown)


michaelmdresser (approved):
Ah! I see, this makes sense.

PR #3129, opened 2024-02-13
helm upgrade check logic

JesseGoodier:
## What does this PR change?

1. fix boolean logic on v2 upgrade with multi-cluster
2. remove block on amp/mimir/gmp v2 upgrades

No

Added support in Kubecost 2.1 for multi-cluster AMP/GMP/Mimir 

Unknown/untested issues with one of the platforms.

## How was this PR tested?
ADOT QA environment.

todo

### DIFF
--- a/cost-analyzer/crds/cluster-turndown-crd.yaml
+++ b/cost-analyzer/crds/cluster-turndown-crd.yaml
@@ -1,3 +1,4 @@
+{{- if (.Values.clusterController).enabled }}
 # TurndownSchedule Custom Resource Definition for persistence
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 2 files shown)


cost-analyzer/crds/cluster-turndown-crd.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if (.Values.clusterController).enabled }}
```
jessegoodier: @ameijer 
does this make sense? Is there a use case for the CRD if the controller isn't enabled?

ameijer: yes, however, the CRDs directory doesn't support templating :(

jessegoodier:
closing, found new bug

PR #3145, opened 2024-02-15
dataBackupConfigured flag

JesseGoodier:

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -1055,8 +1055,10 @@ data:
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;
             return 200 '\n
-                {"ssoConfigured": "{{ template "ssoEnabled" . }}"},\n
-                {"dataBackupConfigured": "{{ template "dataBackupConfigured" . }}"}\n
+                {
+                "ssoConfigured": "{{ template "ssoEnabled" . }}",
+                "dataBackupConfigured": "{{ template "dataBackupConfigured" . }}"
+                }
             ';
         }
     }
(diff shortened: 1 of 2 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -1055,7 +1055,8 @@ data:
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;
             return 200 '\n
-                {"ssoConfigured": "{{ template "ssoEnabled" . }}"}\n
+                {"ssoConfigured": "{{ template "ssoEnabled" . }}"},\n
+                {"dataBackupConfigured": "{{ template "dataBackupConfigured" . }}"}\n
```
thomasvn:
```suggestion
                {
                "ssoConfigured": "{{ template "ssoEnabled" . }}",
                "dataBackupConfigured": "{{ template "dataBackupConfigured" . }}"
                }
```

I think (?) we should be returning a single JSON object and not two?

thomasvn:
Looks good! But in what scenarios in 2.x is `etlBucketConfigSecret` configured and `federatedStorageConfigSecret` not?

jessegoodier:
> Looks good! But in what scenarios in 2.x is `etlBucketConfigSecret` configured and `federatedStorageConfigSecret` not?

single cluster

thomasvn (approved):
Once merging, we should check that any frontend calls to `/model/productConfigs` are still able to properly parse the response being returned by Nginx.

I'm not 100% certain that the formatting here is correct

thomasvn:
Looking at the nightly environment, it seems like calls to `/model/productConfigs` show up like this:

```


            {
            "ssoConfigured": "false",
            "dataBackupConfigured": "true"
            }
        
```

The extra whitespace is weird, but I _think_ this is still valid parseable JSON. We may need to revisit to clean up one day.

PR #3161, opened 2024-02-19
Add loadBalancerSourceRanges for Cost Analyser service

itsmesuniljacob:
## What does this PR change?
This is to support `loadBalancerSourceRanges` for Cost Analyzer service template

No
To provide explicit access to the subnets that are communicating with your load balancer

Check if, Cost Analyser can be accessed  only with the IP's specified in the load balancer range
## How was this PR tested?
This was tested locally

Not made

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-service-template.yaml
@@ -25,6 +25,15 @@ spec:
 {{- else }}
   type: ClusterIP
 {{- end }}
+{{- if (eq .Values.service.type "LoadBalancer") }}
+  {{- if .Values.service.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
+  {{- end }}
+  {{- if .Values.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+{{ toYaml .Values.service.loadBalancerSourceRanges | indent 4 }}
+  {{- end -}}
+{{- end }} 
   ports:
     - name: tcp-model
       port: 9003
--- a/cost-analyzer/templates/cost-analyzer-service-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-service-template.yaml
@@ -26,9 +26,6 @@ spec:
   type: ClusterIP
 {{- end }}
 {{- if (eq .Values.service.type "LoadBalancer") }}
-  {{- if .Values.service.loadBalancerIP }}
-  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
-  {{- end }}
   {{- if .Values.service.loadBalancerSourceRanges }}
   loadBalancerSourceRanges:
 {{ toYaml .Values.service.loadBalancerSourceRanges | indent 4 }}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -731,6 +731,8 @@ service:
   # nodePort:
   labels: {}
   annotations: {}
+  loadBalancerIP: ""
+  loadBalancerSourceRanges: []
   sessionAffinity:
     enabled: false  # Makes sure that connections from a client are passed to the same Pod each time, when set to `true`. You should set it when you enabled authentication through OIDC or SAML integration.
     timeoutSeconds: 10800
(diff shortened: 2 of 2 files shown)


itsmesuniljacob:
@thomasvn @chipzoller @ameijer  Can you review this PR?

cost-analyzer/templates/cost-analyzer-service-template.yaml
```diff
@@ -25,6 +25,15 @@ spec:
 {{- else }}
   type: ClusterIP
 {{- end }}
+{{- if (eq .Values.service.type "LoadBalancer") }}
+  {{- if .Values.service.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
```
ameijer: @itsmesuniljacob this looks like it was deprecated as of 1.24, do you think it still merits inclusion?

cost-analyzer/values.yaml
```diff
@@ -731,6 +731,8 @@ service:
   # nodePort:
   labels: {}
   annotations: {}
+  loadBalancerIP: ""
+  loadBalancerSourceRanges: []
```
ameijer: comment here explaining the use of this param?

cost-analyzer/templates/cost-analyzer-service-template.yaml
itsmesuniljacob:
```
 {{- if .Values.service.loadBalancerIP }}
  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
```

I can remove this code

cost-analyzer/values.yaml
itsmesuniljacob: We want to allow only some IP's to access Kubecost

chipzoller: I would prefer these be commented out by default since they aren't applicable when in ClusterIP mode.

itsmesuniljacob: That's true. That's why,  I have a condition like below:
```
{{- if (eq .Values.service.type "LoadBalancer") }}
  {{- if .Values.service.loadBalancerSourceRanges }}
  loadBalancerSourceRanges:
{{ toYaml .Values.service.loadBalancerSourceRanges | indent 4 }}
  {{- end -}
```}

File: cost-analyzer/templates/cost-analyzer-service-template.yaml (hunk 1 above)
Comment (itsmesuniljacob) [reply cost-analyzer/templates/cost-analyzer-service-template.yaml]: Pushed the new change

File: cost-analyzer/values.yaml (hunk 2 above)
Comment (chipzoller) [reply cost-analyzer/values.yaml]: Yes, I understand, but it's cleaner if non-default and especially corner case values be commented out by default. This is a practice we try and follow in this chart. It's also a common practice.

Comment (itsmesuniljacob) [reply cost-analyzer/values.yaml]: Got it. That means when I need it, I can uncomment and use it. Will make the change

Comment (itsmesuniljacob) [reply cost-analyzer/values.yaml]: Updated the same

**Review by ameijer — COMMENTED** [review]
LGTM, will leave it to @chipzoller for final approval

File: cost-analyzer/values.yaml [hunk 3]
```diff
@@ -731,6 +731,8 @@ service:
   # nodePort:
   labels: {}
   annotations: {}
+  # loadBalancerIP: ""
```
chipzoller: This is no longer present in a template. Remove, please.

chipzoller (changes requested):

cost-analyzer/values.yaml
itsmesuniljacob: Done

chipzoller (approved):

ameijer (approved):

itsmesuniljacob:
@ameijer @chipzoller May I know, which version tag will be applied to this, so that I can use ?

chipzoller:
Will be included in the next version, 2.2.

itsmesuniljacob:
@chipzoller May I know , when this changed is will be tagged ?

chipzoller:
We don't have a release date set for 2.2 yet. I would suspect sometime around the next month.

itsmesuniljacob:
Oh ok ! Thanks for the update

dwbrown2:
Aiming for before Kubecon EU.

PR #3169, opened 2024-02-21
use badger with jaeger

nickcurie:
## What does this PR change?
Enables Jaeger to use badger, an embedded local storage

No

Moved Jaeger to use embedded local storage

https://kubecost.atlassian.net/browse/SELFHOST-1168

Low to no risk, we may want to keep an eye on deployments running badger, just to ensure jaeger is still tracing properly.

## How was this PR tested?
Ran on gke-dev-1 cluster, observed existence of a badger directory in the file system. Pod logs also reference badger. This was let run for ~12 hours and there are no issues


ameijer (approved):
LGTM, nice work

PR #3171, opened 2024-02-22
Kubecost Integrations - Postgres

thomasvn:
## What does this PR change?

Introduces the first Kubecost Integration. The first integration is designed to periodically pull data from Kubecost to store into a Postgres Database.

N/A

Introduces the first Kubecost Integration. This first integration is designed to periodically pull data from Kubecost to store into a Postgres Database.

N/A

N/A

## How was this PR tested?

```yaml
global:
  integrations:
    postgres:
      enabled: true
      databaseHost: REDACTED.postgres.database.azure.com
      databasePort: 5432
      databaseName: postgres
      databaseUser: thomasn
      databasePassword: REDACTED
      schedule: "0 * * * *"
```

The following command correctly templates the Cronjob:

```sh
helm template ./cost-analyzer -f values-integrations.yaml -s templates/integrations-postgres-cronjob.yaml | less
```

TODO

## TODO

- [x] Validate successful Cronjob runs
- [x] Validate data in Postgres
- [x] Promote image to the `gcr.io/kubecost1` registry


thomasvn:
@mclean0328 @AjayTripathy @chipzoller Still some work to be done to test and document this integration. However, let me know any thoughts on this PR so far!

chipzoller:
I think I can infer the usefulness of this, but I'd still like to hear the use case and the value.

That aside, I personally believe this is something which shouldn't be added into the main chart but exist as a separate repository/sub-project in the Kubecost org. This chart is already so overinflated with complexity that we need to find ways to simplify and reduce that complexity and not add to it. There are many ways this integration could be delivered including as a separate Helm chart, a static manifest, a manifest with Kustomize overlay, etc. This direction is also aligned with common industry practices; integrations like this aren't part of the "core" application, are optional and highly circumstantial, and as a result are often left self-contained.

thomasvn:
@chipzoller Thanks for the feedback! I agree that eventually adding a dozen or more integrations into this chart would create quite some complexity/overhead in the values and templates.

I'd still like to prioritize convenience & low friction for users installing these integrations though. What do you think is the path of least friction? I think it may be a separate Helm chart as you mentioned? Anyway, need to pause work on this feature for now so let's discuss at a later point!

PR #3175, opened 2024-02-23
Fix failed install/template with no values on older versions of Helm3

michaelmdresser:
## What does this PR change?
`helm template kubecost ./cost-analyzer` (and thus also installing with no values) was failing with this error:

`Error: template: cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml:57:52: executing "cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml" at <.Values.kubecostFrontend.api.fqdn>: nil pointer evaluating interface {}.fqdn`
and also this error
`Error: template: cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml:589:15: executing "cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml" at <eq .Values.kubecostAggregator.env.MEMORY_INTENSIVE_CLUSTER_SIZING "enabled">: error calling eq: incompatible types for comparison`

This PR fixes both. See commit messages for the explanation of each fix.

N/A

- Fixed an install error when no values are provided -- one or both of the following errors: `nil pointer evaluating interface {}.fqdn` and `error calling eq: incompatible types for comparison`. This error occurred for users of older versions of Helm3. The current version of Helm3, `v3.14.0`, is unaffected.

The new fqdn feature may not work (I expect this is fine, we use the same pattern elsewhere with parens). The intensive cluster sizing feature may not work.

## How was this PR tested?
`helm template kubecost ./cost-analyzer` failed before, succeeds after

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -54,7 +54,7 @@ data:
     upstream api {
 {{- if .Values.kubecostFrontend.useDefaultFqdn }}
         server {{ $serviceName }}.{{ .Release.Namespace }}.svc.cluster.local:9001;
-{{- else if and .Values.kubecostFrontend.api .Values.kubecostFrontend.api.fqdn }}
+{{- else if (.Values.kubecostFrontend.api).fqdn }}
         server {{ .Values.kubecostFrontend.api.fqdn }};
 {{- else }}
         server {{ $serviceName }}.{{ .Release.Namespace }}:9001;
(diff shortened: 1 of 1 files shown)


nikovacevic (approved):

michaelmdresser:
Apparently this is not an issue for recent versions of Helm (e.g. 3.14.0). I was on `version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"clean", GoVersion:"go1.16.5"}` when I saw these errors, which is why this was not caught in CI.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -54,7 +54,7 @@ data:
     upstream api {
 {{- if .Values.kubecostFrontend.useDefaultFqdn }}
         server {{ $serviceName }}.{{ .Release.Namespace }}.svc.cluster.local:9001;
-{{- else if and .Values.kubecostFrontend.api .Values.kubecostFrontend.api.fqdn }}
+{{- else if (.Values.kubecostFrontend.api).fqdn }}
```
thomasvn, line 57: Thanks for this fix! I'm surprised the old logic doesn't work here?

I would've thought that Helm would try to evaluate whether `.Values.kubecostFrontend.api` exists (which it doesn't by default), then stop there before attempting to evaluate `.Values.kubecostFrontend.api.fqdn`. Regardless, thanks!

michaelmdresser: I'm pretty sure that older versions of Helm did not have short-circuiting on the `and` operator

Issue #3172, opened 2024-02-23
[Bug] Secondaries Aggregator can't be turned off.


passionInfinite:
### Kubecost Helm Chart Version

v2.0.2 and v2.1.0 both

### Kubernetes Version

v1.27.7

### Kubernetes Platform

AKS

### Description

If I have `kubecostAggregator.enabled: false` and `kubecostAggregator.deployMethod: disabled` . The nginx.conf file fails to process successfully and thus keeps the frontend pod restarting.

The reason for nginx.conf not get parsed successfully is because upstream for aggregation is not defined due to disabling of aggregator but the endpoints in the conf file still tries to use it which nginx throws error saying [emer] aggregator host not defined in upstream.

I believe all those endpoints server should be configured based on the aggregator settings. Currently they are hardcoded expecting it to have aggregator running.

Example: https://github.com/kubecost/cost-analyzer-helm-chart/blob/v2.1.0/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml#L309

### Steps to reproduce

1. Disable aggregator and also make deployMethod as `disabled`.
2. Deploy the kubecost. The frontend pod will keep on restarting failing to parse the nginx.conf file.

### Expected behavior

I expect even if the aggregator is disabled everything else should work as is.

### Impact

We can't really use v2.x version.

### Logs

```Shell
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration                                                                                            │
│ /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/                                                                                                                   │
│ /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh                                                                                                       │
│ 10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)                                                                               │
│ /docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh                                                                                                               │
│ /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh                                                                                                           │
│ /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh                                                                                                           │
│ /docker-entrypoint.sh: Configuration complete; ready for start up                                                                                                                           │
[...]

PR #3184, opened 2024-02-27
fix: nginx.conf for saml and oidc endpoints

passionInfinite:
## What does this PR change?
This PR fixes the nginx.conf and includes the SAML, OIDC, login and logout endpoints conditional based on the required supplied values. This will help to fix #3172  as those nginx locations doesn't look for aggregator upstream hosts. The SAML, OIDC, Login and Logout endpoints are only supported for Federated Kubecost UI so technically this condition will help achieve those nginx locations.

## Does this PR rely on any other PRs?

Not at the moment.

## How does this PR impact users? (This is the kind of thing that goes in release notes!)

All the users using v2.x.x will be able to use Kubecost without enabling aggregator in the secondaries. This is same issue for v2.1.0 and gets fixed for that version as well.

## Links to Issues or tickets this PR addresses or fixes

#3172 

## What risks are associated with merging this PR? What is required to fully test this PR?

Tested using the `helm template` function by supplying the below values and saw the output template for nginx configmap if resolving correctly or not.

values supplied:

```
clusterController:
  enabled: true

federatedETL:
  agentOnly: false
  federatedCluster: true

kubecostAggregator:
  enabled: false
  deployMethod: 'disabled'

kubecostFrontend:
  enabled: true
```

The resolved templates looks like below:

```
---
# Source: cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-conf
  namespace: default
  labels:
    
    app.kubernetes.io/name: cost-analyzer
    helm.sh/chart: cost-analyzer-2.0.0
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
    app: cost-analyzer
data:
  nginx.conf: |
    gzip_static  on;

    # Enable gzip encoding for content of the provided types of 50kb and higher.
    gzip on;
    gzip_min_length 50000;
    gzip_proxied expired no-cache no-store private auth;
    gzip_types
        application/atom+xml
        application/geo+json
        application/javascript
        application/x-javascript
        application/json
        application/ld+json
        application/manifest+json
        application/rdf+xml
        application/rss+xml
        application/vnd.ms-fontobject
        application/wasm
        application/x-web-app-manifest+json
        application/xhtml+xml
        application/xml
        font/eot
        font/otf
        font/ttf
        image/bmp
        image/svg+xml
        text/cache-manifest
        text/calendar
        text/css
        text/javascript
        text/markdown
[...]

PR #3185, opened 2024-02-27
added helm values to remove aggregtor when disabled via helm

ghost:
## What does this PR change?
Fixes an issue where disabling aggregator doesn't remove the routing in nginx

not that i'm aware of

UI will most likely be operating at limited capacity

https://github.com/kubecost/cost-analyzer-helm-chart/issues/3172

I would say the risks are low, this feature is already not working as designed. 

## How was this PR tested?
I used a helm chart render locally to verify that the proper helm values get added/removed.

no.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -117,12 +117,12 @@ data:
     }
   {{- end }}
 
-  {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
+  {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) (not .Values.kubecostAggregator.enabled) }}
     upstream aggregator {
         {{- if .Values.kubecostFrontend.useDefaultFqdn }}
         server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}.svc.cluster.local:9004;
         {{- else }}
-        {{- if (.Values.kubecostFrontend.aggregator).fqdn }}
+        {{- if and .Values.kubecostFrontend.aggregator .Values.kubecostFrontend.aggregator.fqdn }}
         server {{ .Values.kubecostFrontend.aggregator.fqdn }};
         {{- else }}
         server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}:9004;
@@ -279,6 +279,7 @@ data:
 {{- else }}
             return 404;
 {{- end }}
+{{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
         }
         location /oidc/ {
             proxy_connect_timeout       180;
@@ -356,7 +357,7 @@ data:
         }
     {{- end }}
     {{- end }}
-
+{{- end}}
 
 {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
     # TODO make aggregator route the default, start special-casing


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -117,12 +117,12 @@ data:
     }
   {{- end }}
 
-  {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
+  {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) (not .Values.kubecostAggregator.enabled) }}
```
thomasvn, line 120: While this is correct, I think that `.Values.kubecostAggregator.enabled` is somewhat of a deprecated config. It was primarily used in Kubecost v1.x.

Now, I believe the primary way to configure Aggregator is via `.Values.kubecostAggregator.deployMethod`. More details here:

https://github.com/kubecost/cost-analyzer-helm-chart/blob/bb92e85ffd303fa0063f687728f3931cbf07cde8/cost-analyzer/values.yaml#L2370-L2382

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -117,12 +117,12 @@ data:
...
     upstream aggregator {
         {{- if .Values.kubecostFrontend.useDefaultFqdn }}
         server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}.svc.cluster.local:9004;
         {{- else }}
-        {{- if (.Values.kubecostFrontend.aggregator).fqdn }}
+        {{- if and .Values.kubecostFrontend.aggregator .Values.kubecostFrontend.aggregator.fqdn }}
```
thomasvn, line 125: I believe that the following two lines are equivalent? Did one not work for you?

```
{{- if (.Values.kubecostFrontend.aggregator).fqdn }}
```

```
{{- if and .Values.kubecostFrontend.aggregator .Values.kubecostFrontend.aggregator.fqdn }}
```

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -279,6 +279,7 @@ data:
 {{- else }}
             return 404;
 {{- end }}
+{{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
```
thomasvn, line 282: Good idea. SAML/OIDC are served by the Aggregator, so we shouldn't include these Nginx routes if Aggregator is disabled. Could you add the same for SAML too? Could we also add some documentation around this requirement?

thomasvn:
@williamkubecost Really great PR! Left a few comments.

ghost:
@thomasvn thank you, i'm actually reviewing this [PR](https://github.com/kubecost/cost-analyzer-helm-chart/pull/3184#issuecomment-1966791457)

Not sure what your thoughts are on the changes there. LMK if we should proceed with this method or the other 😄

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
ghost: @thomasvn thanks for the feedback. if it's deprecated, do you see harm leaving this as is? someone could still technically enable/disable it, which could trigger the issue.

jessegoodier: FWIW, I haven't used deployMethod yet. I failed to read the comments closer.

we also have an option for agentOnly
which is probably preferable to deployMethod=false  because it kills grafana/frontend/forecasting/aggregator 
?
https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L3012

These if strings are getting a little long, but I don't have a strong opinion on the right method.

ghost:
Closing this in favor of: https://github.com/kubecost/cost-analyzer-helm-chart/pull/3184#issuecomment-1966791457

PR #3202, opened 2024-03-05
Add 2.1.0 for AWS Marketplace

chipzoller:
## What does this PR change?

Creates a "forked" version of the Kubecost 2.1.0 Helm chart to comply with the requirements of [AWS Marketplace](https://aws.amazon.com/marketplace/pp/prodview-asiz4x22pm2n2) of which there were several that necessitated this path. The requirements are listed [here](https://docs.aws.amazon.com/marketplace/latest/userguide/container-product-policies.html#publishing-eks-add-on).

Note that this is to an entirely new branch called `v2.1-aws-marketplace` and is not being proposed for the `develop` branch.

No

Allows Kubecost to offer version 2.1.0 on the AWS Marketplace by meeting their requirements.

N/A

Only present in the [AWS Marketplace listing](https://aws.amazon.com/marketplace/pp/prodview-asiz4x22pm2n2).

## How was this PR tested?

Templated locally

N/A


thomasvn:
@chipzoller Looks good! So the primary difference here is that there are no dynamic checks to see whether an existing k8s cluster has a capability available or a secret in existence?

Also, could you share what commands you used to generate the `.tar.gz` of the Helm chart? Just for posterity.

thomasvn (approved):

chipzoller:
Correct, AWS Marketplace has several restrictions on Helm charts. These restrictions were not in place when we created the original listings but are now. The pertinent ones were no use of `Release.Service`, no `lookup()`, no `Capabilities.APIVersion` for non-native resources, and the need to have an icon in `Chart.yaml`.

To package a chart all that's needed is `helm package`.

PR #3206, opened 2024-03-07
Add log level to network costs

cliffcolvin:
## What does this PR change?
Bumps Network Costs image to 0.17.3 that changes noisier log entries to debug and allow for a better experience understanding successful pod state at INFO level.
Adds LOG_LEVEL to environment variable defaulted to info for network costs.

Adds info log level as the default for network costs pod, and allows for setting the log level depending on desired level (ERROR|WARN|INFO|DEBUG). 

## How was this PR tested?
We have tested our deployments locally to see that INFO log level is in fact quieter while giving pod state and still showing warnings and errors in logs at the default INFO level.


chipzoller (approved):

chipzoller:
Will you handle the cherry picking?

PR #3218, opened 2024-03-10
helm client version check

JesseGoodier:

### DIFF
--- a/cost-analyzer/templates/NOTES.txt
+++ b/cost-analyzer/templates/NOTES.txt
@@ -1,4 +1,5 @@
 --------------------------------------------------
+{{- include "helmClientVersionCheck" . -}}
 {{- include "kubecostV2-preconditions" . -}}
 {{- include "cloudIntegrationSourceCheck" . -}}
 {{- include "eksCheck" . -}}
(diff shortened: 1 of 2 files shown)


jessegoodier:
@chipzoller 
Do you know what version that fqdn PR needs?

I just made up a version here that is about a year old.

chipzoller:
> Do you know what version that fqdn PR needs?

???

jessegoodier:
if anyone know what version fixes this, I think that would be the check for this PR
```
Error: template: cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml:57:52: executing "cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml" at <.Values.kubecostFrontend.api.fqdn>: nil pointer evaluating interface {}.fqdn
```

chipzoller:
It'll be pre-3.10

jessegoodier:
@chipzoller 
you good with this version check? I'll own the code change.

chipzoller:
This should be tested on CD applications first like Argo CD and Flux. Needs to be documented in the README clearly, and the PR template should be completed.

jessegoodier:
> This should be tested on CD applications first like Argo CD and Flux. Needs to be documented in the README clearly, and the PR template should be completed.

really good point. Need to rethink this one.

chipzoller:
It could still be viable, we would just have to wrap that cicd switch around this one as well or else it could hang up some folks.

PR #3220, opened 2024-03-11
FIX templating for non-existing nested values of frontend upstreams

korjek:
[this commit](https://github.com/kubecost/cost-analyzer-helm-chart/commit/133c886096a616a9fb33b48b3feb76223a490d7d) introduced a bug where helm fails on non-existing nested values.

## What does this PR change?
Correct handling of non-existing nested values introduced in [commit](https://github.com/kubecost/cost-analyzer-helm-chart/commit/133c886096a616a9fb33b48b3feb76223a490d7d)
Example of the error:
```
Error: Failed to render chart: exit status 1: Error: template: cost-analyzer/charts/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml:57:52: executing cost-analyzer/charts/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml at <.Values.kubecostFrontend.api.fqdn>: nil pointer evaluating interface {}.fqdn
```

No

Fixes helm failure if value doesn't exist

I can't think of any risks.
To test just try to install version with the fix and values that worked before v2.1.0 release.

## How was this PR tested?
Helm release was updated from v2.0.2 to v2.1.0 (v2.1.0 was built from this branch)

Not needed

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -54,7 +54,7 @@ data:
     upstream api {
 {{- if .Values.kubecostFrontend.useDefaultFqdn }}
         server {{ $serviceName }}.{{ .Release.Namespace }}.svc.cluster.local:9001;
-{{- else if and .Values.kubecostFrontend.api .Values.kubecostFrontend.api.fqdn }}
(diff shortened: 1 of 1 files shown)


korjek:
This was fixed in [https://github.com/kubecost/cost-analyzer-helm-chart/pull/3175](https://github.com/kubecost/cost-analyzer-helm-chart/pull/3175) (not released yet though). So closing

PR #3234, opened 2024-03-14
agentOnly logic fixes

JesseGoodier:
## What does this PR change?
Removes services that aren't used with agentOnly configs

No

Cleanup of misc services deploy when agentOnly is used

Low risk. just unknown configs

## How was this PR tested?
both with and without agentOnly true

NA


ameijer (approved):
net -2 loc!

PR #3247, opened 2024-03-19
Use array instead of object as tolerations default

jaanjah:
## What does this PR change?

Changes `values.yaml` `tolerations` properties from `{}` to `[]`

No

Fixes warnings when trying to define tolerations:

```
coalesce.go:286: warning: cannot overwrite table with non table for kubecost.cost-analyzer.forecasting.tolerations (map[])
```

Not that I'm aware of

Shouldn't be any risks

## How was this PR tested?

Created `values-test.yaml` with content:

```
forecasting:
  tolerations:
    - key: "test"
      operator: "Equal"
      value: "true"
      effect: "NoSchedule"
```

and ran:

```
> helm template . -f values.yaml -f values-test.yaml > output.yaml                                                            
```

I don't think there is anything to update?

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -665,7 +665,7 @@ etlUtils:
   resources: {}
   env: {}
   nodeSelector: {}
-  tolerations: {}
+  tolerations: []
   affinity: {}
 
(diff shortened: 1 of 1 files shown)


JesseGoodier (approved):
Thank you for fixing this!

PR #3249, opened 2024-03-19
containerStats on by default

JesseGoodier:
## What does this PR change?
Turn on containerStats by default

No

containerStats pipeline is now enabled by default, can be manually disabled. See more detail:  <https://docs.kubecost.com/architecture/containerstats-pipeline>

None, long-term value that is already frequently enabled. Just need to communicate change in release notes.

## How was this PR tested?
Countless environments have this one for many versions

Brett, can you quickly update this to refelct the fact that it is now on by default? <https://docs.kubecost.com/architecture/containerstats-pipeline>


michaelmdresser (approved):

brstuder:
Good thing I'm checking these for the release notes, will get this done

PR #3252, opened 2024-03-20
Fix Kubecost UI stuck at 'Oh No!' for SAML users after upgrade

saweber:
## What does this PR change?
Force aggregator and cost model pods to restart during a helm upgrade

The problem I have been able to reproduce is caused by the following order of operations:
1) helm upgrade initiated
2) There is no update to cost analyzer
3) The saml secret gets regenerated and updated
4) aggregator is redeployed, with the updated secret mounted.

(The reverse is also possible, where aggregator is not updated, and cost analyzer is.)

The secret used to sign SAML tokens no longer match, therefore the cost analyzer can no longer decrypt tokens issued by aggregator. The easiest solution here is to force a restart of both pods to keep the secret synchronized.

If accepted, this will be cherry picked for 2.1 and 2.2.

no

Resolves issue for SAML users after upgrade where the UI is stuck on the 'Oh No!' page. 

https://kubecost.atlassian.net/browse/GTM-236

Minimal

## How was this PR tested?
Manually

None required

### DIFF
--- a/cost-analyzer/templates/aggregator-statefulset.yaml
+++ b/cost-analyzer/templates/aggregator-statefulset.yaml
@@ -44,6 +44,8 @@ spec:
       labels:
         app.kubernetes.io/name: aggregator
         app.kubernetes.io/instance: {{ .Release.Name }}
+        {{/* Force pod restarts on upgrades to ensure the nginx config is current */}}
+        helm-rollout-restarter: {{ randAlphaNum 5 | quote }}
         app: aggregator
         {{- with .Values.global.additionalLabels }}
         {{- toYaml . | nindent 8 }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -37,6 +37,8 @@ spec:
     metadata:
       labels:
         {{- include "cost-analyzer.selectorLabels" . | nindent 8 }}
+        {{/* Force pod restarts on upgrades to ensure the nginx config is current */}}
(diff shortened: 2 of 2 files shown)


cost-analyzer/templates/aggregator-statefulset.yaml
```diff
@@ -44,6 +44,8 @@ spec:
       labels:
         app.kubernetes.io/name: aggregator
         app.kubernetes.io/instance: {{ .Release.Name }}
+        {{/* Force pod restarts on upgrades to ensure the nginx config is current */}}
```
jessegoodier, line 47:
```suggestion
        {{/* Force pod restarts on upgrades to ensure configmaps are in sync with cost-model */}}
```

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -37,6 +37,8 @@ spec:
     metadata:
       labels:
         {{- include "cost-analyzer.selectorLabels" . | nindent 8 }}
+        {{/* Force pod restarts on upgrades to ensure the nginx config is current */}}
```
jessegoodier, line 40:
```suggestion
        {{/* Force pod restarts on upgrades to ensure configmaps are in sync with aggregator */}}
```

JesseGoodier (approved):

cost-analyzer/templates/aggregator-statefulset.yaml
saweber: What I get for blindly copy and pasting! 🤦

PR #3265, opened 2024-03-23
allow future carbonEstimates configuration

JesseGoodier:
## What does this PR change?
Allows for future options on carbonEstimates

No

NA, unpublished feature

NA, unpublished feature

## How was this PR tested?
upgraded nightly with:

```yaml
kubecostProductConfigs:
  carbonEstimates:
    enabled: true
```

@bstuder99 FYI on helm value change.


jessegoodier:
@kaelanspatel my thinking here is the units may need to be configurable in the future. If you disagree with this approach, I am by no means married to it :)

Issue #3261, opened 2024-03-22
[Feature] Reference the 'extraVolumeMounts' value in the 'aggregator' container spec


ravnalexquinn:
### Problem Statement

In the 'cost-model' container there is the option to pass in a productkey secret and mount it, using a combination of the 'extraVolumes' and 'extraVolumeMounts' values. This lets you take advantage of an external secret provider to pass in the productkey.

The aggregator container does not allow referencing of an 'extraVolumeMounts' value, and so when the container starts up it is unable to read the productkey file containing the required string (because it can't be mounted) and produces this error:
```
Error: initializing: failed to check license: failed to get product key: readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
```

### Solution Description

Update the 'aggregator.containerTemplate' helper function to reference either the existing top-level cost-analyzer.extraVolumeMounts value in the same way as in the 'cost-model' container, or create a new 'cost-analyzer.kubecostAggregator.extraVolumeMounts' value so that users can then mount a secret into the aggregator container.

### Troubleshooting

- [X] I have read and followed the [issue guidelines](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/ISSUE_GUIDELINES.md) and this is a feature request only for the Helm chart.
- [X] I have searched other issues in this repository and mine is not recorded.

dwbrown2:
This seems reasonable and is likely pretty quick. @thomasvn any thoughts?

thomasvn:
@ravnalexquinn I've quickly drafted what I think would be the solution here: https://github.com/kubecost/cost-analyzer-helm-chart/pull/3292! It still requires some testing and documenting which I will do soon. Let me know what you think!

Kavinraja-Ganesan:
@dwbrown2 @thomasvn  -- Can this `extraVolumes`, `extraVolumeMounts`, `extraEnvs` be included for CloudCost & Aggregator pods as well?

Previously we had it for federator pod but this is now separated and those references were somehow removed :(

We mainly use these in combination with custom ServiceAccount for [WLI in GKE](https://cloud.google.com/anthos/clusters/docs/multi-cloud/aws/how-to/use-workload-identity-aws#deploy_a_sample_application).

thomasvn:
[...]

PR #3292, opened 2024-03-29
Aggregator extraVolumes and extraVolumeMounts

thomasvn:
## What does this PR change?

- Allows users to add `extraVolumes` and `extraVolumeMounts` to the Aggregator pod and CloudCost pod, when Aggregator is running as a Statefulset.
- Cleanup values.yaml file under `kubecostAggregator`

No

As titled

Implements https://github.com/kubecost/cost-analyzer-helm-chart/issues/3261

None. This is an optional config.

## How was this PR tested?

```yaml
kubecostAggregator:
  deployMethod: "statefulset"
  extraVolumes:
    - name: test1
      secret:
        secretName: test1-secret
  extraVolumeMounts:
    - name: test1
      mountPath: /var/configs/test1
  cloudCost:
    extraVolumes:
      - name: test2
        configMap:
          name: test2-configmap
    extraVolumeMounts:
      - name: test2
        mountPath: /var/configs/test2
kubecostModel:
  federatedStorageConfigSecret: "federated-store"
```

```sh
helm template ./cost-analyzer -f values.yaml
```

Validated that the `test1` and `test2` volumes and volumemounts were templated as expected.

Documented via values.yaml

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -2457,10 +2457,10 @@ kubecostAggregator:
     "DB_CONCURRENT_INGESTION_COUNT": "3"
 
   persistentConfigsStorage:
-    storageClass: "" # default storage class
+    storageClass: ""  # default storage class
     storageRequest: 1Gi
   aggregatorDbStorage:
-    storageClass: "" # default storage class
+    storageClass: ""  # default storage class
     storageRequest: 128Gi
 
   resources: {}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -2450,23 +2450,23 @@ kubecostAggregator:
   # after consulting with Kubecost's support team
   numDBCopyPartitions: 25
 
-  resources: {}
-    # requests:
-    #   cpu: 1000m
-    #   memory: 1Gi
   env:
     "LOG_LEVEL": "info"
     "DB_READ_THREADS": "1"
     "DB_WRITE_THREADS": "1"
     "DB_CONCURRENT_INGESTION_COUNT": "3"
+
   persistentConfigsStorage:
-    # default storage class
-    storageClass: ""
+    storageClass: "" # default storage class
     storageRequest: 1Gi
   aggregatorDbStorage:
-    # default storage class
-    storageClass: ""
+    storageClass: "" # default storage class
     storageRequest: 128Gi
+
+  resources: {}
+    # requests:
+    #   cpu: 1000m
+    #   memory: 1Gi
 
   readinessProbe:
     enabled: true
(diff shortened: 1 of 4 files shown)


cost-analyzer/values.yaml
```diff
@@ -2450,64 +2450,50 @@ kubecostAggregator:
...
-  #   runAsNonRoot: true
-  # containerSecurityContext:
-  #   allowPrivilegeEscalation: false
-  #   readOnlyRootFilesystem: true
-  #   runAsNonRoot: true
-  #   seccompProfile:
-  #     type: RuntimeDefault
-  #   capabilities:
-  #     drop:
-  #       - ALL
```
thomasvn, line 2488: Deleting this comment because we already have `securityContext` and `containerSecurityContext` commented below.

thomasvn:
@ravnalexquinn @Kavinraja-Ganesan let me know what you think!

Kavinraja-Ganesan:
@thomasvn Can we please include `extraEnvs` as well? Others lg2m!

thomasvn:
@Kavinraja-Ganesan `extraEnvs` for Aggregator pod and CloudCost pod should already be available! It would look something like this:

```yaml
kubecostAggregator:
  env:
    "key": "value"
  cloudCost:
    env:
      "key2": "value2"
```

Kavinraja-Ganesan:
Thanks @thomasvn! Subscribed and waiting for this release.

Kavinraja-Ganesan:
Can this be included in v2.2.1? (if possible)

JesseGoodier (approved):

ravnalexquinn:
Hi @thomasvn, thanks for this

Regarding this:
`Allows users to add extraVolumes [...] when Aggregator is running as a Statefulset.`

Could you explain why the aggregator running as a stateful set is a requirement, please? Preventing the mount values being referenced when aggregator is a container running under the 'cost-analyzer' pod

thomasvn:
@ravnalexquinn The original thinking, was that when all containers are running under the `cost-analyzer` pod, they all reference the `persistent-configs` volumeMount. ([here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/7b6e654d96390fefd72dbda75469541a7f524ab7/cost-analyzer/templates/_helpers.tpl#L913-L915) for aggregator, [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/7b6e654d96390fefd72dbda75469541a7f524ab7/cost-analyzer/templates/cost-analyzer-deployment-template.yaml#L553-L554) for cost-model).

In other words, when Aggregator is run in the `cost-analyzer` pod, both the Aggregator and Cost-model have read write access to the `/var/configs` directory. Therefore adding an `extraVolumeMount` to the cost-analyzer (so long as its under `/var/configs`) should also be accessible to the Aggregator. Does that fit your use case here?

PR #3305, opened 2024-04-03
Changes to add security context for cloud-cost

mclean0328:
## What does this PR change?
Adds security context values for cloud-costs when openshift is enabled

No

Only users deploying to OpenShift 

## How was this PR tested?
Helm templates

No

### DIFF
--- a/cost-analyzer/templates/aggregator-cloud-cost-deployment.yaml
+++ b/cost-analyzer/templates/aggregator-cloud-cost-deployment.yaml
@@ -44,6 +44,18 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
+      {{- if .Values.global.platforms.openshift.enabled }}
+      securityContext:
+      {{- toYaml .Values.global.platforms.openshift.securityContext | nindent 8 }}
+      {{- else if .Values.global.securityContext }}
+      securityContext:
+      {{- toYaml .Values.global.securityContext | nindent 8 }}
+      {{- else }}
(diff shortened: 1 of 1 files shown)


thomasvn:
Awesome thanks @mclean0328. Could you quickly document in this PR what `helm template` command you ran? And the successful result you received on the cloud-cost deployment?

thomasvn:
Also, it should be noted that this PR will affect _all_ users of the Cloud Cost pod in this upcoming release, as the new default securityContext we're applying is:

```yaml
  spec:
      securityContext:
        runAsUser: 1001
        runAsGroup: 1001
        fsGroup: 1001
```

Therefore it may also be worthwhile quickly deploying the pod and ensuring CloudCosts correctly runs against AWS, Azure, GCP.

jessegoodier:
> Also, it should be noted that this PR will affect _all_ users of the Cloud Cost pod in this upcoming release, as the new default securityContext we're applying is:
> 
> ```yaml
>   spec:
>       securityContext:
>         runAsUser: 1001
>         runAsGroup: 1001
>         fsGroup: 1001
> ```
> 
> Therefore it may also be worthwhile quickly deploying the pod and ensuring CloudCosts correctly runs against AWS, Azure, GCP.

seems reasonable. 1001 is our default user. simple note good enough?

mclean0328:
I was able to test of Azure and everything came up and the cloud-cost pod came up and cloud integration worked correctly.

JesseGoodier (approved):

PR #3318, opened 2024-04-05
Adds consistency around how we are templating `.Values.global.additionalLabels`

thomasvn:
## What does this PR change?

Templates `.Values.global.additionalLabels` on all Deployments, Statefulsets, Daemonsets, and their pod templates. Previously, it was missing in some of these manifests. Or it was only being applied in the Pod template labels and not the Deployment labels. Or vice versa.

I'm open to feedback as to whether this is the right approach or not. The Helm configuration itself is vague, so I've included a comment stating what it now intends to do.

N/A

Previously `.Values.global.additionalLabels` was inconsistent in how it would apply these additional labels to specific manifests. Although we are now being consistent in templating, this is a potential breaking change for users whose configs only worked in the previous inconsistent templating

N/A

Risks stated above.

## How was this PR tested?

```yaml
# values.yaml
global:
  additionalLabels:
    owner: thomasn
```

```sh
helm template cost-analyzer -f values.yaml | less
```

Validated that I could see the `owner: thomasn` label in all Deployments (and their pod templates) created.

Documentation in `values.yaml`.

### DIFF
--- a/cost-analyzer/templates/NOTES.txt
+++ b/cost-analyzer/templates/NOTES.txt
@@ -11,11 +11,7 @@
 {{- $servicePort := .Values.service.port | default 9090 }}
 Kubecost {{ .Chart.Version }} has been successfully installed.
 
-Welcome to Kubecost 2.0!
-
-Kubecost 2.0 is a major upgrade from previous versions and includes major new features including a brand new API Backend. Please review the following documentation to ensure a smooth transition: https://docs.kubecost.com/install-and-configure/install/kubecostv2
-
-For the full list of enhancements, please see our release notes: https://github.com/kubecost/cost-analyzer-helm-chart/releases/tag/v2.0.0
+Kubecost 2.x is a major upgrade from previous versions and includes major new features including a brand new API Backend. Please review the following documentation to ensure a smooth transition: https://docs.kubecost.com/install-and-configure/install/kubecostv2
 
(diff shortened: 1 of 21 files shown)


thomasvn:
@chipzoller Let me know if you have any thoughts on this from a UX perspective!

chipzoller:
I haven't inspected the contents of this change, but saying "global" means to truly apply to everything. To not have done so seems like obvious oversights to me. If this change makes "global" apply to everything then that's correct.

jessegoodier:
> I haven't inspected the contents of this change, but saying "global" means to truly apply to everything. To not have done so seems like obvious oversights to me. If this change makes "global" apply to everything then that's correct.

which pods are missing?

jessegoodier:
I'll test this, and fix the merge for something that will conflict.

chipzoller:
> > I haven't inspected the contents of this change, but saying "global" means to truly apply to everything. To not have done so seems like obvious oversights to me. If this change makes "global" apply to everything then that's correct.
> 
> which pods are missing?

Don't know personally.

JesseGoodier (approved):
Tested, works!

thomasvn:
I've tested as well. For now, it's documented that we are only adding labels to Deployments, Statefulsets, Daemonsets, and their pod templates. In the future, we may need to add the labels to other resources such as Services, Ingresses, ServiceAccounts, etc.

PR #3330, opened 2024-04-08
quick fix for athena cc integration on single pod deployments

Sean-Holcomb:
## What does this PR change?
quick fix for Athena cc integration on single pod deployments when using helm values. v2.2 has an issue where the cost-analyzer pod runs the cloud-cost integration controller within the Opencost Cost model. On a single pod deployment this causes issues when the controller on one of the containers has access to the secret that is being mounted by the helm chart and not the other, causing them to save conflicting configurations to their backing file on the shared pv. The long term solution is to remove CloudCost from Open Cost cost model which there are PR's for, but the change is too big for this stage in the 2.2 release cycle. This change ensures that when configuring Athena cloud cost with helm values the secret that is created is accessible by the controllers in both containers, which prevents conflict between the two.

## How was this PR tested?
Test on deployment created with helm values set

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -588,7 +588,8 @@ spec:
             - name: azure-storage-config
               mountPath: /var/azure-storage-config
             {{- end }}
-            {{- if or (.Values.kubecostProductConfigs.cloudIntegrationSecret) (.Values.kubecostProductConfigs.cloudIntegrationJSON) }}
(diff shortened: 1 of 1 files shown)


michaelmdresser (commented):
> controller on one of the containers has access to the secret that is being mounted by the helm chart and not the other, causing them to save conflicting configurations to their backing file on the shared pv

Yikes!

What do you think about just merging this into the v2.2 branch and pursuing the proper fix in `develop`?

Sean-Holcomb:
@michaelmdresser sure lets do that, I will put up a new PR that removes this block entirely once the other changes are merged

PR #3333, opened 2024-04-10
Remove cloud integration secret mount from cost-model container

Sean-Holcomb:
## What does this PR change?
Cloud Integration Secret should not be mounted to this container as the processes that uses it should not be running within it.

This PR should not be merged until these PR's are in
https://github.com/kubecost/kubecost-cost-model/pull/2363
https://github.com/opencost/opencost/pull/2691


chipzoller:
Is this still relevant?

nikovacevic (approved):

PR #3334, opened 2024-04-10
add READ_ONLY to aggregator

JesseGoodier:
## What does this PR change?
add missing read_only env var

no

Fix: simple read_only UI setting that was broken in 2.0 with the introduction of aggregator

None

## How was this PR tested?
Added value and upgrade RO env.

add note to values file

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -3108,7 +3108,8 @@ costEventsAudit:
   enabled: false
 
 ## Disable updates to kubecost from the frontend UI and via POST request
-##
+## This feature is considered beta, entrprise users should use teams:
+## https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/teams
 # readonly: false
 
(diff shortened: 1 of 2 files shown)


chipzoller (approved):

cost-analyzer/values.yaml
```diff
@@ -3108,7 +3108,8 @@ costEventsAudit:
   enabled: false
 
 ## Disable updates to kubecost from the frontend UI and via POST request
-##
+## This feature is considered beta, entrprise users should use teams:
```
thomasvn, line 3111: Not sure it's right to call a long-standing feature "beta"? Maybe comment should be changed to:

> Enterprise user are recommended to use the Teams feature:

jessegoodier: Good point. but I think it can be promoted when the UI 1. tells the user it is read only. 2. UI doesn't throw errors when trying to write settings.

PR #3345, opened 2024-04-15
allow multi-cluster prometheus

JesseGoodier:
## What does this PR change?
allow allow multi-cluster prometheus configs / without warnings

No

allow allow multi-cluster prometheus configs, supported as of Kubecost 2.2.0

None

## How was this PR tested?
Helm upgrade of environment with AMP

Already done.


thomasvn (approved):
Multi-cluster Prometheus supported as of v2.2.0

PR #3348, opened 2024-04-15
clusterIDCheck

JesseGoodier:
## What does this PR change?
add check for cluster_id when using federated storage

no

add check for cluster_id when using federated storage

None,  should have done this long ago.

## How was this PR tested?
few tests in qa environments.

NA


thomasvn:
This is good! Although a better alternative would be to use `.Values.kubecostProductConfigs.clusterName` as the source of truth across the repo.

jessegoodier:
> This is good! Although a better alternative would be to use `.Values.kubecostProductConfigs.clusterName` as the source of truth across the repo.

100%. 
Kubecost 3.0 though?

thomasvn (approved):

PR #3367, opened 2024-04-25
Add custom SMTP support to chart

nik-kc:
## What does this PR change?
* Enables the associated KCM changes, allowing for a configurable SMTP server.

* [KCM](https://github.com/kubecost/kubecost-cost-model/pull/2397).

* Users will now be able to configure a custom SMTP server through which to send alert, report, etc. email.

* Minimal risks involved; features are being added, not modified.

## How was this PR tested?
* Validated that SMTP configs are populated as expected via all the various configuration methods:
  * Helm config map.
  * Helm secret.
  * API.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
+++ b/cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.kubecostProductConfigs }}
+{{- if .Values.kubecostProductConfigs.smtp }}
+{{- if .Values.kubecostProductConfigs.smtp.enabled }}
+# If the smtp.config is not specified, the configmap will not be created
+{{- if .Values.kubecostProductConfigs.smtp.config }}
+# If the secretname is specified, the configmap will not be created
+{{- if not .Values.kubecostProductConfigs.smtp.secretname }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ default "smtp-configs" .Values.smtpConfigmapName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+data:
+  {{- if .Values.kubecostProductConfigs.smtp.config }}
+  config: {{  .Values.kubecostProductConfigs.smtp.config | quote }}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
+++ b/cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
@@ -1,6 +1,4 @@
-{{- if .Values.kubecostProductConfigs }}
-{{- if .Values.kubecostProductConfigs.smtp }}
-{{- if .Values.kubecostProductConfigs.smtp.enabled }}
+{{- if (((.Values.kubecostProductConfigs).smtp).enabled) }}
 # If the smtp.config is not specified, the configmap will not be created
 {{- if .Values.kubecostProductConfigs.smtp.config }}
 # If the secretname is specified, the configmap will not be created
@@ -11,13 +9,11 @@ metadata:
   name: {{ default "smtp-configs" .Values.smtpConfigmapName }}
   namespace: {{ .Release.Namespace }}
   labels:
-    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
 data:
   {{- if .Values.kubecostProductConfigs.smtp.config }}
   config: {{  .Values.kubecostProductConfigs.smtp.config | quote }}
   {{- end -}}
-{{- end -}}
-{{- end -}}
 {{- end -}}
 {{- end -}}
 {{- end -}}
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -3272,6 +3272,11 @@ costEventsAudit:
 #     key: ""
 #     secretname: productkeysecret # Reference an existing k8s secret created from a file named productkey.json of format { "key": "enterprise-key-here" }. If the secretname is specified, a configmap with the key will not be created.
 #     mountPath: "/some/custom/path/productkey.json" # (use instead of secretname) Declare the path at which the product key file is mounted (eg. by a secrets provisioner). The file must be of format { "key": "enterprise-key-here" }.
+#   smtp:
+#     enabled: false
+#     config: ""
+#     secretname: smtpconfigsecret
+#     mountPath: "/some/custom/path/smtp.json"
 #   carbonEstimates: false # Enables Kubecost beta carbon estimation endpoints /assets/carbon and /allocations/carbon
 
   ## Specify an existing Kubernetes Secret holding the cloud integration information. This Secret must contain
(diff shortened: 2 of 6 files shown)


cost-analyzer/values.yaml
```diff
@@ -3272,6 +3272,11 @@ costEventsAudit:
 #     key: ""
 #     secretname: productkeysecret # Reference an existing k8s secret created from a file named productkey.json of format { "key": "enterprise-key-here" }. If the secretname is specified, a configmap with the key will not be created.
 #     mountPath: "/some/custom/path/productkey.json" # (use instead of secretname) Declare the path at which the product key file is mounted (eg. by a secrets provisioner). The file must be of format { "key": "enterprise-key-here" }.
+#   smtp:
+#     enabled: false
+#     config: ""
+#     secretname: smtpconfigsecret
+#     mountPath: "/some/custom/path/smtp.json"
```
michaelmdresser: Why do we need a variable mount path?

nik-kc: For the entirety of this PR, I followed the example set by `productkey`. Based on the comments laid out in `productkey`, variable mount paths are used to enable secrets provisioners.

Now, I'd rate my personal experience with secrets provisioners at around zero out of ten, but given the context of what we're working with, I'd say that's possibly somewhat reasonable.

michaelmdresser: Ah, same with the experience. I'll trust those comments! Thanks for the explanation

michaelmdresser (approved):
LGTM but needs to be tested

cost-analyzer/values.yaml
chipzoller: Let's also add a comment above this to clearly explain what this block does (enables use of a custom SMTP server which trumps Kubecost's built-in, external SMTP server).

chipzoller: And what's in `config`?

cost-analyzer/values.yaml
```diff
@@ -3272,6 +3272,11 @@ costEventsAudit:
 #     key: ""
 #     secretname: productkeysecret # Reference an existing k8s secret created from a file named productkey.json of format { "key": "enterprise-key-here" }. If the secretname is specified, a configmap with the key will not be created.
 #     mountPath: "/some/custom/path/productkey.json" # (use instead of secretname) Declare the path at which the product key file is mounted (eg. by a secrets provisioner). The file must be of format { "key": "enterprise-key-here" }.
+#   smtp:
+#     enabled: false
+#     config: ""
+#     secretname: smtpconfigsecret
```
chipzoller: What's the format of this secret?

cost-analyzer/values.yaml
nik-kc: @chipzoller Very valid points; I'll make that change and clarify what should go in `config`.

cost-analyzer/values.yaml
nik-kc: I'll add some clarification here.

cost-analyzer/values.yaml
```diff
@@ -3272,6 +3272,21 @@ costEventsAudit:
...
+#     enabled: false
+#     config:  |
+#       {
+#         "sender_email": "",
+#         "host": "",
+#         "port": 587,
+#         "authentication": true,
+#         "username": "",
+#         "password": "",
+#         "secure": true
```
chipzoller, line 3286: Has a custom cert as an input value been implemented in the code?

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.kubecostProductConfigs }}
+{{- if .Values.kubecostProductConfigs.smtp }}
+{{- if .Values.kubecostProductConfigs.smtp.enabled }}
```
chipzoller: These can be condensed into a single statement.

cost-analyzer/values.yaml
nik-kc: [It has indeed](https://github.com/kubecost/kubecost-cost-model/blob/f1ff0fafec709d9de6d7a5c55f79548bacb26525/pkg/messaging/smtp/config.go#L15).

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
nik-kc: I was following the example laid out [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-pkey-configmap.yaml). How would you condense these conditionals? I'm not intimately familiar with template conditionals, yet.

cost-analyzer/values.yaml
chipzoller: Cool. Is that also an allowable value here? If so, can you update and maybe put a (truncated) example for users to follow?

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+# If the smtp.config is not specified, the configmap will not be created
+{{- if .Values.kubecostProductConfigs.smtp.config }}
+# If the secretname is specified, the configmap will not be created
+{{- if not .Values.kubecostProductConfigs.smtp.secretname }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ default "smtp-configs" .Values.smtpConfigmapName }}
```
chipzoller, line 9: `smtpConfigmapName` is not defined anywhere.

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
```
chipzoller:
```suggestion
    {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
```

When you include templates, you still need to chomp new lines at the beginning.

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+data:
+  {{- if .Values.kubecostProductConfigs.smtp.config }}
+  config: {{  .Values.kubecostProductConfigs.smtp.config | quote }}
```
chipzoller, line 15: Have you seen the value which results from this? It's not stringified JSON but stringified block scalar YAML since the input value is a block scalar. You may want to use a simple map for the values and then convert to JSON in the value or just use a YAML map for the ConfigMap portion.

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
chipzoller:
```
{{- if (((.Values.kubecostProductConfigs).smtp).enabled) }}
```
And remove two `{{- end -}}` statements from the bottom.

chipzoller (changes requested):

cost-analyzer/values.yaml
nik-kc: It is an allowable value, but certificates have not yet been tested. Given the big push to get this into nightly, I'll be focusing on certificates once everyone has access to this feature in nightly.

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
nik-kc: I was following the example given by `productConfigmapName` which, interestingly, also isn't defined anywhere. I'll get both of those added.

cost-analyzer/templates/cost-analyzer-smtp-configmap.yaml
nik-kc: I have, and it functions as expected when decoding the string as JSON. I trust your insight, so if you feel it should change, I'll change it. Just let me know.

chipzoller (approved):
Nice work, let's get this in and tested.

chipzoller:
Seeing if @jessegoodier has any other requests here.

PR #3383, opened 2024-05-02
networkCosts-on-by-default

JesseGoodier:
## What does this PR change?
turn networkCosts on by default

No

Enables network insights for all without having to find the settings to enable it

This has been used by hundreds of users for years. Some may not want this, but it is easily disabled. 

## How was this PR tested?
local helm installs

@bstuder99 could you help with this please?

We need to modify networkcosts doc to show how to disable the DS and or cloud provider support by settings these to false.

```yaml
networkCosts:
  enabled: true
  config:
    services:
      google-cloud-services: true
      amazon-web-services: true
      azure-cloud-services: true
```

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -2235,7 +2235,7 @@ prometheus:
 ## Module for measuring network costs
 ## Ref: https://github.com/kubecost/docs/blob/main/network-allocation.md
 networkCosts:
-  enabled: false
+  enabled: true
   image:
     repository: gcr.io/kubecost1/kubecost-network-costs
     tag: v0.17.3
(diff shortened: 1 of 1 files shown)


srpomeroy (approved):

dwbrown2:
Aiming to turn this all for all users? Any more context on risks that we see? The main reason we disabled by default was due to resource consumption...

Any other docs we need to update in the process?

AjayTripathy:
Yeah, I'm not sure about turning on by default given scaling considerations...

srpomeroy:
At a minimum we should enable the service config for all three providers even if the DS itself is set to disabled by default.

jessegoodier:
> Yeah, I'm not sure about turning on by default given scaling considerations...

we have a .5 cpu limit set by default. 
Are thinking this is still a risk? Are you also concerned with memory?

```yaml
  resources:
    limits:  # remove the limits by setting cpu: null
      cpu: 500m
```

jessegoodier:
> Aiming to turn this all for all users? Any more context on risks that we see? The main reason we disabled by default was due to resource consumption...
> 
> Any other docs we need to update in the process?

We could add to the Helm upgrade Note also.
any other comms you have in mind?

jessegoodier:
for the record, I'm not married to this change. That said, there is value to it.

dwbrown2:
cc @kwombach12 @chipzoller for other input

kwombach12:
I would like to hear more about the resource consumption concern -- i.e. Does this not scale well? What % of users do we believe have this off due to resource consumption? 

Regardless of whether we enable or disable network costs by default, I feel strongly that, if off by default, it should be a matter of enabling just one flag to turn it on... today it is 2-4...

jessegoodier:
After further thought, I believe off by default is preferable. Will create a new PR

PR #3392, opened 2024-05-07
Change imagePullPolicy to `IfNotPresent`

chipzoller:

## What does this PR change?

Changes the imagePullPolicy where previously set to `Always` to now adopt `IfNotPresent` to be consistent across the chart.

No

Only user-facing impact this may have is to improve startup time when Pods are disrupted and, related, to slightly reduce network traffic since the same images are not re-pulled. The other benefit of this change is Kubecost in that we should see, over time, less egress traffic as fewer image pulls occur.

N/A

Risks should be almost none to users. Since image tags are employed throughout the chart, installation and upgrade should result in new tags which would always be pulled.

## How was this PR tested?

Installed a minimal version locally and verified all Pods' imagePullPolicy fields were set to `IfNotPresent`.

N/A


cost-analyzer/templates/_helpers.tpl
```diff
@@ -902,7 +902,7 @@ Begin Kubecost 2.0 templates
     periodSeconds: {{ .Values.kubecostAggregator.readinessProbe.periodSeconds }}
     failureThreshold: {{ .Values.kubecostAggregator.readinessProbe.failureThreshold }}
   {{- end }}
-  imagePullPolicy: Always
+  imagePullPolicy: IfNotPresent
```
thomasvn: Should we use this as an opportunity to template the `imagePullPolicy` here on the Aggregator and also the Forecasting image? Overall this looks good! Just want to make sure that if we change the current default behavior the user has the option to override.

chipzoller: I knew you were going to ask that :) I agree, but I'd like some help with that from someone else if possible.

thomasvn: Haha sure thing! I'll try to commit those changes onto this PR later today.

jessegoodier: There's money to be made in an AI specific helm templater :)

thomasvn:
I've made an update so that `imagePullPolicy` is templated for Aggregator, CloudCost, and Forecasting images. Default is `IfNotPresent`

```txt
helm template ./cost-analyzer | grep -i "imagePullPolicy"
        imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: IfNotPresent
          imagePullPolicy: "IfNotPresent"
```

thomasvn (approved):

chipzoller:
Bless you 🙏

AjayTripathy:
LGTM -- in favor of improved startup time! Opencost does this as well.

PR #3399, opened 2024-05-09
increase default data build to 731d from 91d

JesseGoodier:
## What does this PR change?
increase default retention to 731d from 91d. The previous value was low to save resources with the previous Kubecost architecture. With aggregator, there is no longer a need for this to be 91d. 

No

Changed default Kubecost retention to 731d from 91d

Many users already do this. Slightly more memory usage over time.

## How was this PR tested?
local qa envs

@bstuder99 please take a look for the retention defaults and update?

to use the previous default, set:
```yaml
kubecostModel:
  etlDailyStoreDurationDays: 91
```

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1026,7 +1026,7 @@ Begin Kubecost 2.0 templates
       value: {{ .Values.smtpConfigmapName }}
     {{- end }}
     - name: ETL_DAILY_STORE_DURATION_DAYS
-      value: {{ (quote .Values.kubecostModel.etlDailyStoreDurationDays) | default (quote 731) }}
+      value: {{ (quote .Values.kubecostModel.etlDailyStoreDurationDays) }}
     {{- if (gt (int .Values.kubecostAggregator.numDBCopyPartitions) 0) }}
     - name: NUM_DB_COPY_CHUNKS
       value: {{ .Values.kubecostAggregator.numDBCopyPartitions | quote }}
@@ -1260,7 +1260,7 @@ Begin Kubecost 2.0 templates
       value: "true"
     {{- end}}
     - name: ETL_DAILY_STORE_DURATION_DAYS
-      value: {{ (quote .Values.kubecostModel.etlDailyStoreDurationDays) | default (quote 731) }}
+      value: {{ (quote .Values.kubecostModel.etlDailyStoreDurationDays) }}
     - name: CLOUD_COST_REFRESH_RATE_HOURS
       value: {{ .Values.kubecostAggregator.cloudCost.refreshRateHours | default 6 | quote }}
     - name: CLOUD_COST_QUERY_WINDOW_DAYS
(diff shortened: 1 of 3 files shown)


ameijer (approved):

ameijer (approved):

brstuder:
Is this release related? I don't see any tags or mentions in this PR?

thomasvn:
@jessegoodier @ameijer What is the intended effect here? Updating it under `kubecostModel` will only affect how many days of ETL each agent will build. This doesn't affect aggregator's retention.

jessegoodier:
This is what we needed to pull in older data.

dwbrown2:
cc @AjayTripathy @kwombach12 @chipzoller for any other input!

chipzoller:
Does this change how much data will be stored in KCM's PV?

thomasvn:
My bad here. Re-reading the diff I'm seeing that you updated `ETL_DAILY_STORE_DURATION_DAYS` in three places. 1) Aggregator, 2) CloudCosts, 3) CostAnalyzer. This is correct! For CostAnalyzer + CloudCosts, this envvar determines how far back to build ETL. For Aggregator this envvar determines how long to retain data before rolling that data off.

It's a little funky to have all those knobs be tuned by the single `.Values.kubecostModel.etlDailyStoreDurationDays`. But for now I think that's the simplest solution and simplest UX. We may want to revisit this someday.

jessegoodier:
> Does this change how much data will be stored in KCM's PV?

Biggest impact is not needing to set this for rebuilding historical data. 

PV sizing is a consideration when rebuilding. It would be nice to have a guide to recommend GB needed per GB of ETL.
That's outside the scope of this change.

kwombach12:
im hugely supportive of this move -- I think there a huge benefits to this for forecasting if we think it can scale to handle 731 days of data - the model would be much better. cc: @cliffcolvin

AjayTripathy:
Just double checking. Doesn't the initial derivation step need to go back the full window? No further concerns if that's not true.

thomasvn:
Looking at the nightly environments, the cloudcosts pod is attempting to rebuild data going back to 2022 upon restart. I think there are severe implications here and that's not something we want to do in any environment of scale.

I like the idea of increasing Aggregator's default retention, but can we do it by changing the following config instead?

```yaml
kubecostAggregator:
  env:
    ETL_DAILY_STORE_DURATION_DAYS: "365"
```

That way we are not asking the `cost-model` to attempt to build & push two years worth of daily ETL files. And we are not asking `cloudcost` to fetch 2yrs worth of billing data.

AjayTripathy:
@thomasvn if you feel this has risk I'm supportive of turning this down to 365. 

cc @jessegoodier

jessegoodier:
> @thomasvn if you feel this has risk I'm supportive of turning this down to 365.
> 
> cc @jessegoodier

we may need to put this back to 91d altogether.

PR incoming.

We can hightlight how to bring in the data in release notes.

PR #3401, opened 2024-05-09
Add forecastingEnabled to`/productConfigs` endpoint

thomasvn:
## What does this PR change?

Updates the `/model/productConfigs` endpoint to include information `forecastingEnabled`. Important to note that the existing logic which returns HTTP 405 when forecasting is not enabled has been left in place for backwards compatibility.

N/A

N/A

N/A

N/A. This endpoint is only used by kubecost-frontend. Additional fields will not affect current functionality.

## How was this PR tested?

```sh
# Forecasting is default enabled in the values.yaml file
$ helm template ./cost-analyzer | grep -i -C5 "forecastingEnabled"

                "costEventsAuditEnabled": "false",
                "frontendDeployMethod": "singlepod",
                "pluginsEnabled": "false",
                "carbonEstimatesEnabled": "false",
                "clusterControllerEnabled": "false",
                "forecastingEnabled": "true",
                "chartVersion": "DEVELOP_BRANCH"
                }
            ';
        }
    }

# When disabling forecasting, we return HTTP 405 and also `/productConfigs` returns false.
$ helm template ./cost-analyzer --set forecasting.enabled=false | grep -i -C5 "forecastingEnabled"
            return 200 '{"aggregatorEnabled": "true"}';
        }
        location /forecasting {
            return 405 '{"forecastingEnabled": "false"}';
        }

        location /model/productConfigs {
--
                "costEventsAuditEnabled": "false",
                "frontendDeployMethod": "singlepod",
                "pluginsEnabled": "false",
                "carbonEstimatesEnabled": "false",
                "clusterControllerEnabled": "false",
                "forecastingEnabled": "false",
                "chartVersion": "DEVELOP_BRANCH"
                }
            ';
        }
    }
```

N/A

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -1221,6 +1221,7 @@ data:
             return 405 '{"forecastingEnabled": "false"}';
         }
         {{- end }}
+
         location /model/productConfigs {
             default_type 'application/json';
             add_header 'Access-Control-Allow-Origin' '*' always;
@@ -1235,10 +1236,12 @@ data:
                 "pluginsEnabled": "{{ template "pluginsEnabled" . }}",
                 "carbonEstimatesEnabled": "{{ template "carbonEstimatesEnabled" . }}",
                 "clusterControllerEnabled": "{{ template "clusterControllerEnabled" . }}",
+                "forecastingEnabled": "{{ template "forecastingEnabled" . }}",
                 "chartVersion": "DEVELOP_BRANCH"
                 }
             ';
         }
     }
+
 {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -1221,6 +1221,7 @@ data:
             return 405 '{"forecastingEnabled": "false"}';
```
thomasvn, line 1221: @nealormsbee @rayhatfield I decided to leave this existing functionality in place. This may be a useful HTTP error to fall back on?

nealormsbee: Love it, yes. That's exactly what the endpoint should do if we request it even though the config told us it's disabled :D

nealormsbee (approved):
:100:

PR #3410, opened 2024-05-13
add required flag for upgrade to ack full reingest but not on install

saweber:
## What does this PR change?
Users upgrading to a version through helm starting with 2.3.0 will be required to pass a helm value acknowledging a full re-ingestion may be triggered. Fresh installs will not require the flag (but on an additional helm revision to another 2.3.x version, will be required to pass the flag).

No

Values.upgrade.acknowledgeV2dot3FullReingestion will be required for upgrading users

n/a?

Annoyance - this flag is currently fairly dumb - it will be required on any (non-installation) upgrade for 2.3 bugfixes, even though there is no full reingestion and db migration taking place.

## How was this PR tested?
Manually

Not yet - this PR is just a fast way to collect feedback

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -32,6 +32,18 @@ Set important variables before starting main templates
 {{- end -}}
 
 {{/*
+Kubecost 2.3 precondition that a full re-ingestion will be triggered for aggregator
+*/}}
+{{- define "kubecostV2.3-reingestion-precondition" -}}
+    {{- if semverCompare ">= 2.1.0-0 <2.4.0-0" .Chart.Version }}
+        {{- if gt .Release.Revision 1 -}}
+            {{- if not .Values.upgrade.acknowledgeV2dot3FullReingestion -}}
+                {{- fail "\n\nKubecost 2.3 requires a full reingestion of all data which will could significant time. Please acknowledge by setting 'upgrade.acknowledgeV2dot3FullReingestion' to 'true' in your values file." -}}
+            {{- end -}}
+        {{- end -}}
+    {{- end -}}
+{{- end -}}
+{{/*
 Kubecost 2.0 preconditions
 */}}
 {{- define "kubecostV2-preconditions" -}}
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -35,7 +35,7 @@ Set important variables before starting main templates
 Kubecost 2.3 precondition that a full re-ingestion will be triggered for aggregator
 */}}
 {{- define "kubecostV2.3-reingestion-precondition" -}}
-    {{- if semverCompare ">= 2.1.0-0 <2.4.0-0" .Chart.Version }}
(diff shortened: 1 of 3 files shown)


saweber:
Opening this up for discussion ASAP. I am very open to better wording and a more user friendly way to do this, if there are ideas.

I would prefer to keep track of these migrations so we can be smarter about this - something along the lines of storing the data in a configmap.

cost-analyzer/templates/_helpers.tpl
```diff
@@ -31,6 +31,18 @@ Set important variables before starting main templates
...
 {{- end -}}
 
+{{/*
+Kubecost 2.3 precondition that a full re-ingestion will be triggered for aggregator
+*/}}
+{{- define "kubecostV2.3-reingestion-precondition" -}}
+    {{- if semverCompare ">= 2.1.0-0 <2.4.0-0" .Chart.Version }}
+        {{- if gt .Release.Revision 1 -}}
+            {{- if not .Values.upgrade.acknowledgeV2dot3FullReingestion -}}
+                {{- fail "\n\nKubecost 2.3 requires a full reingestion of all data which will could take significant time. Please acknowledge by setting 'upgrade.acknowledgeV2dot3FullReingestion' to 'true' in your values file." -}}
```
cliffcolvin, line 41: What if we have the value be upgrade.verifyVersion=2.3 and then if we need to use it in the future we use the same field and just increment the value?

cliffcolvin (commented):
Left one comment but I like this approach personally. Will let ajay comment before we approve

cost-analyzer/templates/_helpers.tpl
saweber: We can totally go this route. My concerns with that path are:
1) It is significantly less clear to the customer when they can remove the value from their values file
2) To someone looking at the file 6 months down the road - it is a lot harder to understand what this is, other than something they added in to make the errors go away. A more specific name - `lastAcknowledgedFullReingestionVersion=2.3` could help here.

chipzoller: May want to consider the logic [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/13d4f5bcca8a302a7a2ee153b1b4af17ec6fc2ab/cost-analyzer/templates/_helpers.tpl#L37-L60) if you haven't already.

cost-analyzer/values.yaml
```diff
@@ -304,6 +304,9 @@ global:
 ##
 upgrade:
   toV2: false
+  ## This flag is only required for users upgrading to 2.3 or later from a previous version.
+  ## This is because 2.3 will trigger a full re-ingestion of all available data into aggregator.
+  acknowledgeV2dot3FullReingestion: false
```
chipzoller, line 309: I'd suggest making this value simpler. The whole matter of re-ingestion is an implementation detail. The user-facing point is just "get me to 2.3".

chipzoller (commented):
Additionally, if no delay from a reingestion is to occur from 2.3 to any later version, users shouldn't need to set the flag.

PR #3415, opened 2024-05-16
add fe flag for retention, remove obsolete weekly retention

JesseGoodier:
## What does this PR change?
-Adds FE flag to show what retention to expect for each store
-Recommend disabling hourly data in large environments
-Remove obsolete values for weekly store

No, required for FE enhancement

Kubecost recomends disabling hourly data in large environements- this will double the performance of data ingestion and halve size etl data store in the PV and S3

NA

## How was this PR tested?
Helm template:

```sh
helm template kubecost ./cost-analyzer |ag Retention               
                "hourlyDataRetention": "49",
                "dailyDataRetention": "731",
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -1302,7 +1302,9 @@ data:
                 "carbonEstimatesEnabled": "{{ template "carbonEstimatesEnabled" . }}",
                 "clusterControllerEnabled": "{{ template "clusterControllerEnabled" . }}",
                 "forecastingEnabled": "{{ template "forecastingEnabled" . }}",
-                "chartVersion": "DEVELOP_BRANCH"
+                "chartVersion": "DEVELOP_BRANCH",
+                "hourlyDataRetention": "{{ (.Values.kubecostModel.etlHourlyStoreDurationHours) }}",
+                "dailyDataRetention": "{{ (.Values.kubecostModel.etlDailyStoreDurationDays) }}"
                 }
             ';
         }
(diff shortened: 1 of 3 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -1302,7 +1302,9 @@ data:
                 "carbonEstimatesEnabled": "{{ template "carbonEstimatesEnabled" . }}",
                 "clusterControllerEnabled": "{{ template "clusterControllerEnabled" . }}",
                 "forecastingEnabled": "{{ template "forecastingEnabled" . }}",
-                "chartVersion": "DEVELOP_BRANCH"
+                "chartVersion": "DEVELOP_BRANCH",
+                "hourlyDataRetention": "{{ (.Values.kubecostModel.etlHourlyStoreDurationHours) }}",
+                "dailyDataRetention": "{{ (.Values.kubecostModel.etlDailyStoreDurationDays) }}"
```
thomasvn, line 1307: My only concern here, is that this `.Values.kubecostModel.etlDailyStoreDurationDays` is now an overloaded variable (as mentioned in PR https://github.com/kubecost/cost-analyzer-helm-chart/pull/3399). And making a change like this just digs us deeper into how overloaded the variable has become. (similar to how `.Values.prometheus.server.global.external_labels` is overloaded).

Technically when this env var is set in Aggregator, it does indeed define what the retention is of the database. In cost-model and cloudcost it determines how far back to build/push ETL.

I'm ok approving for now since this endpoint is only used by FE. But I think we should be mindful of changes like this.

thomasvn:
Aside from the one comment I left this all looks good to me! @ameijer Could you give a final look/approval? Also tagging @srpomeroy @williamkubecost for awareness.

thomasvn:
Going to go ahead and approve so that it can go into our next nightly build. @ameijer When you look this over, can you let me know if you have any concerns?

thomasvn (approved):

PR #3417, opened 2024-05-17
revert back to 91d

JesseGoodier:
## What does this PR change?
Reverts https://github.com/kubecost/cost-analyzer-helm-chart/pull/3399

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1025,8 +1025,6 @@ Begin Kubecost 2.0 templates
     - name: SMTP_CONFIGMAP_NAME
       value: {{ .Values.smtpConfigmapName }}
     {{- end }}
-    - name: ETL_DAILY_STORE_DURATION_DAYS
-      value: {{ (quote .Values.kubecostModel.etlDailyStoreDurationDays) }}
     {{- if (gt (int .Values.kubecostAggregator.numDBCopyPartitions) 0) }}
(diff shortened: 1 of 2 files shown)


thomasvn:
Ok this looks good. Notably, this PR changes two things:

1. Do not set Aggregator's environment variables via the `.Values.kubecostModel.etlDailyStoreDurationDays` config. If you want to change Aggregator's `ETL_DAILY_STORE_DURATION_DAYS` which determines database ingestion/retention, that should live as its own config. More documentation [here](https://docs.kubecost.com/install-and-configure/install/multi-cluster/federated-etl/aggregator).
4. Reduce `etlDailyStoreDurationDays` default from `731` back to `91`. Additionally this variable now only has effect on the `costmodel` container and `cloudcost` container, and it determines how far back to query data and build ETL.

thomasvn:
FYI. This has an effect on PR https://github.com/kubecost/cost-analyzer-helm-chart/pull/3415. The following config is no longer representative of what the Aggregator has configured for ingestion/retention.

```yaml
kubecostModel:
  etlDailyStoreDurationDays: ""
```

jessegoodier:
> FYI. This has an effect on PR #3415. The following config is no longer representative of what the Aggregator has configured for ingestion/retention.
> 
> ```yaml
> kubecostModel:
>   etlDailyStoreDurationDays: ""
> ```

we aren't using that yet. really was for the hourly store disabling concern.

thomasvn (approved):

PR #3425, opened 2024-05-21
[CI] Helm upgrade test

chipzoller:

## What does this PR change?

Initializes a Helm upgrade test in CI which runs automatically, every Thursday at 10:30pm, and can be run on-demand. The test will discover the available versions of the cost-analyzer chart and select two previous versions: the current patch release and the previous minor release with its latest patch. For example, if the current release (not pre-release) is 2.2.5, then the test matrix will consist of 2.2.5 and 2.1.2 (where 2.1.2 is the latest patch within the 2.1 version train). If the current release is a dot zero major release, no upgrade tests will be performed as this is determined to likely involve breaking changes. The test will attempt to install these versions in separate jobs and proceed to upgrade to the current development state of the chart.

Although this CI test only performs the simplest Helm installation and upgrade tests, it is intended to get something started now while being extensible later.

No

No impact, CI only.

No

None. The CI test is not triggered by push or PR.

## How was this PR tested?

In a separate repository.

N/A


JesseGoodier (approved):
great work!

PR #3434, opened 2024-05-22
Delete values-agent.yaml

thomasvn:
## What does this PR change?
* Delete the `values-agent.yaml`
* If users need it for reference, they can look at older branches of the repo such as `v2.2`, `v2.1`, `v1.108`, etc.

* N/A

Kubecost 2.x uses a new archeticture and no longer supports the previous `values-agent.yaml` 
Users should now use:
```
federatedETL:
  ## If true, installs the minimal set of components required for a Federated ETL cluster.
  agentOnly: true
```

* Fixes https://github.com/kubecost/cost-analyzer-helm-chart/pull/3388. We should not need to bump the Thanos image anymore. Running the Thanos sidecar is considered non-standard now.

* N/A

## How was this PR tested?

* N/A

* Validated that our current docs do not reference this file.


chipzoller (approved):

jessegoodier:
I edited the release note section. There are going to be users that still use this file. I agree they need to migrate.
Anything else we need to add/communicate? 

@thomasvn @chipzoller

thomasvn:
I _think_ the only reason why a user would still need `values-agent.yaml` is to understand how to deploy Thanos sidecar? If so, I believe our documentation covers how ([docs ref](https://docs.kubecost.com/install-and-configure/install/multi-cluster/federated-etl/federated-etl-backups-alerting#option-2-metrics-backup)).

PR #3438, opened 2024-05-22
Adding support for the Azure Monitor Managed Service for Prometheus

acompton-wayve:
## What does this PR change?

This PR adds support for the [Azure Monitor Managed Service for Prometheus](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/prometheus-metrics-overview) as a target for remote writes and queries. My intention was for it to work in the same way as the other cloud providers' listed in the [Prometheus Configuration Guide](https://docs.kubecost.com/install-and-configure/advanced-configuration/custom-prom).

No.

With this PR, users will be able to use Azure's managed Prometheus offering as a remote_write target and query source, bypassing the bundled Prometheus.

N/A

I am not fully familiar with the risks, but none of the new functionality is enabled by default, so it shouldn't do anything unless someone turns it on.

## How was this PR tested?

I installed it on my own cluster and validated that 1) I was able to see the Kubecost metrics in Azure-managed Prometheus, and 2) Kubecost was able to issue queries even with the bundled Prometheus pod scaled down to zero.

No.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -491,6 +491,77 @@ spec:
           - name: ubbagent-config
             mountPath: /etc/ubbagent
         {{- end }}
+        {{- if .Values.global.ammsp.enabled }}
+        - name: {{ .Values.global.ammsp.aadAuthProxy.name }}
+          image: {{ .Values.global.ammsp.aadAuthProxy.image }}
+          {{- if .Values.global.ammsp.aadAuthProxy.imagePullPolicy }}
+          imagePullPolicy: {{ .Values.global.ammsp.aadAuthProxy.imagePullPolicy }}
+          {{- else }}
+          imagePullPolicy: Always
+          {{- end }}
+          env:
+          - name: AUDIENCE
+            value: {{ .Values.global.ammsp.aadAuthProxy.audience }}
+          - name: TARGET_HOST
+            value: {{ .Values.global.ammsp.queryEndpoint }}
+          - name: LISTENING_PORT
+            value: {{ .Values.global.ammsp.aadAuthProxy.port | quote }}
+          - name: IDENTITY_TYPE
+            value: {{ .Values.global.ammsp.aadAuthProxy.identityType }}
+          {{- if eq .Values.global.ammsp.aadAuthProxy.identityType "userAssigned"}}
+          - name: AAD_CLIENT_ID
+            value: {{ required "aadClientId is required for userAssigned identity types" .Values.global.ammsp.aadAuthProxy.aadClientId | toString | trim | quote }}
+          {{- else if eq .Values.global.ammsp.aadAuthProxy.identityType "aadApplication" }}
+          - name: AAD_CLIENT_ID
+            value: {{ required "aadClientId is required for aadApplication identity types" .Values.global.ammsp.aadAuthProxy.aadClientId | toString | trim | quote }}
+          - name: AAD_TENANT_ID
+            value: {{ required "aadTenantId is required for aadApplication identity type" .Values.global.ammsp.aadAuthProxy.aadTenantId | toString | trim | quote }}
+          - name: AAD_CLIENT_CERTIFICATE_PATH
+            value: {{ required "aadClientCertificatePath is required for aadApplication identity type" .Values.global.ammsp.aadAuthProxy.aadClientCertificatePath | toString | trim | quote }}
+          {{- end}}
+          - name: AAD_TOKEN_REFRESH_INTERVAL_IN_PERCENTAGE
+            value: "10"
+          - name: OTEL_SERVICE_NAME
+            value: {{ .Values.global.ammsp.aadAuthProxy.name | replace "-" "_" }}
+          {{- if .Values.systemProxy.enabled }}
+          - name: HTTP_PROXY
+            value: "{{ .Values.systemProxy.httpProxyUrl }}"
+          - name: http_proxy
+            value: "{{ .Values.systemProxy.httpProxyUrl }}"
+          - name: HTTPS_PROXY
+            value: "{{ .Values.systemProxy.httpsProxyUrl }}"
+          - name: https_proxy
+            value: "{{ .Values.systemProxy.httpsProxyUrl }}"
+          - name: NO_PROXY
+            value: "{{ .Values.systemProxy.noProxy }}"
+          - name: no_proxy
+            value: "{{ .Values.systemProxy.noProxy }}"
+          {{- end }}
+          ports:
+          - name: http
+            containerPort: {{ .Values.global.ammsp.aadAuthProxy.port | int }}
+            protocol: TCP
+          readinessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /ready
+              port: http
+              scheme: HTTP
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 5
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /health
+              port: http
+              scheme: HTTP
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 5
+        {{- end }}
         {{- if .Values.kubecostModel }}
         {{- if .Values.kubecostModel.fullImageName }}
         - image: {{ .Values.kubecostModel.fullImageName }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -492,6 +492,8 @@ spec:
             mountPath: /etc/ubbagent
         {{- end }}
         {{- if .Values.global.ammsp.enabled }}
+        # This section of the chart borrows liberally from
+        # https://github.com/Azure/aad-auth-proxy/blob/main/deploy/chart/aad-auth-proxy/templates/deployment.yaml
         - name: {{ .Values.global.ammsp.aadAuthProxy.name }}
           image: {{ .Values.global.ammsp.aadAuthProxy.image }}
           {{- if .Values.global.ammsp.aadAuthProxy.imagePullPolicy }}
@@ -508,7 +510,7 @@ spec:
             value: {{ .Values.global.ammsp.aadAuthProxy.port | quote }}
           - name: IDENTITY_TYPE
             value: {{ .Values.global.ammsp.aadAuthProxy.identityType }}
-          {{- if eq .Values.global.ammsp.aadAuthProxy.identityType "userAssigned"}}
+          {{- if eq .Values.global.ammsp.aadAuthProxy.identityType "userAssigned" }}
           - name: AAD_CLIENT_ID
             value: {{ required "aadClientId is required for userAssigned identity types" .Values.global.ammsp.aadAuthProxy.aadClientId | toString | trim | quote }}
           {{- else if eq .Values.global.ammsp.aadAuthProxy.identityType "aadApplication" }}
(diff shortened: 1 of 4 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -491,6 +491,77 @@ spec:
...
+          - name: LISTENING_PORT
+            value: {{ .Values.global.ammsp.aadAuthProxy.port | quote }}
+          - name: IDENTITY_TYPE
+            value: {{ .Values.global.ammsp.aadAuthProxy.identityType }}
+          {{- if eq .Values.global.ammsp.aadAuthProxy.identityType "userAssigned"}}
+          - name: AAD_CLIENT_ID
+            value: {{ required "aadClientId is required for userAssigned identity types" .Values.global.ammsp.aadAuthProxy.aadClientId | toString | trim | quote }}
+          {{- else if eq .Values.global.ammsp.aadAuthProxy.identityType "aadApplication" }}
+          - name: AAD_CLIENT_ID
+            value: {{ required "aadClientId is required for aadApplication identity types" .Values.global.ammsp.aadAuthProxy.aadClientId | toString | trim | quote }}
```
thomasvn, line 518: It looks like this is a duplicate definition of `AAD_CLIENT_ID`?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -491,6 +491,77 @@ spec:
...
+          - name: AAD_TENANT_ID
+            value: {{ required "aadTenantId is required for aadApplication identity type" .Values.global.ammsp.aadAuthProxy.aadTenantId | toString | trim | quote }}
+          - name: AAD_CLIENT_CERTIFICATE_PATH
+            value: {{ required "aadClientCertificatePath is required for aadApplication identity type" .Values.global.ammsp.aadAuthProxy.aadClientCertificatePath | toString | trim | quote }}
```
thomasvn, line 522: We should document `.Values.global.ammsp.aadAuthProxy.aadClientCertificatePath` in the `values.yaml` file?

thomasvn:
@acompton-wayve Thanks so much for your contribution here! I left a few minor comments, but this looks really good! Do you happen to have any notes on how you setup/configured your AMMSP? Something for us to reference when building out our docs on this.

@kwombach12 @chipzoller Thoughts on this?

kwombach12:
I'm supportive of rolling this out in our 2.4 release if we think we can quickly get this into a shippable state!

acompton-wayve:
Sure, sorry, I was away for a long holiday weekend. I'll address these notes ASAP today for you all. Thanks!

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
acompton-wayve: I stole this logic from [the aad-auth-proxy helm chart](https://github.com/Azure/aad-auth-proxy/blob/6c0f849997fa0d2ca3082b00710a2df130f31dc8/deploy/chart/aad-auth-proxy/templates/deployment.yaml#L48). I believe the idea is to provide `AAD_CLIENT_ID` in the `userAssigned` and `aadApplication` cases, but not otherwise. The "required" explainer bit is different in each case.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
acompton-wayve: I'll be frank, I do not understand Azure identity management well enough to know what ought to go here. I was only able to test in the `userAssigned` mode, which does not leverage this option.

acompton-wayve:
So here's what I put together on the Terraform side to set things up for this chart to function.

```
resource "azurerm_monitor_workspace" "kubecost" {
  name                = "kubecost"
  resource_group_name = <appropriate resource group name>
  location            = <appropriate location>
}
```

This sets up the [Azure Monitor Workspace](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/azure-monitor-workspace-overview), which the managed Prometheus requires to function.

```
# the referents of these data resources are automatically created by the workspace
data "azurerm_resource_group" "kubecost-managed-resource-group" {
  name = "MA_${azurerm_monitor_workspace.kubecost.name}_${azurerm_monitor_workspace.kubecost.location}_managed"
}

data "azurerm_monitor_data_collection_endpoint" "kubecost" {
  name                = azurerm_monitor_workspace.kubecost.name
  resource_group_name = data.azurerm_resource_group.kubecost-managed-resource-group.name
}

data "azurerm_monitor_data_collection_rule" "kubecost" {
  name                = azurerm_monitor_workspace.kubecost.name
  resource_group_name = data.azurerm_resource_group.kubecost-managed-resource-group.name
}
```
The workspace automatically generates a new resource group with a predictable name, and then creates a data collection endpoint and rule within it. These resources refer to those items so we can leverage them later.

```
resource "azurerm_role_assignment" "monitoring-metrics-publisher" {
  scope                = data.azurerm_resource_group.kubecost-managed-resource-group.id
  role_definition_name = "Monitoring Metrics Publisher"
  principal_id         = <appropriate principal ID>
}

resource "azurerm_role_assignment" "monitoring-reader" {
  scope                = <Kubernetes cluster resource group ID>
  role_definition_name = "Monitoring Data Reader"
  principal_id         = <appropriate principal ID>
}
```

Dealing with the metrics requires two permissions: the user-managed identity that aad-auth-proxy assumes needs to be able to write to managed Prometheus, and things running in the Kubernetes cluster (such as kubecost) need to be able to read from the managed Prometheus. I adapted [these instructions](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/prometheus-remote-write-virtual-machines?tabs=managed-identity#remote-write-using-user-assigned-managed-identity-authentication) for working this all out.

```
locals {
  principal_client_id      = <appropriate client ID>
  principal_tenant_id      = <appropriate tenant ID>
  query_endpoint           = azurerm_monitor_workspace.kubecost.query_endpoint
  data_collection_endpoint = "${replace(data.azurerm_monitor_data_collection_endpoint.kubecost.logs_ingestion_endpoint, ".ingest.monitor", ".metrics.ingest.monitor")}/dataCollectionRules/${data.azurerm_monitor_data_collection_rule.kubecost.immutable_id}/streams/Microsoft-PrometheusMetrics/api/v1/write?api-version=2023-04-24"
}
```
I pulled these out into locals so I could use them as helm chart parameters. Note that the data collection endpoint is specified in terraform for collecting _logs_, so I had to modify it with a regex to get the endpoint that collects _metrics_. This is to say, this whole mishegoss is just to arrive at the same URL that appears on the Azure Monitor Workspace web UI but which is not expressed anywhere in Terraform. :sob: 

```
resource "helm_release" "kubecost" {
  name             = "kubecost"
  chart            = "https://github.com/wayveai/cost-analyzer-helm-chart/releases/download/v2.2.2/cost-analyzer-2.2.2.tgz"
  namespace        = "infrastructure--helm--kubecost"
  create_namespace = true
  version          = "2.2.2"
  timeout          = 600
  values = [
    templatefile("${path.module}/helm_values/kubecost-values.tmpl",
      {
        "environment" : "prod",
        "remoteWriteService" : local.data_collection_endpoint,
        "queryEndpoint" : local.query_endpoint,
        "aadClientId" : local.principal_client_id,
        "aadTenantId" : local.principal_tenant_id,
      }
    )
  ]
}
```

This helm_release installs the chart from this specific branch, and feeds it with the parameters derived from creating the Azure Monitor Workspace and other resources.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
thomasvn: Oh! I see that now. Thank you for the clarification.

thomasvn:
This is looking great! Thanks for the detailed explanation here. Very helpful. Final request would be to include a few comments with links to helpful information?

For example in the cost-analyzer-deployment template it may help to add a comment that some of the configuration has been borrowed from [aad-auth-proxy](https://github.com/Azure/aad-auth-proxy/blob/main/deploy/chart/aad-auth-proxy/templates/deployment.yaml)? Refer there for more details.

acompton-wayve:
Sure, I added comments referencing the aad-auth-proxy repo and the relevant prometheus remote_write config for Azure. I'm happy to add anything else you all would like to see.

thomasvn (approved):

jessegoodier:
Thanks for this @acompton-wayve. just one small issue and you can just merge the suggestion if you agree.

acompton-wayve:
@jessegoodier , I do not see any suggestions to merge? What would you like me to do here to help?

teevans:
@jessegoodier - I don't think your review made it in - can you double check and see if it's in draft state?

jessegoodier:
@acompton-wayve 
Suggested change:
```
-      image: mcr.microsoft.com/azuremonitor/auth-proxy/prod/aad-auth-proxy/images/aad-auth-proxy:0.1.0-main-04-10-2024-7067ac84
+     image: $<IMAGE> # example: mcr.microsoft.com/azuremonitor/auth-proxy/prod/aad-auth-proxy/images/aad-auth-proxy:0.1.0-main-04-10-2024-7067ac84
```
given that there are other settings required, I think I would rather the image also be input by the user. This removes the image as a dependency of our vulnerability scanners.

jessegoodier:
odd, github has had a few minor issues lately. 
@teevans

acompton-wayve:
@jessegoodier , I was copying the other proxy sidecars:
https://github.com/kubecost/cost-analyzer-helm-chart/blob/01365b9c35ab000fedf5d70d0f4853f689269128/cost-analyzer/values.yaml#L32

For your suggested change, do you want it to literally say `$<IMAGE>` with the example comment as you described? Should I add a `required` bit so if the value isn't specified, it fails with a comprehensible error instead of a vague "ErrImagePull" further on down the line?

jessegoodier:
> @jessegoodier , I was copying the other proxy sidecars:
> 
> https://github.com/kubecost/cost-analyzer-helm-chart/blob/01365b9c35ab000fedf5d70d0f4853f689269128/cost-analyzer/values.yaml#L32
> 
> For your suggested change, do you want it to literally say `$<IMAGE>` with the example comment as you described? Should I add a `required` bit so if the value isn't specified, it fails with a comprehensible error instead of a vague "ErrImagePull" further on down the line?

you absolutely did this correctly. I'm trying to balance images that we have to manage because they are included by default in the helm chart with simplicity for users.

@AjayTripathy- what are your thoughts on this?

cost-analyzer/values.yaml
```diff
@@ -63,6 +63,27 @@ global:
...
+  ammsp:
+    enabled: false
+    prometheusServerEndpoint: http://localhost:8081/
+    remoteWriteService: $<AMMSP_METRICS_INGESTION_ENDPOINT>
+    queryEndpoint: $<AMMSP_QUERY_ENDPOINT>
+
+    aadAuthProxy:
+      enabled: false
+      # per https://github.com/Azure/aad-auth-proxy/releases/tag/0.1.0-main-04-10-2024-7067ac84
+      image: mcr.microsoft.com/azuremonitor/auth-proxy/prod/aad-auth-proxy/images/aad-auth-proxy:0.1.0-main-04-10-2024-7067ac84
```
```suggestion
      image: $<IMAGE>  # Example: mcr.microsoft.com/azuremonitor/auth-proxy/prod/aad-auth-proxy/images/aad-auth-proxy:0.1.0-main-04-10-2024-7067ac84
```

Yep I'm in agreement with @jessegoodier here. If the user runs into `ErrImagePullBackoff` I think it should be straightforward as to why that would happen. I don't think we need to add extra error checking here.

And yes I think it's preferred to comment out the image name to reduce the surface area for our image vulnerability scanner. Particularly because this is an optional additional feature.

acompton-wayve:
Do I need to do anything else to allow the auto-merge to proceed? Should I update with the latest changes from the `develop` branch?

thomasvn:
@acompton-wayve Correct! Once you merge `develop` into this branch and all CI tests pass, it should auto-merge.

PR #3447, opened 2024-05-31
add option for in memory ipc

JesseGoodier:
## What does this PR change?
Enables in memory IPC per file to improve ingestion performance. Roughly 20% increase in performance in initial tests

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1008,6 +1008,12 @@ Begin Kubecost 2.0 templates
     {{- toYaml .Values.kubecostAggregator.extraVolumeMounts | nindent 4 }}
     {{- end }}
   env:
+    {{- if .Values.kubecostAggregator.useInMemoryIPC }}
+    - name: IPC_DIR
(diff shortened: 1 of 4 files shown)


cost-analyzer/templates/_helpers.tpl
```diff
@@ -1002,6 +1008,12 @@ Begin Kubecost 2.0 templates
     {{- toYaml .Values.kubecostAggregator.extraVolumeMounts | nindent 4 }}
     {{- end }}
   env:
+    {{- if .Values.kubecostAggregator.useInMemoryIPC }}
+    - name: IPC_DIR
+      value: /ipc1
+    - name: PARQUET_DIR
```
ameijer, line 1014: jesse, I'd suggest that these go under separate flags, and while IPC_DIR is on by default, parquet is off. We have an idea of the bound of the IPC_CIR, but parquet dir could be unbounded, more detail here: https://kubecost.atlassian.net/browse/ENG-1232?focusedCommentId=17994

AjayTripathy:
Is this still relevant? Can we close the PR?

PR #3456, opened 2024-06-06
fix option for non-federated primary

JesseGoodier:
## What does this PR change?
allows the primary to serve UI without shipping the local cluster metrics to s3. This can be useful when 
1. a global Kubecost agent is deployed to all clusters already
2. for a parallel kubecost environment on the same primary

No

Add option for readOnly primary Kubecost configurations

Low risk as this is a new key pair and would only be used for specific goals

## How was this PR tested?
Tested by setting readOnlyPrimary=true and check the /federated folder to verify the cluster_id was not present

```yaml
federatedETL:
  readOnlyPrimary: true
```
Added comments. I see this as mostly an option for support to help our enterprise customers as of now.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -3148,8 +3148,10 @@ federatedETL:
   ## If true, push ETL data to the federated storage bucket
   federatedCluster: false
 
-  ## If true, the primary cluster will only read from the primary cluster and not ship it's own metrics
-  ## this allows a global kubecost-agent to be deployed to clusters, without a separate config for the primary cluster
+  ## If true, this cluster will be able to read from the federated-store but will
+  ## not write to it. This is useful in situations when you want to deploy a
+  ## primary cluster, but don't want the primary cluster's ETL data to be
+  ## pushed to the bucket
   readOnlyPrimary: false
 
   ## If true, changes the dir of S3 backup to the Federated combined store.
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -3148,6 +3148,10 @@ federatedETL:
   ## If true, push ETL data to the federated storage bucket
   federatedCluster: false
 
+  ## If true, the primary cluster will only read from the primary cluster and not ship it's own metrics
+  ## this allows a global kubecost-agent to be deployed to clusters, without a separate config for the primary cluster
+  readOnlyPrimary: false
+
   ## If true, changes the dir of S3 backup to the Federated combined store.
   ## Commonly used when transitioning from Thanos to Federated ETL architecture.
   redirectS3Backup: false
--- a/cost-analyzer/Chart.yaml
+++ b/cost-analyzer/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: development
 description: Kubecost Helm chart - monitor your cloud costs!
 name: cost-analyzer
-version: v2024.06.06
+version: 2.2.2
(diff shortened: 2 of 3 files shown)


thomasvn:
Thanks for this fix! So it seems that we are currently unable to create a "readonly" primary? Based on the existing helm chart's templating it seems that as long as `federated-store` exists, the template will not respect `federatedCluster=false`. (example below).

```yaml
federatedETL:
  federatedCluster: false
kubecostModel:
```

There is one main risk I see. What if there are users who only have the following config across all their deployments? This below config is sufficient to push ETL to the bucket. Additionally because `federatedETL.federatedCluster=false` by default ([code ref](https://github.com/kubecost/cost-analyzer-helm-chart/blob/b128331ceaabb44e5bc4b2b1d3fd3dbe4abca52e/cost-analyzer/values.yaml#L3148-L3149)), this would cause users upgrading to the new version to suddenly stop pushing ETL to the bucket based on the current proposed logic in this PR.

```yaml
kubecostModel:
```

jessegoodier:
> Thanks for this fix! So it seems that we are currently unable to create a "readonly" primary? Based on the existing helm chart's templating it seems that as long as `federated-store` exists, the template will not respect `federatedCluster=false`. (example below).
> 
> ```yaml
> federatedETL:
>   federatedCluster: false
> kubecostModel:
>   federatedStorageConfigSecret: federated-store
> ```
> 
> There is one main risk I see. What if there are users who only have the following config across all their deployments? This below config is sufficient to push ETL to the bucket. Additionally because `federatedETL.federatedCluster=false` by default ([code ref](https://github.com/kubecost/cost-analyzer-helm-chart/blob/b128331ceaabb44e5bc4b2b1d3fd3dbe4abca52e/cost-analyzer/values.yaml#L3148-L3149)), this would cause users upgrading to the new version to suddenly stop pushing ETL to the bucket based on the current proposed logic in this PR.
> 
> ```yaml
> kubecostModel:
>   federatedStorageConfigSecret: federated-store
> ```

This is why you are the best.

I'm thinking we just use a new key pair to avoid any backwards compatibility issues. 
like
```yaml
federatedETL:
  ReadOnlyPrimary: true
```

new commit coming.

cost-analyzer/values.yaml
```diff
@@ -3148,6 +3148,10 @@ federatedETL:
   ## If true, push ETL data to the federated storage bucket
   federatedCluster: false
 
+  ## If true, the primary cluster will only read from the primary cluster and not ship it's own metrics
+  ## this allows a global kubecost-agent to be deployed to clusters, without a separate config for the primary cluster
```
```suggestion
  ## If true, this cluster will be able to read from the federated-store but will
  ## not write to it. This is useful in situations when you want to deploy a
  ## primary cluster, but don't want the primary cluster's ETL data to be
  ## pushed to the bucket
```

Thoughts on the comment adjustment?

thomasvn:
Thanks for the update! LGTM.

thomasvn (approved):

thomasvn:
@jessegoodier Although I think this change is good for _most_ use cases, I think there are a few scenarios where a user may want to set `.Values.aggregator.deployMethod=singlepod` on a secondary cluster. Notably, if they want to run an aggregator + UI on that secondary cluster which displays the non-federated data.

The following logic shows that if Aggregator's deploymethod is StatefulSet on the secondary, it will add env vars for the `FEDERATED_STORE_CONFIG` and therefore read in multi-cluster data.

https://github.com/kubecost/cost-analyzer-helm-chart/blob/fd04f3111f8a541c9f72af8d3d53709d9145444d/cost-analyzer/templates/_helpers.tpl#L1096-L1115

PR #3480, opened 2024-06-18
remove env: from aggregator

JesseGoodier:
## What does this PR change?
Removed kubecostAggregator.env from helm values to avoid unknown issues that would be caused by previous customizations.

kubecostAggregator.env has been removed from the helm values to avoid unknown issues that would be caused by previous customizations. Aggregator has been improved and should not need any of these previous settings. Any environment variables that are needed can be set with kubecostAggregator.extraEnv

Unknown user configurations

## How was this PR tested?
Test in QA environments and verified env: values were not used.

Will have a strong note in release notes, but there should be little to no impact.


cliffcolvin (approved):
LGTM

PR #3490, opened 2024-06-20
fix carbonEstimates typo

JesseGoodier:
## What does this PR change?
Fix inconsistent carbonEstimates values

No

NA, never was a release. Shout out to @srpomeroy for the find.

None

## How was this PR tested?
helm install with 
```
kubecostProductConfigs:
  carbonEstimates: true
```

NA


nealormsbee (approved):
LGTM. Nice catch.

PR #3495, opened 2024-06-20
Add a skip validation in cloudIntegrationSecretCheck function to allow ArgoCD to sync

magnonta:


chipzoller:
We already have this mechanism [here](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L271-L273).

PR #3510, opened 2024-06-24
[CI] Bump/add cluster versions

### LINKED ISSUE #3493: [CI] Add Kubernetes 1.30 to test matrix
KinD currently has a node image available but Replicated doesn't yet offer EKS testing on K8s 1.30. For the EKS testing, we will probably need to add another value based on [changes in that version](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions-standard.html#kubernetes-1.30).

chipzoller:

## What does this PR change?

* Adds EKS v1.30 to the test matrix and forces the `gp2` StorageClass for all EKS installations.
* Updates all KinD minor versions to latest patch image
* Adds 1.30 to KinD test matrix
* Adds OKD 4.15 to test matrix

No

None, CI only.

Closes #3493

None, CI only.

## How was this PR tested?

Will be tested manually after PR is merged. Testing of this change will not be reflected in the Actions run as part of this PR (known limitation).

N/A

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -61,17 +61,19 @@ jobs:
           - name: v1.23
             version: v1.23.17
           - name: v1.24
-            version: v1.24.15
+            version: v1.24.17
           - name: v1.25
-            version: v1.25.11
+            version: v1.25.16
           - name: v1.26
-            version: v1.26.6
+            version: v1.26.15
           - name: v1.27
-            version: v1.27.3
+            version: v1.27.13
           - name: v1.28
-            version: v1.28.0
+            version: v1.28.9
           - name: v1.29
-            version: v1.29.0
+            version: v1.29.4
+          - name: v1.30
+            version: v1.30.0
     needs: test-chart
     name: ${{ matrix.k8s-version.name }} test
     steps:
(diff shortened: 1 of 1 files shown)


chipzoller:
An open question here:  Should we add into the EKS values file the equivalent value to set `gp2` as the StorageClass? Pros: it will "just work" for users of EKS 1.30+. Cons: it's a default change which may not work for users who have removed this StorageClass.

thomasvn:
I like the idea of defaulting to a storageClass that "just works" if one is not already defined. As you've mentioned I think it can get quite complicated for existing users if they are already on a different storageClass. Maybe for now it's best to doc as a troubleshooting step?

thomasvn (approved):

chipzoller:
I did at least document it here: https://docs.kubecost.com/install-and-configure/install/provider-installations/aws-eks-cost-monitoring#prerequisites-1

thomasvn:
Perfect! I'd be open to attempting a gp2 default. My biggest concern would b that swapping storageClasses may be a destructive operation on some people's PVs? Not 100% sure.

chipzoller:
Adding a default value sounds good to me. Does that work for you, @jessegoodier ?

PR #3538, opened 2024-07-11
Add extraScrapeConfig to scrape DCGM Exporter

thomasvn:
## What does this PR change?

Default enables the Kubecost Helm Chart to discover and scrape DCGM Exporter metrics. Specifically, scrapes endpoints which have the key "dcgm-exporter" (case insensitive) for the following labels:

- "app"
- "app.kubernetes.io/component"
- "app.kubernetes.io/name"

N/A

For users who do not run DCGM Exporter, this change has no impact. For users who do, they must ensure the following two conditions are true:

1. Their DCGM Exporter service endpoint is correctly labeled as described above
2. They must not currently be overriding the default `.Values.prometheus.extraScrapeConfigs`. Otherwise they will not automatically inherit this change.

Closes [ENG-2428](https://kubecost.atlassian.net/browse/ENG-2428)

Minimal risk for clusters not currently running DCGM Exporter. I don't anticipate significant additional overhead on Prometheus either. For clusters that do run DCGM Exporter, they must abide by the conditions specified above.

## How was this PR tested?

Deployed [DCGM Exporter](https://github.com/NVIDIA/dcgm-exporter) and this new scrapeconfig. Port-forwarding into Prometheus's UI showed that Prometheus was able to successfully service discover and scrape DCGM exporter:

Screenshot 2024-07-15 at 12 13 32 PM

Screenshot 2024-07-15 at 12 15 23 PM

https://github.com/kubecost/docs/pull/1092

[ENG-2428]: https://kubecost.atlassian.net/browse/ENG-2428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chipzoller (commented):
I'm thinking this is best left commented out and remain an opt-in config option for the reasons of diagnostics/alerting that we have discussed previously. Do you think otherwise?

thomasvn:
@chipzoller I don't think there would be any harm in leaving this default enabled! For similar reasons that there is a scrapeconfig available for kubecost-network-costs default enabled, however not all users run the network costs daemonset.

chipzoller:
Would having this enabled by default impact our ability to diagnose and differentiate between the scenarios where:
1. No nodes in the cluster have any NVIDIA GPUs
2. At least one node in the cluster has at least one NVIDIA GPU which is currently unused
3. Same as 2 but is currently used

thomasvn:
> No nodes in the cluster have any NVIDIA GPUs

If no nodes are running NVIDIA GPUs, then presumably they're unable to run any DCGM exporter as well. In which case we would not be scraping any DCGM metrics for this cluster. No GPU cost and no GPU metrics, which would be correct.

> At least one node in the cluster has at least one NVIDIA GPU which is currently unused

Yes! This is differentiable. If they have unused GPUs and are also exporting DCGM exporter metrics, Prometheus would be able to scrape the DCGM metrics for these nodes and identify that usage is zero. GPU cost exists, but the scraped GPU metrics would indicate no usage.

If, however DCGM Exporter is not configured, we would not have metrics/visibility into GPU usage. So GPU cost would exist but we would not be unable to calculate any level of efficiency.

> Same as 2 but is currently used

Again, I think this is differentiable as long as they have DCGM Exporter running. Prometheus would be able to scrape DCGM metrics and identify some level of GPU usage. GPU cost exists and GPU usage exists.

chipzoller:
Ok, so then we'd want to alert/diagnose when the following conditions are all true:
1. DCGM exporter is found.
2. At least one node has at least one NVIDIA GPU.
3. At least one container on said node is currently using a GPU.
4. No indicative metric(s) is/are observed from DCGM exporter.

Does this change support being able to solve for this?

thomasvn:
@chipzoller Yes! This change would support the above. We would have to build the additional alerting/diagnostic logic into KCM though, likely into the multicluster diagnostics package.

chipzoller:
Yes, understood. Seems good to me then! I will leave one recommendation.

cost-analyzer/values.yaml
```diff
@@ -909,6 +909,14 @@ prometheus:
         {{- else }}
         port: 9004
         {{- end }}
+    - job_name: kubecost-dcgm-exporter
```
chipzoller, line 919:
```suggestion
    # Enables scraping of NVIDIA GPU metrics via dcgm-exporter
    - job_name: kubecost-dcgm-exporter
```

chipzoller (commented):
Putting in a comment here so it's a little more clear to users what this does.

thomasvn:
Ready for final review! Would love your thoughts @chipzoller @jessegoodier @AjayTripathy. Let me know if there are any risks.

chipzoller (approved):
LGTM

thomasvn:
@jessegoodier Let me know what you think!

JesseGoodier (approved):

PR #3561, opened 2024-07-28
grafana ingress comments

JesseGoodier:
## What does this PR change?
Add comments to Grafana ingress to avoid confusion

No

Add comments to Grafana ingress to avoid confusion

Support spent an hour trying to get this working with a user and found that the default ingress was all that was required. 

NA

## How was this PR tested?
Helm upgrade of QA environment

NA


thomasvn (approved):
Thanks for adding the comment! In the future we may consider removing the grafana-ingress template altogether.

PR #3569, opened 2024-08-02
Add useClientSecretPost to fix okta oidc redirect loop

kaelanspatel:
## What does this PR change?
Adds `useClientSecretPost` config to `.Values.oidc` which explicitly disables all auth methods except `client_secret_post` when doing client authentication while verifying tokens through the introspection endpoint.

Yes, see linked KCM PR.

Adds ability to disable all client auth methods except `client_secret_post`, which fixes issues where the Authorization header was invalid.

Testing done in linked PR.

## How was this PR tested?
Testing done in linked PR.

No.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
@@ -24,7 +24,7 @@ data:
         "discoveryURL" : "{{ .Values.oidc.discoveryURL }}",
         "hostedDomain" : "{{ .Values.oidc.hostedDomain }}",
         "skipOnlineTokenValidation" : "{{ .Values.oidc.skipOnlineTokenValidation | default "false" }}",
-        "useClientSecretPost": {{ .Values.oidc.useClientSecretPost | default "false" }},
+        "useClientSecretPost": {{ .Values.oidc.useClientSecretPost }},
         "rbac" : {
           "enabled" : {{ .Values.oidc.rbac.enabled }},
           "groups" : [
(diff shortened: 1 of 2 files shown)


nickcurie (approved):

cost-analyzer/templates/cost-analyzer-oidc-config-map-template.yaml
```diff
@@ -24,6 +24,7 @@ data:
         "discoveryURL" : "{{ .Values.oidc.discoveryURL }}",
         "hostedDomain" : "{{ .Values.oidc.hostedDomain }}",
         "skipOnlineTokenValidation" : "{{ .Values.oidc.skipOnlineTokenValidation | default "false" }}",
+        "useClientSecretPost": {{ .Values.oidc.useClientSecretPost | default "false" }},
```
jessegoodier:
```suggestion
        "useClientSecretPost": {{ .Values.oidc.useClientSecretPost }},
```

cost-analyzer/values.yaml
```diff
@@ -376,6 +376,7 @@ oidc:
   # loginRedirectURL: "http://my.kubecost.url/model/oidc/authorize" # Kubecost url configured in provider for redirect after authentication
   # discoveryURL: "https://my.auth.server/.well-known/openid-configuration" # url for OIDC endpoint discovery
   skipOnlineTokenValidation: false  # if true, will skip accessing OIDC introspection endpoint for online token verification, and instead try to locally validate JWT claims
+  # useClientSecretPost: true # if true, client secret will specifically only use client_secret_post method, otherwise it will attempt to send the secret in both the header and the body.
```
```suggestion
  useClientSecretPost: false  # if true, client secret will specifically only use client_secret_post method, otherwise it will attempt to send the secret in both the header and the body.
```

jessegoodier:
Is there a reason to comment out the value? 

I realize we have a mix of standards here, but I think it is preferred to set the default value in the values file.
Allows tools to easily parse our settings.

kaelanspatel:
@jessegoodier Sorry, didn't see this last week! I'm a little leery to these accumulating configs boating the experience, but I don't have a strong feeling either way. Considering all sections in `saml` and nessesary sections of `oidc` are commented out, I opted to follow that.

JesseGoodier (approved):

thomasvn:
@kaelanspatel Could you help update [the docs](https://docs.kubecost.com/install-and-configure/advanced-configuration/user-management-oidc) with a brief paragraph about when this config should be used? Perhaps in the troubleshooting section?

thomasvn:
@kaelanspatel @cliffcolvin I see this is in v2.3.5. Could we make some time to get this documented?

PR #3576, opened 2024-08-09
Resolved issue where aggregator pod fails due to missing smtp-secret volume (fix for v2.3.5 rc.0)

patrick-tavares:
Resolved issue where, when using an SMTP secret, the aggregator statefulset pod would fail to start due to a missing smtp-secret volume.


chipzoller:
Not sure what you're trying to do here, but if it's to land the fix in #3575 in the v2.3 branch then there's no need to do this.

PR #3590, opened 2024-08-19
option to use nginx cache

JesseGoodier:


cost-analyzer/values.yaml
```diff
@@ -487,6 +487,19 @@ kubecostFrontend:
...
+  cache:
+    enabled: false
+    cacheSize: 2000m
+    cacheTTL: 1h
+    proxyCachePath: "/tmp/cache levels=1:2:2 keys_zone=kubecost_cache:2g max_size=10g inactive=6h use_temp_path=off"
+    proxyCacheMethods: "GET POST HEAD"
+
+# aggregatorFailoverSvc is experimental and only supported with Kubecost Enterprise
+# location of the aggregator service to use as a failover
+  # aggregatorFailoverSvc: kubecost2-aggregator.kubecost2.svc.cluster.local:9004
```
jessegoodier, line 515: Should this be uncommented and just add a section for "aggregatorFailover.enabled"?

chipzoller: I'd probably have both an `enabled` field as well as a service address field.

PR #3591, opened 2024-08-19
Add ingestion config for standard discount

kaelanspatel:
## What does this PR change?
Adds configmap handling and env to support standard discount in aggregator (see linked PR).

https://github.com/kubecost/kubecost-cost-model/pull/2668

See linked PR

https://kubecost.atlassian.net/browse/ENG-2412

See linked PR

## How was this PR tested?
See linked PR

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1107,7 +1107,7 @@ Begin Kubecost 2.0 templates
       value: "true"
     - name: FEDERATED_CLUSTER # this ensures the ingester runs assuming federated primary paths in the bucket
       value: "true"
-    {{- if .Values.kubecostProductConfigs.standardDiscount }}
+    {{- if (.Values.kubecostProductConfigs).standardDiscount }}
     {{- if .Values.ingestionConfigmapName }}
     - name: INGESTION_CONFIGMAP_NAME
       value: {{ .Values.ingestionConfigmapName }}
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1107,6 +1107,12 @@ Begin Kubecost 2.0 templates
       value: "true"
     - name: FEDERATED_CLUSTER # this ensures the ingester runs assuming federated primary paths in the bucket
       value: "true"
+    {{- if .Values.kubecostProductConfigs.standardDiscount }}
+    {{- if .Values.ingestionConfigmapName }}
+    - name: INGESTION_CONFIGMAP_NAME
+      value: {{ .Values.ingestionConfigmapName }}
+    {{- end }}
+    {{- end }}
       {{- end }}
     {{- end }}
     - name: LOG_LEVEL
--- a/cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
@@ -1,4 +1,3 @@
-{{- if eq (include "aggregator.deployMethod" .) "statefulset" }}
 {{- if .Values.kubecostProductConfigs.standardDiscount }}
 apiVersion: v1
 kind: ConfigMap
@@ -10,5 +9,4 @@ metadata:
 data:
   standardDiscount: "{{ .Values.kubecostProductConfigs.standardDiscount }}"
   helmConfig: "true"
-{{- end -}}
 {{- end -}}
--- a/cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
+++ b/cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
@@ -1,0 +1,14 @@
+{{- if eq (include "aggregator.deployMethod" .) "statefulset" }}
+{{- if .Values.kubecostProductConfigs.standardDiscount }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ default "ingestion-configs" .Values.ingestionConfigmapName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+data:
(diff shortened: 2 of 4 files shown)


cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if eq (include "aggregator.deployMethod" .) "statefulset" }}
```
thomasvn, line 1: Do you think it's a requirement we check that Aggregator is being deployed as a StatefulSet? If so, do we need to perform this same check in `_helpers.tpl`?

cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+{{- if .Values.kubecostProductConfigs.standardDiscount }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ default "ingestion-configs" .Values.ingestionConfigmapName }}
```
thomasvn, line 6: Thoughts on changing `.Values.ingestionConfigmapName` to something like `.Values.kubecostProductConfigs.standardDiscountConfigmapName`?

cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
kaelanspatel: It's not a requirement, just that `statefulset` is the only valid deployment method of a fed primary aggregator, so I thought it would be worth checking. There's not really any risk.

thomasvn: I'd opt to remove it! I think >90% of enterprise configurations will use the StatefulSet, but a few don't.

cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
kaelanspatel: The idea was (ref discussion with Alex in: https://github.com/kubecost/kubecost-cost-model/pull/2668) to have a specific dedicated configmap + config struct for all ingestion-level config items (Obviously this discount is the only one present currently, but for posterity's sake). So I think `ingestionConfigmap` is correct for the intent, at least.

thomasvn: Ok, good with me!

cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
kaelanspatel: Done.

cost-analyzer/templates/cost-analyzer-ingestion-configmap.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.kubecostProductConfigs.standardDiscount }}
```
thomasvn:
```suggestion
{{- if (.Values.kubecostProductConfigs).standardDiscount }}
```

I believe you will need this to fix the Chart Build issue. "Only try to dereference standardDiscount if kubecostProductConfigs is defined".

kaelanspatel: Yeah oops, just saw that. My bad.

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1107,6 +1107,12 @@ Begin Kubecost 2.0 templates
       value: "true"
     - name: FEDERATED_CLUSTER # this ensures the ingester runs assuming federated primary paths in the bucket
       value: "true"
+    {{- if .Values.kubecostProductConfigs.standardDiscount }}
+    {{- if .Values.ingestionConfigmapName }}
+    - name: INGESTION_CONFIGMAP_NAME
+      value: {{ .Values.ingestionConfigmapName }}
+    {{- end }}
+    {{- end }}
```
thomasvn: Currently this is defined within the the following conditionals:

```
{{- if eq (include "aggregator.deployMethod" .) "statefulset" }}
{{- if and (not .Values.kubecostModel.federatedStorageConfigSecret) (not .Values.kubecostModel.federatedStorageConfig) }}
```

Should we move INGESTION_CONFIGMAP_NAME out of these conditionals? Since it shouldn't depend on Aggregator being deployed as StatefulSet?

Also, needs `(.Values.kubecostProductConfigs).standardDiscount`

kaelanspatel: Tbh the thought process was that this configuration essentially doesn't have a purpose outside of `(.Values.federatedETL).primaryCluster`, which forces `statefulset` IIRC; gating it behind that I thought was fine. There is no purpose in adding this to anything other than a fed etl architecture config.

thomasvn: Makes sense, but what about future ingestionConfigs?

kaelanspatel: This is not strictly wholistic, but the imagined use-case for an ingestionConfig is explicitly "I want to bulk-modify everything I ingest in the aggregator regardless of source". If not in a federated environment, I'd argue there's no point that this configmap exist.

thomasvn: I can imagine a (low-likelihood) case where we want to modify what was ingested but even in single cluster setup! Anyways, I think this is a minor point for now. And with respect to this PR it only defines `INGESTION_CONFIGMAP_NAME` which should only be set if the user provides an override. I'm good approving as is!

thomasvn (approved):

cost-analyzer/templates/_helpers.tpl
thomasvn: If you want to add the following back to the top of your configmap, that may be good for the sake of consistency then.

```
{{- if eq (include "aggregator.deployMethod" .) "statefulset" }}
```

PR #3598, opened 2024-08-20
reduced memory

JesseGoodier:
## What does this PR change?
change db copy partitions to 25 from 1, which was a typo
remove obsolete env var DB_COPY_FULL

NA

Reduced memory during read db copy operation
remove obsolete env var DB_COPY_FULL

NA, many configurations running this already.

## How was this PR tested?
confirmed with Bolt on obsolete var
25 copy partitions was and should be our default. tested on nightly. 

NA


cliffcolvin (approved):

ameijer (commented):
LGTM

PR #3621, opened 2024-08-29
chore: use ghcr.io instead of defaulting to DockerHub due to rate limit

rajpratik71:
## What does this PR change?

Use ghcr.io instead of defaulting to DockerHub due to rate limit

No

Use ghcr.io instead of defaulting to DockerHub due to rate limit
Internal note: this will be disabled by default in 2.5.0, disregard this PR in the release notes.


JesseGoodier (approved):
Thanks for the good idea. 
We are going to disable this by default, but will suggest ghcr.io if anyone wants to enable it. 

This should be done in 2.5.0

PR #3622, opened 2024-08-29
Make basic-health image configurable

clive-jevons:
in order to be able to pull all images from a private registry, we need to have the basic-health image configurable

## What does this PR change?

Makes `basic-health` image configurable.

No

Gives them the ability to configure the image used for basic-health, e.g. to pull it from a private registry


chipzoller (changes requested):
Thanks for the PR. Although I hate introducing new top-level values, I'm not sure of a better place. I'd rather see them commented out by default though.

clive-jevons:
@chipzoller  <- thanks for the feedback 🤩 Have addressed it in my last commit - WDYT?

chipzoller (approved):

thomasvn:
@chipzoller I believe this `basic-health` pod is only used for this repo's CI/CD tests. Would a better path forward here, be to give users the option to not template out the `basic-health` pod at all?

chipzoller:
The test is usable by any users who perform a `helm test` command and so we wouldn't quite know how many people are doing so. The idea was that both we and users could perform some basic smoke tests on Kubecost. The pod is not templated until a `helm test` command is issued and so therefore not really applicable unless folks want to run the smoke test.

thomasvn:
@chipzoller Ok I see what you're saying. In that case this lgtm!

thomasvn (approved):

AlfonsoCampodonico:
> The test is usable by any users who perform a `helm test` command and so we wouldn't quite know how many people are doing so. The idea was that both we and users could perform some basic smoke tests on Kubecost. The pod is not templated until a `helm test` command is issued and so therefore not really applicable unless folks want to run the smoke test.

Quick question regarding this, it does appear on the generate template when i do `helm template kubecost kubecost/cost-analyzer`

Is there a way to not create this pod?

chipzoller:
`--skip-tests`

thomasvn:
@chipzoller Do you think this should go in v2.4.x or v2.5.0?

chipzoller:
Would suggest 2.5 unless there's somehow a burning need for 2.4.

PR #3651, opened 2024-09-10
update prometheus.server.retion to 97h

doge95:
## What does this PR change?
According to v2.3.5 chart, this value is set to 97h in the `values.yaml`
https://github.com/kubecost/cost-analyzer-helm-chart/blob/v2.3.5/cost-analyzer/values.yaml#L1303C16-L1303C19


JesseGoodier (approved):

jessegoodier:
@doge95 nice find! Thanks for the PR

PR #3672, opened 2024-09-16
update test in ci-cd, remove k8s image from helm repo

cliffcolvin:
Signed-off-by: Cliff Colvin <ccolvin@kubecost.com>

## What does this PR change?
Remove k8s image from helm, and insert it only during ci/cd. This will assist in passing helm scans.

NA

NA, cicd workflow only

NA

## How was this PR tested?
Locally

NA


JesseGoodier (approved):

cliffcolvin:
branch build is still pulling from develop. We will merge, test in new pr and if issues fix/revert.

chipzoller (commented):
I don't think this is the right approach. Removing the test will prevent users from being able to perform a basic e2e smoke test on Kubecost. It also impacts, if I recall, marketplace catalogs like Google because they require the test be in place. If the idea is to simply quiet the scanners on our side, I think we can just omit the test command.

cliffcolvin:
> I don't think this is the right approach. Removing the test will prevent users from being able to perform a basic e2e smoke test on Kubecost. It also impacts, if I recall, marketplace catalogs like Google because they require the test be in place. If the idea is to simply quiet the scanners on our side, I think we can just omit the test command.

@chipzoller you're saying we need the test in place in the packaged chart?

chipzoller:
I'm not sure if this is specifically a "need", but I do know that a test was required by some marketplace. Can't remember if that was Google or Amazon.

What is the overall motivation for removing this test?

cliffcolvin:
Its definitely not any major issue. Any tool that looks for images in the chart sees k8s and it has vulns. Easy enough to explain away, its not actually used in our application when you install it.

cliffcolvin:
Given the feedback here I'm closing this.

PR #3673, opened 2024-09-16
add warning for federated storage

JesseGoodier:
## What does this PR change?
add warning to helm note when federated storage is not configured completely.

```
***Configuration issue detected:***
When a federated store is provided, Kubecost should either be running as agentOnly or as a statefulset.
.Values.federatedETL.agentOnly=true
Or
.Values.kubecostAggregator.deployMethod=statefulset
***
```

No

add warning to helm note when federated storage is not configured completely.

## How was this PR tested?

helm upgrades with and without federatedStorage, with and without agentOnly and statefulsets.


JesseGoodier (commented):
Will add to 2.4.1

Sean-Holcomb (approved):
LGTM

PR #3677, opened 2024-09-19
Adding OpenShift SCC clusterController, removing hostPort

kc-adawson:
Adding OpenShift SCC for clusterController, removing hostPort on clusterController

## How was this PR tested?

Ran on Openshift cluster successfully


cost-analyzer/values.yaml
```diff
@@ -247,6 +247,7 @@ global:
       scc:
         nodeExporter: false  # Creates an SCC for Prometheus Node Exporter. This requires Node Exporter be enabled.
         networkCosts: false  # Creates an SCC for Kubecost network-costs. This requires network-costs be enabled.
+        clusterController: false # Creates an SCC for Kubecost clusterContoller. This requires clusterController be enabled.
```
```suggestion
        clusterController: false  # Creates an SCC for Kubecost clusterContoller. This requires clusterController be enabled.
```

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -256,7 +256,6 @@ spec:
         ports:
         - name: http-server
           containerPort: 9731
-          hostPort: 9731
```
jessegoodier, line 259: @alexkubecost I can dig here, but do you know off hand why we would need hostPort?

I don't understand why anything would be connecting to the clusterController that doesn't know the service name.

cost-analyzer/templates/kubecost-cluster-controller-ocp-scc.yaml
```diff
@@ -0,0 +1,33 @@
...
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+seccompProfiles:
+- runtime/default
+volumes:
+  - hostPath
+  - projected
+  - configMap
+hostPorts:
```
jessegoodier, line 28: You can kill these hostPort lines if Alex is good with removing hostPort.

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
thomasvn: I can't currently think of any reason why hostPort would be needed. @ameijer ?

thomasvn:
Once this is approved, let's make sure this is merged into both `v2.4` and `develop` branches.

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
ameijer: I'll bet you that this yaml was written from an example where they exposed an http port on a node and the ingress controller routed to that. I agree there is rarely if ever a use case for a host port on non daemon set controllers. I think getting rid of it is a great idea

chipzoller:
Changed branch to `develop` so need to go back through this.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -1425,7 +1425,7 @@ data:
                 "carbonEstimatesEnabled": "{{ template "carbonEstimatesEnabled" . }}",
                 "clusterControllerEnabled": "{{ template "clusterControllerEnabled" . }}",
                 "forecastingEnabled": "{{ template "forecastingEnabled" . }}",
-                "chartVersion": "DEVELOP_BRANCH",
+                "chartVersion": "2.4.0",
```
chipzoller, line 1428: I don't think we want to change this now it's targeting `develop`.

cost-analyzer/templates/kubecost-cluster-controller-ocp-scc.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if and (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") (.Values.global.platforms.openshift.scc.clusterController) (.Values.clusterController.enabled) }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{ template "kubecost.clusterControllerName" . }}
+priority: 10
+allowPrivilegedContainer: true
+allowHostDirVolumePlugin: true
+allowHostNetwork: true
+allowHostPorts: true
```
chipzoller, line 10: So this would now be `false`, right, if there are no hostPorts?

cost-analyzer/values-openshift.yaml
```diff
@@ -12,14 +12,11 @@ global:
       scc:
         nodeExporter: false  # Creates an SCC for Prometheus Node Exporter. This requires Node Exporter be enabled.
         networkCosts: false  # Creates an SCC for Kubecost network-costs. This requires network-costs be enabled.
+        clusterController: false # Creates an SCC for Kubecost clusterContoller. This requires clusterController be enabled.
```
chipzoller, line 15:
```suggestion
        clusterController: false # Creates an SCC for Kubecost clusterController. This requires clusterController be enabled.
```

cost-analyzer/values.yaml
```diff
@@ -247,6 +247,7 @@ global:
       scc:
         nodeExporter: false  # Creates an SCC for Prometheus Node Exporter. This requires Node Exporter be enabled.
         networkCosts: false  # Creates an SCC for Kubecost network-costs. This requires network-costs be enabled.
+        clusterController: false  # Creates an SCC for Kubecost clusterContoller. This requires clusterController be enabled.
```
chipzoller, line 250:
```suggestion
        clusterController: false  # Creates an SCC for Kubecost clusterController. This requires clusterController be enabled.
```

kc-adawson:
Hi all, after further testing, it seems all that is needed it the removal of the hostport from kubecost-cluster-controller-template.yaml. Once that is done, no specific SCCs will be needed. Maybe the Platforms.Openshift.True can remove hostport if we want to leave it on for other deployments? Either way, the SCC file is not needed.

chipzoller:
So what's the plan here?

kc-adawson:
This can be closed with no action needed for this specific PR. Maybe a
clusterController.hostPort=false or removing it all together would do the
trick.

Andrew Dawson
Solutions Engineer / Book Some Time
<https://calendly.com/stackwatch-adawson/15min>

PR #3700, opened 2024-10-06
Add kubeRBACProxy property to support querying in cluster prometheus in openshift

### LINKED ISSUE #3690: [OpenShift]: Using the existing Prometheus in an OpenShift cluster?
Hi all,

is it possible to use the Prometheus instance that is already running in an OpenShift cluster?

I found some issues (#752 , #2128) related to OpenShift and not using the kubecost-bundled prometheus. So I guess it could be possible, but those issues contain no hint if they succeeded without workarounds.

I would prefer to not run Prometheus twice and do all the scraping twice.

If this is possible, it would be nice if this could be included in the documentation.
https://docs.kubecost.com/install-and-configure/install/provider-installations/openshift-kubecost-install

Kind Regards,
Johannes

mittal-ishaan:
## What does this PR change?
* Add `kubeRBACProxy` property to set `KUBE_RBAC_PROXY_ENABLED` as true in the env var of the cost-model.
* `createPrometheusClusterRoleBinding` property to create cluster role binding to grant required permissions to the serviceaccount to query prometheus with [kube-rbac-proxy](https://catalog.redhat.com/software/containers/rhacm2/kube-rbac-proxy-rhel8/611d4320eb34b73652492bb8) enabled. 
* Add the role binding to grant permissions this external prometheus to scrape metrics from kubecost. 
* Add a check that both `kubeRBACProxy` and `BearerToken` property can not be set together,

https://github.com/opencost/opencost/pull/2944

THis will enable openshift cluster users to configure installing kubecost while using their in-cluster prometheus.

Closes #3690

NA

## How was this PR tested?
tested by installing kubecost using custom built cost-model image having [these](https://github.com/opencost/opencost/pull/2944) changes while disabling the bundled prometheus and using the in-cluster prometheus. 

https://github.com/kubecost/docs/pull/1144/files

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
@@ -33,9 +33,9 @@ subjects:
   - kind: ServiceAccount
     name: {{ template "cost-analyzer.serviceAccountName" . }}
     namespace: {{ .Release.Namespace }} 
-{{- end }} 
 ---
-{{- if and (.Values.platforms.opensift.enabled .Values.platforms.opensift.createPrometheusClusterRoleBinding) }}
+{{- end }} 
+{{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.createPrometheusClusterRoleBinding }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
@@ -34,3 +34,20 @@ subjects:
     name: {{ template "cost-analyzer.serviceAccountName" . }}
     namespace: {{ .Release.Namespace }} 
 {{- end }} 
+---
+{{- if and (.Values.platforms.opensift.enabled .Values.platforms.opensift.createPrometheusClusterRoleBinding) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "cost-analyzer.serviceAccountName" . }}-cluster-monitoring-operator
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-monitoring-operator
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "cost-analyzer.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }} 
+{{- end }}
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
@@ -35,7 +35,7 @@ subjects:
     namespace: {{ .Release.Namespace }} 
 ---
 {{- end }} 
-{{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.createPrometheusClusterRoleBinding }}
+{{- if and .Values.global.platforms.openshift.enabled .Values.global.platforms.openshift.createMonitoringClusterRoleBinding }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
@@ -39,13 +39,13 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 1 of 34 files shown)


jessegoodier:
This looks good to me. Let's merge when the opencost PR is merged.

JesseGoodier (approved):
LGTM!

cost-analyzer/templates/monitoring-role-binding-template.yaml
```diff
@@ -0,0 +1,17 @@
...
+{{- if (.Values.global.platforms.openshift.createMonitoringResourceReaderRoleBinding) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ template "cost-analyzer.fullname" . }}-reader
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.global.platforms.openshift.monitoringServiceAccountName | quote }}
+  namespace: openshift-monitoring
```
thomasvn, line 11: Is it safe to assume that the ServiceAccount we want to bind to will always be in `openshift-monitoring`?

cost-analyzer/values-openshift.yaml
```diff
@@ -1,9 +1,19 @@
...
+    # kubeRBACProxy: false # If true, kubecost will use kube-rbac-proxy to authenticate with in cluster Prometheus for openshift
+
   # Platforms is a higher-level abstraction for platform-specific values and settings.
   platforms:
     # Deploying to OpenShift (OCP) requires enabling this option.
     openshift:
       enabled: true  # Deploy Kubecost to OpenShift.
+      # createMonitoringClusterRoleBinding: false  # Create a Cluster Role Binding to allow using in-cluster prometheus or thanos.
+      # createMonitoringResourceReaderRoleBinding: false  # Create a Role and Role Binding to allow in-cluster prometheus or thanos to list and watch resources. This will be necessary if you are not using bundled prometheus and need to add scrape config for resources.
+      # monitoringServiceAccountName: prometheus-k8s  # Name of the service account to bind to the Resource Reader Role Binding.
```
thomasvn, line 16: We should add these comments to the main `values.yaml` file as well!

thomasvn: If possible, I think we should also uncomment them and set default values when possible. Example:

```yaml
  # Platforms is a higher-level abstraction for platform-specific values and settings.
  platforms:
    # Deploying to OpenShift (OCP) requires enabling this option.
    openshift:
      enabled: true  # Deploy Kubecost to OpenShift.
      createMonitoringClusterRoleBinding: false  # Create a Cluster Role Binding to allow using in-cluster prometheus or thanos.
      createMonitoringResourceReaderRoleBinding: false  # Create a Role and Role Binding to allow in-cluster prometheus or thanos to list and watch resources. This will be necessary if you are not using bundled prometheus and need to add scrape config for resources.
      monitoringServiceAccountName: prometheus-k8s  # Name of the service account to bind to the Resource Reader Role Binding.
```

cost-analyzer/templates/monitoring-role-binding-template.yaml
mittal-ishaan: No, you are correct here. It is not safe to assume the namespace. I would have to correct this.

cost-analyzer/values-openshift.yaml
mittal-ishaan: sure, will uncomment these. I added these specific to openshift and should make it more generalized. Should I move these to global instead? instead of having them under openshift?

jessegoodier:
Good finds Thomas. 

Ishaan- we merged because we have an urgent need with some customers and needed to test. Thank you for getting this done so quickly!

cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
```diff
@@ -33,4 +33,21 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "cost-analyzer.serviceAccountName" . }}-openshift-monitoring
+  labels:
+    {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-monitoring-view
```
thomasvn, line 48: What ClusterRole is this referring to? I couldn't find any reference to this in the Helm chart. Is this an existing ClusterRole that's specific to OpenShift installs?

mittal-ishaan: yes its specific to openshift installs.
https://docs.redhat.com/en/documentation/openshift_container_platform/4.2/html/monitoring/cluster-monitoring#monitoring-accessing-prometheus-alerting-ui-grafana-using-the-web-console_accessing-prometheus

thomasvn: Awesome, thanks for clarifying! Lets leave a comment in the template?

mittal-ishaan: sure :+1:

mittal-ishaan:
Thank you Thomas for pointing these out.
I will create a PR to fix these today itself.

PR #3717, opened 2024-10-17
Add logLevel as a helm value property under kubecostModel instead of extraEnv configuration

mittal-ishaan:
## What does this PR change?
Adds log level as a helm value property for kubecostModel to make it easier to be configured by the user. 

NA

Adds log level as a helm value property for kubecostModel to make it easier to be configured by the user. Right now the user has to add an extraEnv for log level to make it work out.

NA

## How was this PR tested?
deployed kubecost by setting the helm values as
```yaml
kubecostModel:
  logLevel: "debug"
```
I was able to see debug logs in cost-model container

Not yet but will do

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -740,7 +740,7 @@ spec:
             - name: GRAFANA_ENABLED
               value: "{{ template "cost-analyzer.grafanaEnabled" . }}"
             - name: LOG_LEVEL
-              value: {{ .Values.kubecostAggregator.logLevel }}
+              value: {{ .Values.kubecostModel.logLevel }}
             {{- end}}
             {{- if .Values.kubecostModel.extraEnv -}}
             {{ toYaml .Values.kubecostModel.extraEnv | nindent 12 }}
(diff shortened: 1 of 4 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -739,6 +739,8 @@ spec:
             {{- if .Values.global.grafana }}
             - name: GRAFANA_ENABLED
               value: "{{ template "cost-analyzer.grafanaEnabled" . }}"
+            - name: LOG_LEVEL
+              value: {{ .Values.kubecostModel.logLevel | default "info" }}
```
thomasvn: @mittal-ishaan I'd prefer we not set any default values in the template files. Instead I prefer all defaults be set via the `values.yaml` file (which you've already done).

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -739,6 +739,8 @@ spec:
             {{- if .Values.global.grafana }}
             - name: GRAFANA_ENABLED
               value: "{{ template "cost-analyzer.grafanaEnabled" . }}"
+            - name: LOG_LEVEL
+              value: {{ .Values.kubecostModel.logLevel }}
```
thomasvn, line 744: @mittal-ishaan Is there a reason why this `LOG_LEVEL` env var is within an `if` statement here? We should probably be setting this outside the `if` statement.

mittal-ishaan: my bad here.  I should have been more careful. I will correct it.

thomasvn (approved):

PR #3721, opened 2024-10-25
fix openshift values to default with in-cluster prometheus

mittal-ishaan:
## What does this PR change?
* This PR adds an example values file to have default properties set to deploy kubecost with in-cluster prometheus. (As this is what most users want)
* Add the general configurations to default values file and remove unnecessary configurations and comments from openshift and openshift prometheus example values file.
* Adds clarifying comment in the template while binding the cluster role to the kubecost service account
* Moves in-cluster prometheus service account namespace name to values to make it more generalized and to cover the case where it can be of any other name.
No

Added standard configuration template for openshift to use the openshift-monitoring prometheus and cleaned up openshift example values file.

No risk, Testing that it gets deployed on openshift cluster easily and successfully need to be done.

## How was this PR tested?
Installing kubecost in openshift environment with network cost enabled without helm hook:
Screenshot from 2024-10-26 02-54-50
Installing with pre-install helm hook
Screenshot from 2024-10-26 02-59-27
Resulting kubecost diagnostics page
image

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
@@ -43,6 +43,9 @@ metadata:
   labels:
     {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
 roleRef:
+  # Grant the kubecost service account the cluster-monitoring-view role to enable it to query OpenShift Prometheus.
+  # This is necessary for Kubecost to get access and query the in-cluster Prometheus instance using its service account token.
+  # https://docs.redhat.com/en/documentation/openshift_container_platform/4.2/html/monitoring/cluster-monitoring#monitoring-accessing-prometheus-alerting-ui-grafana-using-the-web-console_accessing-prometheus
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: cluster-monitoring-view
--- a/cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
+++ b/cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
@@ -3,12 +3,6 @@ apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
   name: {{ template "cost-analyzer.networkCostsName" . }}
-  # Add a pre-install hook to the Security Context Constraints (SCC) to ensure it is created before the network-costs DaemonSet.
-  # This is crucial because the DaemonSet relies on the SCC for proper security configurations.
-  # By using a pre-install hook, we guarantee that the SCC is in place before the DaemonSet is deployed, preventing potential installation failures.
-  # https://helm.sh/docs/topics/charts_hooks/#the-available-hooks
-  annotations:
-    helm.sh/hook: pre-install
 priority: 10
 allowPrivilegedContainer: true
 allowHostDirVolumePlugin: true
--- a/cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
+++ b/cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
@@ -3,6 +3,10 @@ apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
   name: {{ template "cost-analyzer.networkCostsName" . }}
+  # Add a pre-install hook to the Security Context Constraints (SCC) to ensure it is created before the network-costs DaemonSet.
+  # This is crucial because the DaemonSet relies on the SCC for proper security configurations.
+  # By using a pre-install hook, we guarantee that the SCC is in place before the DaemonSet is deployed, preventing potential installation failures.
+  # https://helm.sh/docs/topics/charts_hooks/#the-available-hooks
   annotations:
     helm.sh/hook: pre-install
 priority: 10
--- a/cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
+++ b/cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
@@ -3,6 +3,8 @@ apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
   name: {{ template "cost-analyzer.networkCostsName" . }}
+  annotations:
+    helm.sh/hook: pre-install
 priority: 10
 allowPrivilegedContainer: true
 allowHostDirVolumePlugin: true
--- a/cost-analyzer/values-openshift-cluster-prometheus.yaml
+++ b/cost-analyzer/values-openshift-cluster-prometheus.yaml
@@ -22,7 +22,6 @@ global:
 
       # Create Security Context Constraint resources for the DaemonSets requiring additional privileges.
       scc:
-        nodeExporter: false  # Creates an SCC for Prometheus Node Exporter. This requires Node Exporter be enabled.
         networkCosts: false  # Creates an SCC for Kubecost network-costs. This requires network-costs be enabled.
       # When OpenShift is enabled, the following securityContext will be applied to all resources unless they define their own.
       securityContext:
@@ -37,13 +36,8 @@ global:
 #     enabled: true  # Enable Prometheus Node Exporter.
 serviceMonitor:
   enabled: true
-  # additionalLabels:
-  #   label-key: label-value
   # networkCosts:
   #   enabled: false
-    # additionalLabels:
-    #   label-key: label-value
+
 prometheusRule:
   enabled: true
-  # additionalLabels:
-  #   label-key: label-value
--- a/cost-analyzer/values-openshift-cluster-prometheus.yaml
+++ b/cost-analyzer/values-openshift-cluster-prometheus.yaml
@@ -1,0 +1,54 @@
+# This helm values file is a modified version of the values-openshift.yaml file that is used to deploy Kubecost on OpenShift. 
+# The main difference is that the values-openshift-cluster-prometheus.yaml file is configured to connect to the in-cluster Prometheus instance running on the OpenShift cluster and not use bundled prometheus. 
+
+global:
+  prometheus:
+    enabled: false  # Kubecost depends on Prometheus data, it is not optional. When enabled: false, Prometheus will not be installed and you must configure your own Prometheus to scrape kubecost as well as provide the fqdn below. -- Warning: Before changing this setting, please read to understand the risks https://docs.kubecost.com/install-and-configure/install/custom-prom
+    fqdn: https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091  # example address of a Prometheus to connect to. Include protocol (http:// or https://) Ignored if enabled: true
+    insecureSkipVerify: false  # If true, kubecost will not check the TLS cert of prometheus
+    # queryServiceBearerTokenSecretName: mcdbsecret  # kubectl create secret generic mcdbsecret -n kubecost --from-file=TOKEN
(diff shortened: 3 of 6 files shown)


chipzoller (commented):
I'm wondering if this is the right approach (on by default) as it deviates from our standard architecture. How extensively has this been tested? Do you have any concerns here, @jessegoodier and @kwombach12 ?

jessegoodier:
> I'm wondering if this is the right approach (on by default) as it deviates from our standard architecture. How extensively has this been tested? Do you have any concerns here, @jessegoodier and @kwombach12 ?

This is what users are asking for, and I do think it is the best architecture.

A consistent configuration will reduce support concerns with openshift- in comparison to what we have seen with the various custom configs attempted.

Historically, issues with prometheus are due to misconfiguration, not version.

cost-analyzer/templates/cost-analyzer-cluster-role-binding-template.yaml
```diff
@@ -43,6 +43,9 @@ metadata:
   labels:
     {{ include "cost-analyzer.commonLabels" . | nindent 4 }}
 roleRef:
+  # Grant the kubecost service account the cluster-monitoring-view role to enable it to query OpenShift Prometheus.
+  # This is necessary for Kubecost to get access and query the in-cluster Prometheus instance using its service account token.
+  # https://docs.redhat.com/en/documentation/openshift_container_platform/4.2/html/monitoring/cluster-monitoring#monitoring-accessing-prometheus-alerting-ui-grafana-using-the-web-console_accessing-prometheus
```
thomasvn, line 48: Very helpful comment!

cost-analyzer/values-openshift.yaml
thomasvn, line 1: Kubecost is always recommended to be run with the helm chart's bundled Prometheus. This is usually the best user experience and results in the fewest issues.

Although we should certainly document how to run with Openshift's pre-installed Prometheus, I don't think that we should default to that experience.

thomasvn:
Also any new configurations that you've added into `values-openshift.yaml` should also be in `values.yaml`. Helm will always use the `values.yaml` as the default configuration for the app ([docs ref](https://helm.sh/docs/chart_template_guide/values_files/)).

For example the following install command will use all default values from the `values.yaml` in the chart:

```
helm install kubecost cost-analyzer \
--repo https://kubecost.github.io/cost-analyzer/ \
--namespace kubecost --create-namespace
```

cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
```diff
@@ -3,6 +3,12 @@ apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
   name: {{ template "cost-analyzer.networkCostsName" . }}
+  # Add a pre-install hook to the Security Context Constraints (SCC) to ensure it is created before the network-costs DaemonSet.
+  # This is crucial because the DaemonSet relies on the SCC for proper security configurations.
+  # By using a pre-install hook, we guarantee that the SCC is in place before the DaemonSet is deployed, preventing potential installation failures.
+  # https://helm.sh/docs/topics/charts_hooks/#the-available-hooks
+  annotations:
+    helm.sh/hook: pre-install
```
chipzoller: Keep in mind that some vendor marketplaces and other delivery form factors do not support various Helm hooks and so this is more surface area that will have to be removed when these listings are updated.

cost-analyzer/values-openshift.yaml
chipzoller: Additionally, when users choose not to deploy Kubecost's bundled Prometheus they will not get the [Kubecost-curated extraScrapeConfigs](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L756C3-L806) which, among them, newly include targets for NVIDIA's DCGM Exporter. I am willing to bet this discrepancy will be the root cause of many future support cases.

cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
mittal-ishaan: ohh, okay understood. Will try to find a better alternative for this then

cost-analyzer/values-openshift.yaml
jessegoodier: Can we add create a service monitor with the needed scrape config for DCGM? 

Perhaps we should have two value files to avoid requests asking for how to correctly implement using openshift prometheus?

Do we need additional tests to validate the non-bundled prometheus? Are we missing anything today that diagnostics would miss?

cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
jessegoodier: the hook here is simply fixing a warning during helm install. As long as the vendor marketplaces do not fail when they see the annotation, it should work as you have done it.

cost-analyzer/values-openshift.yaml
```diff
@@ -1,35 +1,51 @@
...
       # When OpenShift is enabled, the following securityContext will be applied to all resources unless they define their own.
       securityContext:
         runAsNonRoot: true
         seccompProfile:
           type: RuntimeDefault
 
-# networkCosts:
-#   enabled: true  # Enable network costs.
+networkCosts:
+  enabled: true  # Enable network costs.
```
chipzoller: In my opinion, we shouldn't be enabling optional components by default in OpenShift that aren't enabled by default for other platforms. We should be consistent as far as which components are on by default and those which are not.

cost-analyzer/values-openshift.yaml
```diff
@@ -1,35 +1,51 @@
...
+    proxy: false
   # Platforms is a higher-level abstraction for platform-specific values and settings.
   platforms:
     # Deploying to OpenShift (OCP) requires enabling this option.
     openshift:
       enabled: true  # Deploy Kubecost to OpenShift.
-      # createMonitoringClusterRoleBinding: false  # Create a Cluster Role Binding to allow using in-cluster prometheus or thanos.
-      # createMonitoringResourceReaderRoleBinding: false  # Create a Role and Role Binding to allow in-cluster prometheus or thanos to list and watch resources. This will be necessary if you are not using bundled prometheus and need to add scrape config for resources.
-      # monitoringServiceAccountName: prometheus-k8s  # Name of the service account to bind to the Resource Reader Role Binding.
+      createMonitoringClusterRoleBinding: true  # Create a Cluster Role Binding to allow using in-cluster prometheus or thanos.
```
chipzoller:
```suggestion
      createMonitoringClusterRoleBinding: true  # Create a ClusterRoleBinding to allow using in-cluster Prometheus or Thanos.
```

Etc. elsewhere

cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
chipzoller: AWS Marketplace as one example will fail when it sees a chart with a Helm hook as they are not supported. See [here](https://docs.aws.amazon.com/marketplace/latest/userguide/container-product-policies.html#publishing-eks-add-on) for details. Just saying be mindful that when preparing the AWS Marketplace (possibly others) version, this is something that can't be present.

> Helm hooks and the `lookup` function are not supported.

cost-analyzer/values-openshift.yaml
```diff
@@ -1,35 +1,51 @@
...
     fqdn: https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091  # example address of a prometheus to connect to. Include protocol (http:// or https://) Ignored if enabled: true
-    # insecureSkipVerify: false  # If true, kubecost will not check the TLS cert of prometheus
+    insecureSkipVerify: false  # If true, kubecost will not check the TLS cert of prometheus
     # queryServiceBearerTokenSecretName: mcdbsecret  # kubectl create secret generic mcdbsecret -n kubecost --from-file=TOKEN
-    # kubeRBACProxy: false # If true, kubecost will use kube-rbac-proxy to authenticate with in cluster Prometheus for openshift
-
+    kubeRBACProxy: true # If true, kubecost will use kube-rbac-proxy to authenticate with in cluster Prometheus for openshift
+  grafana:
+    enabled: false  # If false, Grafana will not be installed
```
chipzoller: Similarly here we're disabling Grafana by default just for OpenShift whereas it's enabled for all other platform types which creates an inconsistent deployment topology.

cost-analyzer/values-openshift.yaml
jessegoodier: agree, my mistake here.

cost-analyzer/values-openshift.yaml
jessegoodier: Will leave values-openshift.yaml with the existing defaults and create a new file with the values that Ishaan has here. Stay tuned.

cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
jessegoodier: thank you. I'm good with removing the hood and having the warning. we can add a NOTES to ignore the warning.

cost-analyzer/values-openshift.yaml
chipzoller: Yeah, I think that's a better approach. Definitely glad to see this though! Looking forward to also seeing this [in the docs](https://github.com/kubecost/docs/pull/1144) referencing the new values file!

thomasvn: @jessegoodier I like that approach!

cost-analyzer/values-openshift.yaml
mittal-ishaan: commenting out network costs configuration

cost-analyzer/templates/cost-analyzer-networks-costs-ocp-scc.yaml
mittal-ishaan: removing the hook

cost-analyzer/values-openshift-cluster-prometheus.yaml
```diff
@@ -0,0 +1,54 @@
+# This helm values file is a modified version of the values-openshift.yaml file that is used to deploy Kubecost on OpenShift. 
+# The main difference is that the values-openshift-cluster-prometheus.yaml file is configured to connect to the in-cluster Prometheus instance running on the OpenShift cluster and not use bundled prometheus.
```
thomasvn:
```suggestion
# This Helm values file is a modified version of `values-openshift.yaml`.
# The primary difference is that this file is configured to disable the Kubecost-bundled Prometheus, and instead leverage the Prometheus instance that is typically pre-installed in OpenShift clusters.
```

cost-analyzer/values-openshift-cluster-prometheus.yaml
thomasvn, line 1: Please also remove any excess comments & unused configs which are present in this file.

thomasvn:
@mittal-ishaan and I discussed this PR on a call. We made the following changes together:

* The `values.yaml` should be the source of truth, and should capture all possible configurations. It should also contain detailed documentation on what each of the configurations do.
* The `values-openshift.yaml` is an example of the minimal set of configurations needed to deploy Kubecost on OpenShift assuming you are also deploying Kubecost's bundled Prometheus
* The `values-openshift-cluster-prometheus.yaml` is an example of the minimal set of configurations needed to deploy Kubecost on OpenShift assuming you are leveraging a Prometheus instance which already exists in the cluster.

thomasvn (approved):

thomasvn:
@mittal-ishaan @jessegoodier I think we should cherry-pick this into the v2.5 branch so that it gets released with v2.5.0? Thoughts?

jessegoodier:
100%

PR #3725, opened 2024-10-29
Update comments about OIDC RBAC groups.

thomasvn:
## What does this PR change?

More accurate comments regarding OIDC RBAC groups. Previously, the comments were identical to those in SAML RBAC. However, the underlying implementations differ.

SAML allows all authenticated users through on read routes even if they don't belong to a group. It will only enforce groups if the `readonly` group is enabled ([code ref](https://github.com/kubecost/kubecost-cost-model/blob/0d344ebbad1fdd0654996fcdb87d36249a2e8465/pkg/auth/samlauth.go#L184-L197)).

OIDC blocks users from entry to the application if they do not belong to a valid group ([code ref](https://github.com/kubecost/kubecost-cost-model/blob/0d344ebbad1fdd0654996fcdb87d36249a2e8465/pkg/auth/oidcauth.go#L210-L213)).

N/A

N/A

N/A

N/A

## How was this PR tested?

N/A

N/A

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -374,19 +374,19 @@ oidc:
   rbac:
     enabled: false
     # groups:
-    #   - name: admin
-    #     enabled: false  # If admin is disabled, all authenticated users will be able to make configuration changes to the kubecost frontend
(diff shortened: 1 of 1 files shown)


ivankube (approved):
LGTM

PR #3763, opened 2024-12-05
add init container for updating ca trust and shift getting ca cert from secret to config map

mittal-ishaan:
## What does this PR change?
* Create a init container to update ca trust which we earlier expected the user to do inside the container itself.
Taken reference from https://jfrog.com/help/r/artifactory-how-to-add-custom-ssl-certificates-to-an-artifactory-pod-using-helm-charts/artifactory-how-to-add-custom-ssl-certificates-to-an-artifactory-pod-using-helm-charts

https://github.com/kubecost/cost-analyzer-helm-chart/pull/3760

Users will now not need to run update-ca-trust command inside the container. The init container will do it.

## How was this PR tested?
Deployed the kubecost deployment with `caCertsSecret` helm value set and ca cert as Secret created.
reading through `/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt` we see it updated with the given ca cert in Secret.

helm values:

```yaml
global:
  updateCaTrust:
    enabled: false  # Set to true to enable the init container for updating CA trust
    # Security context settings for the init container.
    securityContext:
      runAsUser: 0
      runAsGroup: 0
      runAsNonRoot: false
      readOnlyRootFilesystem: true
      seccompProfile:
        type: RuntimeDefault
    caCertsSecret: ca-certs-secret  # The name of the Secret containing custom CA certificates to mount to the cost-model container.
    # caCertsConfig: ca-certs-config  # The name of the ConfigMap containing the CA trust configuration.
    resources: {}  # Resource requests and limits for the init container.
    caCertsMountPath: /etc/pki/ca-trust/source/anchors  # The path where the custom CA certificates will be mounted in the init container

```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -126,11 +126,12 @@ spec:
            defaultMode: 420
            secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret | default "federated-store" }}
         {{- end }}
-        {{- if .Values.kubecostModel.caCertsSecret }}
-        - name: ca-certs-secret
-          secret:
-           defaultMode: 420
-           secretName: {{ .Values.kubecostModel.caCertsSecret}}
+        {{- if .Values.kubecostModel.caCertsConfig }}
+        - name: ca-certs-config
+          configMap:
+            name: {{ .Values.kubecostModel.caCertsConfig}}
+        - name: ssl-path
+          emptyDir: {}
         {{- end }}
         {{- if .Values.kubecostProductConfigs }}
         {{- if and ((.Values.kubecostProductConfigs).productKey).enabled ((.Values.kubecostProductConfigs).productKey).secretname }}
@@ -351,6 +352,28 @@ spec:
           securityContext:
             runAsUser: 0
 {{ end }}
+        {{- if .Values.kubecostModel.caCertsConfig }}
+        - name: "sslsetup"
+          image: redhat/ubi9-minimal:latest
+          imagePullPolicy: Always
+          securityContext:
+            privileged: true
+            runAsUser: 0
+            runAsGroup: 0
+            runAsNonRoot: false
+          command:
+            - 'sh'
+            - '-c'
+            - >
+              mkdir -p /etc/pki/ca-trust/extracted/{edk2,java,openssl,pem};
+              /usr/bin/update-ca-trust extract;
+          volumeMounts:
+            - name: ca-certs-config
+              mountPath: "/etc/pki/ca-trust/source/anchors"
+            - name: ssl-path
+              mountPath: "/etc/pki/ca-trust/extracted"
+              readOnly: false
+              {{- end}}
       containers:
         {{- if .Values.global.gmp.enabled }}
         - name: {{ .Values.global.gmp.gmpProxy.name }}
@@ -621,9 +644,12 @@ spec:
               mountPath: /var/configs/etl/federated
               readOnly: true
             {{- end }}
-            {{- if .Values.kubecostModel.caCertsSecret }}
-            - name: ca-certs-secret
-              mountPath: /etc/pki/ca-trust/source/anchors
+            {{- if .Values.kubecostModel.caCertsConfig }}
+            - name: ca-certs-config
+              mountPath: "/etc/pki/ca-trust/source/anchors"
+            - name: ssl-path
+              mountPath: "/etc/pki/ca-trust/extracted"
+              readOnly: false
             {{- end }}
             {{- if .Values.kubecostAdmissionController }}
             {{- if .Values.kubecostAdmissionController.enabled }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -126,15 +126,15 @@ spec:
            defaultMode: 420
            secretName: {{ .Values.kubecostModel.federatedStorageConfigSecret | default "federated-store" }}
         {{- end }}
-        {{- if .Values.kubecostModel.updateCaTrust.enabled }}
+        {{- if .Values.global.updateCaTrust.enabled }}
         - name: ca-certs-secret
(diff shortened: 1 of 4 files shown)


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -351,6 +353,28 @@ spec:
           securityContext:
             runAsUser: 0
 {{ end }}
+        {{- if .Values.kubecostModel.caCertsSecret }}
+        - name: "sslsetup"
```
jessegoodier:
```suggestion
        - name: "update-ca-trust"
```
what do you think of changing the container name?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -351,6 +353,28 @@ spec:
...
+          image: {{ include "cost-model.image" . | trim | quote}}
+          imagePullPolicy: Always
+          securityContext:
```
jessegoodier: can we template this and add these as the defaults in values.yaml?

jessegoodier: thinking `kubecostModel.update-ca-trust.securityContext`

mittal-ishaan: agree here

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
mittal-ishaan: agree

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -351,6 +353,30 @@ spec:
           securityContext:
             runAsUser: 0
 {{ end }}
+        {{- if .Values.kubecostModel.updateCaTrust.enabled }}
+        - name: "update-ca-trust"
+          image: {{ include "cost-model.image" . | trim | quote}}
+          imagePullPolicy: Always
```
jessegoodier, line 367:
```suggestion
          imagePullPolicy: IfNotPresent
```

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -351,6 +353,30 @@ spec:
           securityContext:
             runAsUser: 0
 {{ end }}
+        {{- if .Values.kubecostModel.updateCaTrust.enabled }}
+        - name: "update-ca-trust"
```
jessegoodier:
```suggestion
        - name: update-ca-trust
```

jessegoodier:
This is looking good to me. Tested with custom secret name.

Will ping customer and merge after we get feedback.

chipzoller:
But why a secret? Root certificates aren't sensitive.

cost-analyzer/values.yaml
```diff
@@ -575,8 +575,19 @@ kubecostModel:
...
 
-  # the name of the Secret containing custom CA certs to mount to cost model container
-  # caCertsSecret: ca-certs-secret
+  # Install the cost-model with a custom CA certs.
+  # The following configuration is for an init container to update the CA trust.
+  updateCaTrust:
+    enabled: false  # Set to true to enable the init container for updating CA trust
+    # Security context settings for the init container.
+    securityContext:
+      privileged: true
```
```suggestion
```

jessegoodier: looks like it works without this line?

mittal-ishaan:
@chipzoller The user requested it to be a secret as they store/retrieve certs as secrets internally.

chipzoller:
Then maybe make this so it accepts a Secret or ConfigMap?

mittal-ishaan:
sure, we should do that. let me get that in.

thomasvn:
Future TODO. This updates the CA trust and certificates for the cost-analyzer pod. But it's likely that users would need to perform the same updates across other pods like the aggregator and cloudcost pod.

jessegoodier:
@mittal-ishaan great work here. Thanks for all of the iterations while we figured out the best approach.

Agree with Thomas that we should add to aggregator and cloud cost when you have a moment.

I'm now questioning if we should have put this under a more general section instead of nesting under `kubecostModel` ?

@chipzoller and @thomasvn what are your thoughts?

thomasvn:
Yes! I like the idea of doing `.Values.updateCaTrust` instead of `.Values.kubecostModel.updateCaTrust`.

chipzoller:
Also agree that I'd try and pull this out and generalize it, but I'm not sure the root level is the best place.

You guys have any other suggestions for where these values may best go?

thomasvn:
Only other place that seems suitable is under `.Values.global`. I'm open to other ideas though?

jessegoodier:
> Only other place that seems suitable is under `.Values.global`. I'm open to other ideas though?

I'm thinking `global.` is better than `productConfigs`

cost-analyzer/values.yaml
```diff
@@ -226,6 +226,24 @@ global:
       drop:
       - ALL
 
+  # Install the cost-model with a custom CA certs.
+  # The following configuration is for an init container to update the CA trust.
```
```suggestion
  # Installs custom CA certificates onto Kubecost pods
```

thomasvn (approved):
This looks good to me. These configs are now available under `.Values.global.updateCaTrust`.

Approving for immediate merge so fixes are available to customer. However we now need a follow-up PR to make similar changes to the Kubecost Aggregator and Kubecost CloudCost pods, then test those changes.

jessegoodier:
Thanks Thomas. great work here team.

srpomeroy:
Did we ever add docs for this feature?

thomasvn:
@srpomeroy Thanks for following up. I'm not sure we ever did.

@mittal-ishaan Could you please create a Docs PR to describe how to configure `.Values.global.updateCaTrust`? It may end up being a new markdown page under the "Advanced Configuration" section [here](https://github.com/kubecost/docs/blob/main/install-and-configure/advanced-configuration/README.md). Alternatively, you could add more robust comments about how to configure this directly into the `values.yaml` file.

PR #3780, opened 2024-12-17
fix log message

JesseGoodier:
## What does this PR change?
add emptydir to nginx log path to remove log warning

NA

Fix nginx log warning with emptyDir to remove log warning

NA


jessegoodier:
going to fix dockerfile instead

Issue #3727, opened 2024-10-31
[Bug] No api service on 9001 - crashloop


matt-tvg:
### Kubecost Helm Chart Version

2.4.1

### Kubernetes Version

v1.31.1-eks-ce1d5eb 

### Kubernetes Platform

EKS

### Description

On fresh deployments`cost-analyzer-frontend` creash loops with the error:

```
nginx: [emerg] host not found in upstream "cost-analyzer.build-ci-kubecost:9003" in /etc/nginx/conf.d/default.conf:42  
``` 

The config map for the nginx config shows the upstream configured as i'd expect:

```
upstream api {                                                                                                                                               │
    server cost-analyzer.build-ci-kubecost:9001;                                                                                                             │
}
```

However there doesnt seem to be a service running on 9001:

```
 kubectl get services -n build-ci-kubecost --sort-by=.metadata.name
NAME                        TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)             AGE
cost-analyzer               ClusterIP   172.20.104.180   <none>        9003/TCP,9090/TCP   24m
cost-analyzer-aggregator    ClusterIP   172.20.120.179   <none>        9004/TCP            24m
cost-analyzer-cloud-cost    ClusterIP   172.20.193.29    <none>        9005/TCP            24m
cost-analyzer-forecasting   ClusterIP   172.20.152.48    <none>        5000/TCP            24m
```

Out of curiousity i tried setting it to 9090 via the below:

```
  useDefaultFqdn: false
  api:
    fqdn: cost-analyzer.build-ci-kubecost:9090
```

However that gave a similar error:

```
nginx: [emerg] host not found in upstream "cost-analyzer.build-ci-kubecost:9090" in /etc/nginx/conf.d/default.conf:38
```                                  

### Steps to reproduce

1. Fresh helm installation against an EKS cluster v1.31

### Expected behavior

all pods to become ready and a frontend to be available on 9090

### Impact

cannot access

### Troubleshooting

- [X] I have read and followed the [issue guidelines](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/ISSUE_GUIDELINES.md) and this is a bug impacting only the Helm chart.
- [X] I have searched other issues in this repository and mine is not recorded.

chipzoller:
Would you please provide the values you used for installation?

matt-tvg:
Hi,

templated values are below :) 

```yaml
global:
  prometheus:
[...]

PR #3781, opened 2024-12-17
Remove-old-api-port

JesseGoodier:
## What does this PR change?
Removes reference to old /api nginx config pointing to port 9001 that was removed in 2.0

NO

Removes reference to old /api nginx config pointing to port 9001 that was removed in 2.0

https://github.com/kubecost/cost-analyzer-helm-chart/issues/3727

No risk from what I can find. Also confirmed with Bianca

## How was this PR tested?
QA test on both sts and single pod aggregator. 

NA

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -216,17 +216,7 @@ data:
         listen [::]:{{ .Values.service.targetPort }};
 {{- end }}
 {{- end }}
-        location /api/ {
-            {{- if or .Values.saml.enabled .Values.oidc.enabled }}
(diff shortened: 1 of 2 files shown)


jessegoodier:
@biancaburtoiu I believe when we removed federator, we removed 9001 /api
I do not see any calls to it in the UI. Any place else you can think of to check?

jessegoodier:
We discussed as a team, we all believe it was just overlooked in the 2.0 migration.

ameijer (approved):
yeah I think this was downright misleading. still concerned though that it only dies sometimes with invalid upstreams as in https://github.com/kubecost/cost-analyzer-helm-chart/issues/3727

biancaburtoiu (approved):
LGTM

PR #3794, opened 2024-12-27
Update ca cert in cloud cost and aggregator container too

mittal-ishaan:
## What does this PR change?
Update ca cert in cloud cost and aggregator container too

https://github.com/kubecost/cost-analyzer-helm-chart/pull/3763

Users will now be able update or add custom ca certs in aggregator and cloud cost containers too as they were able to do it for cost-model container.

https://github.com/kubecost/cost-analyzer-helm-chart/pull/3763#issuecomment-2532286874

## How was this PR tested?
1. For single Pod deploy method checked in Aggregator, Cloud cost and cost model container if the custom ca cert was added
2. For statefulSet deploy method checked in all three pods containers if the custom ca cert was added
v2.6


JesseGoodier (approved):
Nice work as always!

PR #3804, opened 2025-01-09
Add sample limit to service monitor

pomyslowynick:
## What does this PR change?

It adds a new value to the helm chart

No

It adds a new value to the helm chart

No risks, it's a simple change


jessegoodier:
Closing as the risk outweigh benefits. Please correct me if I am not understanding this correctly.

pomyslowynick:
> Closing as the risk outweigh benefits. Please correct me if I am not understanding this correctly.

What are the risks here? It's exposing a field on a service monitor, it doesn't affect anything unless this value is explicitly set, or if the default is not set to allow unlimited number of samples.

jessegoodier:
If a user sees the option, they may set it. What would the outcome be if the limit is enforced?

We have plans around prometheus configuration changes in 3.0 that will likely make the serviceMonitor obsolete.

PR #3807, opened 2025-01-14
when cloud costs is not enabled, don't deploy it

ameijer:
## What does this PR change?
changes the scope of the if cloud costs enabled control statement to avoid rendering the cloud costs deployment (thereby installing it), when cloud costs is disabled

no

should allow them to not deploy cloud costs if they don't want to

Low risk

## How was this PR tested?
tested via argo install on clickhouse demo env


jessegoodier:
I think that this is the right design, but the frontend would need an update to prevent the user from adding a cloud integration and instruct them on the best practices?

ameijer:
closing since there are FE/product concerns

PR #3810, opened 2025-01-16
fix: nginx read only file issue for ubi  images for frontend images

mittal-ishaan:
## What does this PR change?
Mount an emptyDir to the /var/log/nginx directory in the frontend deployment to resolve access issues caused by a read-only file system in the new UBI-based frontend images.

https://github.com/kubecost/kubecost-frontend/pull/2125/files
https://github.com/kubecost/kubecost-frontend/pull/2109

Users would be able to deploy kubecost using the UBI frontend image which earlier went on a `CrashLoopBackOff`

No

## How was this PR tested?
Deployed kubecost using these changes and frontend UBI image.

NA


jessegoodier:
Plz merge when we move to UBI base image.

JesseGoodier (approved):
Thanks Ishaan- merging now as there is no harm in this either way.

PR #3812, opened 2025-01-16
fix: add missing url in grafana datasource

takirala:
## What does this PR change?

`grafana-datasource` configmap is being created with invalid datasource as the url is empty. I think this was an overlook when thanos support was removed (in https://github.com/kubecost/cost-analyzer-helm-chart/commit/279d7beabc3c40627944a666726f92338a888ee0#diff-406c7314a1029b6fe683e8a2c4bf1f7f52222794f542eb30a5a0aec78e022e28). Please let me know if I am missing some context.

- Fixes the url in grafana datasource configmap when using an external grafana using the sidecar.

- Create a kubecost cluster by disabling in-built grafana and configure an external grafana. The datasource will be created with an invalid value. Adding the fix in this PR will make the datasource work correctly.

## How was this PR tested?

I updated the configmap for grafana-datasource and was able to browse grafana correctly.


takirala:
Ping @jessegoodier @ivankube @cliffcolvin PTAL 🙏 
I am not sure i pinged the right people, please redirect as needed, thank you!

JesseGoodier (approved):

jessegoodier:
Good find @takirala !

PR #3816, opened 2025-01-22
Route /model auth endpoints to the aggregator

kaelanspatel:
## What does this PR change?
Any `/model` routes relating to auth now go to the aggregator instead of cost-analyzer.

No.

This should not affect users. Auth solutions and documentation were created before the aggregator, and so referred to `/model` for the functionality. 

However, now the aggregator, which is required to run the UI, handles the majority of auth-protected endpoints, as well as new Teams functionality. Therefore, in order to prevent incorrect behavior, we should route specifically the auth endpoints to the aggregator.

There should be no risk, as all values/configmaps/secrets should be mounted on both aggregator and cost-analyzer, as both have been functioning as expected already.

## How was this PR tested?
Tested in an existing auth setup with `/model` specified; observed routing of handing to the aggregator.

https://github.com/kubecost/docs/pull/1169


thomasvn:
Why do we need to route `/model/saml` to Aggregator container, if `/saml` already routes there? Same for oidc, login, logout, auth. I think these requests only come from two places:

* The frontend. For this, we should perform all fixes in frontend instead of nginx routing?
* User-defined configuration in Helm or in Identity Provider. I can understand wanting to be robust against any slight misconfigurations. Are we sure this is the best way to do it?

kaelanspatel:
@thomasvn this isn't about "misconfig" more that the intended behavior. This does relate to RBAC Teams; as we cannot be sure the filesystem will be shared between waterfowl vs the cost-analyzer, so we must make sure login (which refers to the teams filestore) is handled by the aggregator. However, it is also a general concern.

When debugging OIDC, for example, I have consistently seen the cost-analyzer handle all logins. This is entirely needless; the full Kubecost application can't be served without the aggregator, and is only because our pre-aggregator documentation on auth specifies config of `/model/oidc/authorize`.

nealormsbee:
Is my understanding here correct:

- Users have existing configs that specify, e.g., `/model/saml`
- We don't want to maintain two code paths at `/model/saml` AND `/saml`
- We also don't want to just break the existing setups of users with the `/model/saml` config, so we're routing `/model/saml` to the same aggregator endpoint as `/saml`

And the idea is that we'd be noisy about `/model/saml` being deprecated, and we remove it in the future? Or am I totally off base here. cc @kaelanspatel @thomasvn

kaelanspatel:
Correct, except for the deprecation. I don't see the need to specifically call out a removal; there is nothing "wrong" with the specific endpoint `/model/saml`. That is what it was previously before the aggregator, and we can still continue to serve it today to (as you say) prevent anyone's config from breaking. 

The main point of this change is that there is no need to route any login functionality (that is, `saml/acs`, `oidc/authorize`, `/login`, `/logout`) through the cost-model. It causes issues when debugging, potentially breaks the new Teams functionality, and will eventually be removed. Because `/model/*` automatically goes to the cost-model unless otherwise stated e.g. for `/model/allocations`), the current nginx config was erroneously routing auth handling over to the cost-model. This simply changes it to hit the aggregator instead.

thomasvn:
Thanks for the explanation, I'm good to approve! Could you add a comment in the nginx config above all `/model/*` stating something along the lines of

```
# This location is used to remain compatible with older configurations of Kubecost SSO. All SSO logic should now be handled by the Aggregator container.
```

I agree that at some point it would be good to deprecate the old `/model/*` SSO configs just to keep the config file from becoming too polluted. We don't have to do that now though, perhaps when there is a major version change.

nealormsbee:
Added the comments @thomasvn.
If it looks good can you smash the ✅ button and this'll merge once the checks finish?

thomasvn (approved):
👊

PR #3817, opened 2025-01-22
Add options to set additional nginx headers

nealormsbee:
## What does this PR change?
Adds options to values.yaml and the nginx template for injecting extra `add_header` directives at the server and location levels. I haven't gone and stubbed this out for every location directive yet - I'd like directional feedback on the concept.

The primary motivation here is that security tools (correctly) flag that `Access-Control-Allow-Origin '*'` is, in many cases, overly-permissive, and that the absence of a `Content-Security-Policy` header opens attack vectors for XSS and clickjacking.

The default values supplied in `values.yaml` with this commit keep the CORS behavior consistent, by defaulting to the permissive approach. Users will now be able to override this permissive configuration to harden their application. The default values for CSP disallow iframe embedding to prevent clickjacking, and forbid pulling in sources (scripts, images, etc) from domains other than self, and Userway (an accessibility application leveraged by some Kubecost users). Users who embed Kubecost in an iframe in their own context would need to update their Helm charts to allow this behavior again.

No.

Users who currently embed Kubecost in an iframe will need to update their `values.yaml` files. Users who want to add extra headers on a per-location basis will now be able to do so via Helm.

One risk is that the CSP could inadvertently block externally-loaded content that I missed in testing.

## How was this PR tested?
It hasn't been (yet), hence draft mode. My test plan will be to deploy to an environment using this chart and explore the application with the network tab open, and look for any content requests blocked by the CSP or CORS errors.

Not yet, but I believe this will require an update.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -539,11 +539,12 @@ kubecostFrontend:
 #    fqdn: cluster-controller.kubecost.svc.cluster.local:9731
 #
 
-  # Configurable, optional headers for nginx responses.
+  # Configurable headers for nginx responses.
   nginxHeaders:
     # applied to all route locations
     server:
       - Content-Security-Policy "default-src 'self' cdn.userway.org; frame-ancestors 'none';";
+      - Cache-Control "must-revalidate";
     # per-location headers
     location:
       root: []
@@ -2491,4 +2492,3 @@ extraObjects: []
 # -- Optional override for the image used for the basic health test container
 # basicHealth:
 #   fullImageName: alpine/k8s:1.26.9
-
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -543,34 +543,10 @@ kubecostFrontend:
   nginxHeaders:
     # applied to all route locations
     server:
+      - "'Access-Control-Allow-Origin' '*' always;"
+      - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
       - Content-Security-Policy "default-src 'self' 'unsafe-inline' api.userway.org cdn.userway.org api-js.mixpanel.com keyper.kubecost.com; frame-ancestors 'none'; font-src 'self' data:;";
       - Cache-Control "must-revalidate";
-    # per-location headers
-    location:
-      root: []
-      healthz: []
-      model: []
-      clusterTurndown:
-        - "'Access-Control-Allow-Origin' '*' always;"
-        - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
-      hideOrphanedResources:
-        - "'Access-Control-Allow-Origin' '*' always;"
-        - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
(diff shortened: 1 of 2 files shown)


cost-analyzer/values.yaml
```diff
@@ -536,6 +536,40 @@ kubecostFrontend:
 #    fqdn: kubecost-multi-diag.kubecost.svc.cluster.local:9007
 #  clusterController:
 #    fqdn: cluster-controller.kubecost.svc.cluster.local:9731
+#
+
+  # Configurable headers for nginx responses.
+  nginxHeaders:
+    # applied to all route locations
+    server:
+      - Content-Security-Policy "default-src 'self' cdn.userway.org; frame-ancestors 'none';";
```
thomasvn: This looks like it'd be a new header. I'm not super familiar with CSPs, do we know this will work?

cost-analyzer/values.yaml
```diff
@@ -536,6 +536,40 @@ kubecostFrontend:
...
+        - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
+      multiClusterDiagnostics:
+        - "'Access-Control-Allow-Origin' '*' always;"
+        - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
+      forecastingEnabled:
+        - "'Access-Control-Allow-Origin' '*' always;"
+        - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
+      productConfigs:
+        - "'Access-Control-Allow-Origin' '*' always;"
+        - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
```
thomasvn: I'm worried that one configuration per route could quickly become unwieldy. Would it be possible start with:

```yaml
nginxHeaders:
  routes:
    - "'Access-Control-Allow-Origin' '*' always;"
    - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
```

These headers would then be added to all routes. Do you see an issue with this from a frontend or security perspective?

cost-analyzer/values.yaml
nealormsbee: No, I still need to test it 😅 

The idea is that sources like scripts, images, styles, etc can only be loaded from the same domain the UI is being accessed through, and cdn.userway.org. And also that the application cannot be embedded in an iframe. MDN has great CSP documentation if you feel like digging in more to check my understanding

cost-analyzer/values.yaml
nealormsbee: I don't think that's an issue. I was attempting to keep the same behavior we have today, where some `location` directives don't support CORS...personally I have no problem with broadly applying it! I can update.

cost-analyzer/values.yaml
nealormsbee: Just pushed an update that I have tested, and it actually does work now.

I had to include directives for `font-src data:;` and I added `unsafe-inline` to `default-src`. I'm not sure whether these will still be flagged. It's safer than having no CSP, and we have something, somewhere, that is adding inline styles to the document.

Sean-Holcomb (approved):
Makes sense to me

PR #3821, opened 2025-01-24
Update values.yaml to define teams configuration, as opposed to commented out

thomasvn:
## What does this PR change?

I believe that the better Helm UX, is to define all the possible configurations the user could possibly make, as opposed to leaving them commented out. I believe we should only be commenting out _examples_ of configurations.

I've updated the Teams configuration to instantiate `.Values.teams.teamsConfigMapName` and `.Values.teams.teamsConfig` as empty values. This should have the same exact effect of when it was previously commented out. Tests shown below.

N/A

N/A

N/A

N/A

## How was this PR tested?

```sh
helm template ./cost-analyzer > kubecost1.yaml

git checkout develop
helm template ./cost-analyzer > kubecost2.yaml

diff kubecost1.yaml kubecost2.yaml
```

The diff shows there is no difference between the two files, with the exception of [this line](https://github.com/kubecost/cost-analyzer-helm-chart/blob/0feac4119142c25ced1d68be76fad9656087c08a/kubecost.yaml#L23306-L23307) which is a base64 encoded string of all the helm values.

N/A

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -423,43 +423,44 @@ oidc:
     #     claimValues:
     #       - "editor"
 
-# teams:
-  # teamsConfigMapName: kubecost-rbac-teams-config # Name of the ConfigMap containing the teams configuration, if manually created, which overrides any other configured teams
-  # teamsConfig: # Values-defined teams configuration, if teamsConfigMapName is not set, which will override UI-configured teams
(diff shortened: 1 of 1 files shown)


thomasvn:
Requesting review from @chipzoller as well for Helm UX opinion

nealormsbee (approved):

PR #3837, opened 2025-02-04
Do not put the Kubecost logo behind auth

nealormsbee:
## What does this PR change?
Prevents access to a logo used by the "auth failed" error page from becoming inaccessible due to...auth failing.

This is a wildly-overly-specific nginx location directive, but when I tried to use `/images/` or `/images`, the image was still blocked by invalid auth. Maybe `^/images/*$`, or something, works. I'd need to read more about nginx. I think we'll want to revise the way our route handling is set up anyway.

No

Prevent access to a logo used by the "auth failed" error page from becoming inaccessible due to...auth failing.

No real risk - I tested by seeing that the logo failed to load before the change, and afterward it loads fine.

## How was this PR tested?
I tested by seeing that the logo failed to load before the change, and afterward it loads fine.

No.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -186,6 +186,7 @@ data:
             return 200 "healthy\n";
         }
         location = /teamsError.html { }
+        location = /images/kubecost-logo.svg { }
 
 {{- else }}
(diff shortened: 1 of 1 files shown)


kaelanspatel (approved):
There is probably an elegant way to handle this, but I think this is fine rn.

thomasvn:
Agree, this is good for now!

PR #3845, opened 2025-02-06
Remove thread limit on read database

biancaburtoiu:
## What does this PR change?
This PR removes the default thread limit (1) on the read database.

No

The number of threads available on the read database can go up to the number of cores on the resource unless it is explicitly set.

The default limit of 1 caused https://kubecost.atlassian.net/browse/ENG-3277

N/A

## How was this PR tested?
We have manually validated that setting the value to zero effectively disables setting the number of threads on the read db.

Public docs PR: https://github.com/kubecost/docs/pull/1176 modifying:
- https://docs.kubecost.com/install-and-configure/install/multi-cluster/federated-etl/aggregator
- https://docs.kubecost.com/install-and-configure/advanced-configuration/resource-consumption#aggregator

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1627,7 +1627,7 @@ kubecostAggregator:
   numDBCopyPartitions: 25
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If no value is set, the number of threads is bound by the
+  # UI queries). If value is 0, the number of threads is bound by the
   # number of cores
   # default: 0 is no limit
   dbReadThreads: 0
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1627,7 +1627,7 @@ kubecostAggregator:
   numDBCopyPartitions: 25
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If value is 0, the number of threads is bound by the
+  # UI queries). If value is 0, the number of threads is bounded by the
   # number of cores
   # default: 0 is no limit
   dbReadThreads: 0
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1627,10 +1627,10 @@ kubecostAggregator:
   numDBCopyPartitions: 25
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If increasing this value, it is recommended to increase the
-  # aggregator's memory requests & limits.
-  # default: 1
-  dbReadThreads: 1
+  # UI queries). If no value is set, the number of threads is bound by the
+  # number of cores
+  # default: 0 is no limit
+  dbReadThreads: 0
 
   # How many threads the write database is configured with (i.e. ingestion of
   # new data from S3). If increasing this value, it is recommended to increase


cost-analyzer/values.yaml
```diff
@@ -1627,10 +1627,10 @@ kubecostAggregator:
   numDBCopyPartitions: 25
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If increasing this value, it is recommended to increase the
-  # aggregator's memory requests & limits.
-  # default: 1
-  dbReadThreads: 1
+  # UI queries). If no value is set, the number of threads is bound by the
```
ameijer:
```suggestion
  # UI queries). If value is 0, the number of threads is bound by the
```

ameijer (approved):

thomasvn (approved):
This should be a safe change for v2.6? We should definitely call this out in the release notes.

@biancaburtoiu If users had a default of 1, what kind of changes in behavior should they expect? Higher but more efficient usage of the node's CPU?

cost-analyzer/values.yaml
```diff
@@ -1627,10 +1627,10 @@ kubecostAggregator:
   numDBCopyPartitions: 25
 
   # How many threads the read database is configured with (i.e. Kubecost API /
-  # UI queries). If increasing this value, it is recommended to increase the
-  # aggregator's memory requests & limits.
-  # default: 1
-  dbReadThreads: 1
+  # UI queries). If value is 0, the number of threads is bound by the
```
biancaburtoiu:
```suggestion
  # UI queries). If value is 0, the number of threads is bounded by the
```

biancaburtoiu:
> This should be a safe change for v2.6? We should definitely call this out in the release notes.
> 
> @biancaburtoiu If users had a default of 1, what kind of changes in behavior should they expect? Higher but more efficient usage of the node's CPU?

Thanks for calling this out @thomasvn -- That's right, this change would allow the aggregator to use more CPU resources.

chipzoller:
Labeled for 2.6

thomasvn:
@chipzoller Since v2.6 is already released, do you want this cherry-picked into a patch release? Or does this go into v2.7?

chipzoller:
I think either may be fine here. Do we have enough for a 2.6 patch?

thomasvn:
Actually after second thought. Because this changes default behavior of the aggregator, I'd recommend v2.7

chipzoller:
That'd be safer provided no one is dying to get this. Do you know?

thomasvn:
@chipzoller Original change was made for a hosted user! No other users have asked.

biancaburtoiu:
@thomasvn @chipzoller I will relabel this to `v2.7` as per your discussion above. Could you please offer some guidance on how the corresponding docs PR https://github.com/kubecost/docs/pull/1176 should be handled in this case, given the upcoming docs freeze/migration?

chipzoller:
> @thomasvn @chipzoller I will relabel this to `v2.7` as per your discussion above. Could you please offer some guidance on how the corresponding docs PR [kubecost/docs#1176](https://github.com/kubecost/docs/pull/1176) should be handled in this case, given the upcoming docs freeze/migration?

The docs modifications will have to happen at a later time and in IBM's system. We can keep it open but just mark it to not be merged so you have a copy of the changes needed.

biancaburtoiu:
Thank you for the clarification @chipzoller. Please see the note I CC-ed you in in the docs PR.
I will merge this labeled as `v2.7`.

PR #3864, opened 2025-02-10
fix: ksm metrics were getting duplicated in openshift env using in-cluster prometheus

mittal-ishaan:
## What does this PR change?
Adds default option for disabling emitting ksm metrics when using in-cluster prometheus in an openshift environment.

No

Users installing kubecost in openshift environment using in-cluster prometheus will now not get degraded warning and will see correct number of CPU in openshift overview page.

https://github.com/opencost/opencost-helm-chart/issues/249

No

## How was this PR tested?
Installed kubecost in openshift environment with in-cluster prometheus. earlier was getting degraded warning after setting:
```yaml
kubecostMetrics:
  emitKsmV1Metrics: false
```
the warning was gone and nothing else was affected(We were getting correct etls and cost was showing up in UI).

### DIFF
--- a/cost-analyzer/values-openshift-cluster-prometheus.yaml
+++ b/cost-analyzer/values-openshift-cluster-prometheus.yaml
@@ -24,3 +24,6 @@ serviceMonitor:
 
 prometheusRule:
   enabled: true
+
+kubecostMetrics:
+  emitKsmV1Metrics: false


JesseGoodier (approved):

cost-analyzer/values-openshift-cluster-prometheus.yaml
```diff
@@ -24,3 +24,6 @@ serviceMonitor:
 
 prometheusRule:
   enabled: true
+
+kubecostMetrics:
```
kastl-ars, line 28: I daresay this should read `kubeStateMetrics`, not `kubecostMetrics`...

mittal-ishaan: This should be `kubecostMetrics`
The helm config is different from the one that is present in opencost 
https://github.com/kubecost/cost-analyzer-helm-chart/blob/6c4b54e78a60e5252d666dc5c6a3d8e8c323836c/cost-analyzer/values.yaml#L554-L558

PR #3898, opened 2025-02-24
Add new configuration for CSV custom pricing on primary

nikovacevic:
## What does this PR change?
Allows users to enable and configure Enterprise Custom Pricing.

Requires https://github.com/kubecost/kubecost-cost-model/pull/3149

New enterprise feature: configure the primary cluster to use CSVs to set custom pricing for nodes, volumes, load balancers, and GPUs across all secondary clusters

Closes https://apptio.atlassian.net/browse/KCM-3318

Pricing changes are inherently risky. If something goes wrong with this feature, all downstream monitoring, optimization, and governance features will be affected.

Testing requirements include checking status, pricing, and logs for the following options:
 1. deployment method (singlepod and statefulset) 
 2. feature disabled and enabled
 3. with feature enabled, changing pricing spec, and applying retroactive pricing

## How was this PR tested?
* The main PR (https://github.com/kubecost/kubecost-cost-model/pull/3149) has ample unit test coverage
* We've tested singlepod and statefulset manually, using Kind and a GCP test cluster

> [!WARNING]  
> We need to re-test after getting a new Clickhouse base image

Documentation is in progress


cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -889,6 +889,14 @@ spec:
             - name: KUBE_RBAC_PROXY_ENABLED
               value: {{ (quote .Values.global.prometheus.kubeRBACProxy) }}
             {{- end }}
+            {{- if .Values.csvCustomPricing }}
```
jessegoodier:
```suggestion
            {{- if .Values.enterpriseCustomPricing }}
```

jessegoodier: just thinking that this may literally be csv, but end naming should reflect the purpose for the user?

nikovacevic: Got it, thanks Jesse! Going to change all three instances in a forthcoming commit.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -889,6 +889,14 @@ spec:
             - name: KUBE_RBAC_PROXY_ENABLED
               value: {{ (quote .Values.global.prometheus.kubeRBACProxy) }}
             {{- end }}
+            {{- if .Values.enterpriseCustomPricing }}
```
chipzoller:
```suggestion
            {{- if (.Values.enterpriseCustomPricing).enabled }}
```

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -889,6 +889,14 @@ spec:
...
+            {{- if .Values.enterpriseCustomPricing.enabled }}
```
chipzoller:
```suggestion
```

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -889,6 +889,14 @@ spec:
...
+            - name: CSV_CUSTOM_PRICING_ENABLED
+              value: "true"
+            - name: CSV_CUSTOM_PRICING_LOCATION_URI
+              value: {{ .Values.enterpriseCustomPricing.location.URI }}
+            {{- end }}
+            {{- end }}
```
chipzoller:
```suggestion
```

chipzoller (commented):
Minor change but it simplifies the templating logic. Good pattern to use in Helm.

jessegoodier:
@nikovacevic or team- can we merge this for easier testing of this and dbv3?

Understand if you would prefer to wait.

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1266,6 +1267,14 @@ Begin Kubecost 2.0 templates
...
     {{- end }}
     {{- end }}
+    {{- if (.Values.enterpriseCustomPricing).enabled }}
+    - name: ENTERPRISE_CUSTOM_PRICING_ENABLED
+      value: "true"
+    - name: ENTERPRISE_CUSTOM_PRICING_CSV_LOCATION_URI
+      value: {{ .Values.enterpriseCustomPricing.location.URI }}
+    - name: ENTERPRISE_CUSTOM_PRICING_APPLY_RETROACTIVELY
+      value: {{ (quote .Values.enterpriseCustomPricing.applyRetroactively) | default (quote false) }}
+    {{- end }}
```
nikovacevic: Good catch. Does this work for both the deployment and the statefulset?

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1049,10 +1049,11 @@ Begin Kubecost 2.0 templates
       mountPath: "/etc/pki/ca-trust/extracted"
       readOnly: false
     {{- end }}
-    {{- /* Only adds extraVolumeMounts if aggregator is running as its own pod */}}
-    {{- if and .Values.kubecostAggregator.extraVolumeMounts (eq (include "aggregator.deployMethod" .) "statefulset") }}
+    {{- /* TODO: CONFIRM THAT THIS IS OKAY!!! */}}
```
nikovacevic: Did we get an answer to this "to do" yet?

nikovacevic:
> @nikovacevic or team- can we merge this for easier testing of this and dbv3?
> 
> Understand if you would prefer to wait.

Looks to me like we have some outstanding TODOs that would be better left open until we're certain they're safe. @nik-kc care to comment and confirm, or correct me?

chipzoller (commented):
I know this is draft, but while it's great to have a separate values file all the same values (~including ones that still haven't been reflected in this custom values file~) need to get added into the main values file before we can merge.

cost-analyzer/templates/_helpers.tpl
chipzoller: Haven't tried StatefulSet yet.

cost-analyzer/templates/_helpers.tpl
chipzoller: Have not tried SS.

nik-kc:
> > @nikovacevic or team- can we merge this for easier testing of this and dbv3?
> > Understand if you would prefer to wait.
> 
> Looks to me like we have some outstanding TODOs that would be better left open until we're certain they're safe. @nik-kc care to comment and confirm, or correct me?

Yes, I can confirm that they would be better left open until we're certain they're safe.

cost-analyzer/values.yaml
```diff
@@ -360,6 +360,24 @@ pricingCsv:
...
     csvAccessCredentials: pricing-schema-access-secret
 
+# Enterprise custom pricing, enterprise key required
+enterpriseCustomPricing:
+  # Enabled the feature by setting this flag to "true". If enabled, the feature
+  # requires a CSV pricing spec to be provided at the location, below, using a
+  # ConfigMap, for which we provide instructions, also below.
+  enabled: false
+  # location:
+  #  URI: /var/configs/enterprise-pricing/spec.csv  # local ConfigMap
```
thomasvn: As a best practice, avoid commenting out configurations. It's preferred to explicitly define them, even if they're disabled. In the example below, setting `enabled: false` ensures the related configs won't be templated:

```suggestion
  location:
    URI: /var/configs/enterprise-pricing/spec.csv  # local ConfigMap
```

Here are additional style guidelines I wrote up. Let me know if you disagree though! https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/DEVELOPMENT.md#yaml-style-guidelines

cost-analyzer/values.yaml
```diff
@@ -360,6 +360,24 @@ pricingCsv:
...
+  # Use the following command to create a ConfigMap from your pricing spec CSV.
+  # You may change the ConfigMap name, or file name, as long as you set the
+  # correct configMapName and location.URI, respectively.
+  #   kubectl create configmap -n kubecost kubecost-enterprise-pricing --from-file spec.csv
+  # configMapName: kubecost-enterprise-pricing
```
thomasvn: Same applies to this config! Please do not comment the config.

cost-analyzer/values.yaml
```diff
@@ -360,6 +360,24 @@ pricingCsv:
...
+  # By default, enterprise custom pricing takes effect moving forward from when
+  # it is enabled. To price all historic data according to the given pricing
+  # spec, set "applyRetroactively" to true, below.
+  # applyRetroactively: true
```
thomasvn: Same applies to this config! Please do not comment the config.

cost-analyzer/templates/_helpers.tpl
```diff
@@ -969,6 +969,16 @@ Begin Kubecost 2.0 templates
...
+      # mount the clickhouse directories on the same PV as the duckdb, 
+      # this way they can seamlessly share the same PV before, during, and after the upgrade
+    - name: aggregator-db-storage
+      mountPath: /var/lib/clickhouse
+      {{- end }}
+    {{- end }}
+    {{- if and .Values.kubecostAggregator.useDBv3 (eq (include "aggregator.deployMethod" .) "singlepod") }}
+    - name: persistent-configs
+      mountPath: /var/lib/clickhouse
     {{- end }}
```
thomasvn, line 982: @ameijer Did you make this change? Did this work when you tested it?

I ask because based on the current logic, if `useDBv3=true`, then there will be duplicate volumeMounts for "aggregator-db-storage" when Agg is StatefulSet and "persistent-configs" when Agg is SinglePod

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1256,6 +1270,14 @@ Begin Kubecost 2.0 templates
     {{- end}}
     {{- end }}
     {{- end }}
+    {{- if (.Values.enterpriseCustomPricing).enabled }}
+    - name: ENTERPRISE_CUSTOM_PRICING_ENABLED
+      value: "true"
+    - name: ENTERPRISE_CUSTOM_PRICING_CSV_LOCATION_URI
+      value: {{ .Values.enterpriseCustomPricing.location.URI }}
+    - name: ENTERPRISE_CUSTOM_PRICING_APPLY_RETROACTIVELY
+      value: {{ (quote .Values.enterpriseCustomPricing.applyRetroactively) | default (quote false) }}
```
thomasvn: Minor comment. My preference is to always set defaults in the `values.yaml` as opposed to in the template here. That way it's more transparent to the user what the default behavior should be.

nikovacevic: Makes sense. I'm happy to follow whatever the best practices are. I just need to be re-trained after years of helm slop!

nikovacevic:
@thomasvn removed defaults and uncommented values! Thanks for your feedback!

cost-analyzer/templates/_helpers.tpl
ameijer: I did indeed make this change, and I did test it.  The duplicate volume mounts is deliberate. Explanation below: 
 
Currently, duckdb writes to the persistent-configs mountpoint at /var/configs. While we can debate whether this was the right thing in the first place, that's the way it is. I added a second mount point here so that clickhouse will write its data to the same PVC as duck did. A similar thing happens in statefulset mode, people expect the db data to get written to aggregator-db-storage PVC - this lets clickhouse and duck (who are writing to different points in the file system) write to the same PVC. Our other option would be to create a second PVC for clickhouse only in both single pod and statefulset cases, but I decided against this because that's added complexity to the chart and an extra PVC adds a lot of weight... and anyway we can maybe go back to one mount point when duck goes away

thomasvn: Ok, makes sense that we want DuckDB and ClickHouse to share the same PVC, but write at different mountPaths. And thanks for testing!

Also, `/var/lib` seems like the right place for the mountPath, given that this is the default for ClickHouse installs ([ref](https://clickhouse.com/docs/operations/server-configuration-parameters/settings#path))

ameijer: if we mount at /var/lib, we would clobber the other stuff in there - I think we are forced into a corner with a mount at /var/lib/clickhouse

thomasvn: Oops sorry, I meant `/var/lib/clickhouse`! I agree with the changes we've already got here

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1256,6 +1270,14 @@ Begin Kubecost 2.0 templates
     {{- end}}
     {{- end }}
     {{- end }}
+    {{- if (.Values.enterpriseCustomPricing).enabled }}
+    - name: ENTERPRISE_CUSTOM_PRICING_ENABLED
+      value: "true"
+    - name: ENTERPRISE_CUSTOM_PRICING_CSV_LOCATION_URI
+      value: {{ .Values.enterpriseCustomPricing.location.URI }}
+    - name: ENTERPRISE_CUSTOM_PRICING_APPLY_RETROACTIVELY
+      value: (quote {{ .Values.enterpriseCustomPricing.applyRetroactively }})
```
thomasvn: Any reason why one env var adds quotes and the other doesn't?

thomasvn (approved):
Nice work!

cost-analyzer/templates/_helpers.tpl
nikovacevic: I think it's a remnant of the default... I'll remove it

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1256,6 +1270,14 @@ Begin Kubecost 2.0 templates
     {{- end}}
     {{- end }}
     {{- end }}
+    {{- if (.Values.enterpriseCustomPricing).enabled }}
+    - name: ENTERPRISE_CUSTOM_PRICING_ENABLED
+      value: "true"
+    - name: ENTERPRISE_CUSTOM_PRICING_CSV_LOCATION_URI
+      value: {{ .Values.enterpriseCustomPricing.location.URI }}
+    - name: ENTERPRISE_CUSTOM_PRICING_APPLY_RETROACTIVELY
+      value: {{ .Values.enterpriseCustomPricing.applyRetroactively }}
```
jessegoodier: see my new PR. must be quoted.

JesseGoodier (approved):

PR #3910, opened 2025-02-27
Teams config example

JesseGoodier:
## What does this PR change?
Adding a sample teams config to the helm chart as the yaml is fairly long for our docs.

### DIFF
--- a/.github/workflows/link-checker.yaml
+++ b/.github/workflows/link-checker.yaml
@@ -23,4 +23,4 @@ jobs:
         with:
           fail: true
           debug: false
-          args: --verbose --include-fragments --no-progress --base . -E './**/*.md' './**/*.yaml' './**/*.yml' './cost-analyzer/values-teams-example-config.yaml'
+          args: --verbose --include-fragments --no-progress --base . -E './**/*.md' './**/*.yaml' './**/*.yml'
--- a/.github/workflows/link-checker.yaml
+++ b/.github/workflows/link-checker.yaml
@@ -23,4 +23,4 @@ jobs:
         with:
           fail: true
           debug: false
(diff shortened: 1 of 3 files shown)


.github/workflows/link-checker.yaml
```diff
@@ -23,4 +23,4 @@ jobs:
         with:
           fail: true
           debug: false
-          args: --verbose --include-fragments --no-progress --base . -E './**/*.md' './**/*.yaml' './**/*.yml'
+          args: --verbose --include-fragments --no-progress --base . -E './**/*.md' './**/*.yaml' './**/*.yml' './cost-analyzer/values-teams-example-config.yaml'
```
chipzoller: Not clear why this arg modification is necessary or why we'd want to check links in this file at all. I can't confirm on my end that the example config here is correct so may need to defer to @kaelanspatel or @thomasvn for independent verification.

kaelanspatel:
My only concern here is that including `Group` in an example will lead people to believe this is an inherently exported value of a SAML response, which it is not AFAIK. This is why the example provided [in the docs](https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/teams#enable-teams-and-define-in-helm) uses `NameID`.

jessegoodier:
> My only concern here is that including `Group` in an example will lead people to believe this is an inherently exported value of a SAML response, which it is not AFAIK. This is why the example provided [in the docs](https://docs.kubecost.com/using-kubecost/navigating-the-kubecost-ui/teams#enable-teams-and-define-in-helm) uses `NameID`.

thanks, great feedback! I added some stronger comments to the top. Does that help?

kaelanspatel (approved):

PR #3924, opened 2025-03-07
CP 2.5 network-costs and modeling image bumps

JesseGoodier:
Do not merge 
## What does this PR change?
bumps network cost and modeling images to fix CVEs
fixing merge conflicts


jessegoodier:
this would require #3934 
holding off until we have a specific request.

PR #3926, opened 2025-03-10
fix openshift helm values example to have aggregator metrics service monitor

mittal-ishaan:
## What does this PR change?
Update the in cluster prometheus openshift helm values to also add a service monitor for aggregator metrics 

No

Users will be able to get aggregator metrics in openshift prometheus.

## How was this PR tested?
Installed kubecost on an openshift cluster to confirm if aggregator metrics are coming which were missing earlier


jessegoodier:
We are doing some more testing  before we merge this.

cost-analyzer/values-openshift-cluster-prometheus.yaml
```diff
@@ -21,6 +21,8 @@ global:
 
 serviceMonitor:
   enabled: true
+  aggregatorMetrics:
+    enabled: true
```
jessegoodier:
```suggestion
    enabled: false
```

jessegoodier:
good to have the value defined, but no need to enabled by default

JesseGoodier (approved):

PR #3962, opened 2025-03-20
add helm flag to allow opt-in to new DB

ameijer:
## What does this PR change?
Adds a helm flag to enable opting in to using clickhouse. this will be needed pre-3.0 to run clickhouse so users can use clickhouse features like CSV pricing.

@chipzoller , would love your input as to the ergonomics of this one esp since you are interested in playing around with CSV pricing. This switch will need to be flipped to get CSV pricing working

no

adds a flag that users can optionally leverage in their helm config

no risks, is opt-in

## How was this PR tested?
tested via helm template, manually verified rendering was correct

There are comments in values.yaml. I think we deliberately don't want to document this, since it will only be enabled for customers working with Kubecost directly. 

cc @nikovacevic @nik-kc

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1672,7 +1672,7 @@ kubecostAggregator:
 
   # Enable the use of the v3 database (ClickHouse) as the database for the aggregator.
   # the existing database will be migrated if it exists
-  # This is required to be 'true' to use the CSV pricing feature pre kubecost 3.0.0
+  # This is required to be 'true' to use the enterpriseCustomPricing feature prior to kubecost 3.0.0
   # this will not be optional starting in kubecost 3.0.0 and will be removed
   # default: false
   useDBv3: false
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1670,12 +1670,12 @@ kubecostAggregator:
   dbMemoryLimit: 0GB
   dbWriteMemoryLimit: 0GB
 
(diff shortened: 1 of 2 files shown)


chipzoller (commented):
The change is simple and the position within the values file is logical. I'm wondering if we should use a different name for the value though. Whatever we set now we'll probably want to carry into 3.0. I'm wondering if there's a better way to communicate this (or dull it) to users. If we considered DuckDB "Database 2.0" would we consider this as something more generic like `useDBv3`? At the end of the day, these are implementation details users won't care anything about, so I'm wondering what's the best way to present it.

ameijer:
great, I like it Chip, let me make that change

ameijer:
@chipzoller I made the update, WDYT?

chipzoller:
Looks fine to me. Any feedback, @jessegoodier and @thomas? This seems like a good approach and name?

cost-analyzer/values.yaml
```diff
@@ -1670,6 +1670,13 @@ kubecostAggregator:
   dbMemoryLimit: 0GB
   dbWriteMemoryLimit: 0GB
 
+  # Enable the use of the v3 database (ClickHouse) as the database for the aggregator.
+  # the existing database will be migrated if it exists
+  # This is required to be 'true' to use the CSV pricing feature pre kubecost 3.0.0
```
```suggestion
  # This is required to be 'true' to use the enterpriseCustomPricing feature prior to kubecost 3.0.0
```

JesseGoodier (approved):

thomasvn:
Looks great! Thanks for the comments in `values.yaml`.

PR #3995, opened 2025-04-02
Fix duplicate ksm metrics openshift

mittal-ishaan:
## Release Notes Headline
Fixes duplicate KSM metrics in openshift when using in cluster prometheus causing API server degraded warning and double no. of CPU cores in the dashboard.

## Commentary for Reviewers
We are disabling metrics that are already exposed by openshift prometheus to avoid duplicates. These metrics were checked manually to be present in the openshift cluster prometheus comparing to the ones kubecost emits.

## Links to Issues or Tickets
Opencost Issue: https://github.com/opencost/opencost-helm-chart/issues/249
But has the same affect on kubecost too

## Testing
Kubecost deployment on openshift env.

## Documentation Updates
NA


JesseGoodier (approved):
Thank you!

PR #3998, opened 2025-04-09
WIP

cpetersen5:


thomasvn:
@cpetersen5 @ameijer Let me know if you need any assistance with the NGINX config! It looks like the main changes in this config file, are to remove all routing towards the `kubecost-cost-analyzer.kubecost.svc.cluster.local:9003` service.

PR #4011, opened 2025-04-24
[KCM-170] Prevent always deploying a kubecost replica when kubecostDeployment.replicas is explicitly set to 0

nealormsbee:
## Release Notes Headline
Prevent always deploying a Kubecost replica when `kubecostDeployment.replicas` is explicitly set to 0

## Commentary for Reviewers

Due to Helm treating `0` as "falsey", the idiom `{{ .Values.myValue | default 1 }}` does _not_ behave as the docs claim, where the `default` is used in the case of an _omitted_ value. Explicitly setting `.Values.myValue` to `0` also triggers the default. As a result, it was impossible to actually set the number of deployment replicas to `0`.

Why someone would want to deploy 0 of something is beyond me, but we do have an old customer ticket for it.

This change gets around the odd Helm behavior by casting the input to a string, then piping through the `default` function, and then casting the result back to an `int`.

## Links to Issues or Tickets
- Closes https://apptio.atlassian.net/browse/KCM-170

## User Impact and Risks
I someone is relying on setting 0 replicas to get a single replica, this will break them by causing them to actually deploy 0 replicas.

## Testing
I tested by deploying Kubecost, both with no custom values (resulted in 1 replica) and with a custom value for 0 replicas. The result of the 0-replica deployment is shown here:
image

## Documentation Updates

I don't believe this should require a docs update.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -21,7 +21,7 @@ metadata:
   {{- end }}
 spec:
 {{- if .Values.kubecostDeployment }}
-  replicas: {{ toString .Values.kubecostDeployment.replicas | default 1 | int }}
+  replicas: {{ .Values.kubecostDeployment.replicas }}
 {{- end }}
   selector:
     matchLabels:
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -21,7 +21,7 @@ metadata:
   {{- end }}
 spec:
 {{- if .Values.kubecostDeployment }}
-  replicas: {{ .Values.kubecostDeployment.replicas | default 1 }}
+  replicas: {{ toString .Values.kubecostDeployment.replicas | default 1 | int }}
 {{- end }}
   selector:
     matchLabels:


JesseGoodier (approved):

thomasvn:
Nice work @nealormsbee! Thanks for sharing the test results as well.

thomasvn (approved):

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
   {{- end }}
 spec:
 {{- if .Values.kubecostDeployment }}
-  replicas: {{ .Values.kubecostDeployment.replicas | default 1 }}
+  replicas: {{ toString .Values.kubecostDeployment.replicas | default 1 | int }}
```
thomasvn:
```suggestion
  replicas: {{ toString .Values.kubecostDeployment.replicas | int }}
```

One minor change if you're open to it. From a UX perspective, we try to have all defaults defined in the `values.yaml` as opposed to hidden in the templates ([ref](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/DEVELOPMENT.md#yaml-style-guidelines)). Let me know what you think!

nealormsbee: That works for me!

If that's the case....could we just drop both `toString` and `int`, and have `{{ .Values.kubecostDeployment.replicas }}` ?

thomasvn: Hm, I _think_ we would still need the `toString` and `int`.

```yaml
# values.yaml
kubecostDeployment:
  replicas: 1  # default value of 1, but what if somebody sets this to 0?
```

nealormsbee: If somebody sets `0` in their `values.yaml`, but we aren't passing the value through the `default` function, then it will just get translated directly into the template, right?

Like, we stringify the value so that it isn't "falsey" when it runs through the `default` function, and then we go back to an int. But, if there's no `default` function to worry about, then I don't think we need to do the `int -> string -> int` dance.

I'll just try it out and post results - probably better than any amount of guessing :)

nealormsbee: Tried it out! It works. If you want to verify, you can pull down the updated branch and with values.yaml of

```
kubecostDeployment:
  replicas: 0
```
Run the install and see that we get no cost-analyzer pods!

thomasvn: Wooh thanks for testing!

PR #4015, opened 2025-04-24
[KCM-3988] Make CSP more permissive by default

nealormsbee:
## Release Notes Headline

Make the default Content-Security-Policy for Kubecost more permissive so as not to break compatibility with old default behaviors. Note that this default errs on the side of flexibility over security. Users running in production are encouraged to harden this setting.

## Commentary for Reviewers

A while back, I introduced the notion of a Content Security Policy to Kubecost's nginx config. This is set via the Helm chart.
I tried to make the default behavior as restrictive/secure as possible without breaking any longstanding default behavior. This mostly succeeded, however we have found that the "Switch Context" feature, which allows a UI to point at an arbitrary Kubecost Primary API breaks, because the `connect-src` directive is not set, and the `default-src` directive does not allow fetch requests to target arbitrary hosts. Another issue is that some images which are embedded as `data://` URLs fail to render, because only fonts were allowed to use `data://` schemes.

This change updates the CSP to allow requests on the `https:`, `http:`, `ws:`, and `data:` schemes by default for all request types. We still set `frame-ancestors` to `'none'` to prevent clickjacking attacks.

## Links to Issues or Tickets

Closes https://apptio.atlassian.net/browse/KCM-3988

## User Impact and Risks

The primary impact is to restore the function of the "Switch Context" feature. This also slightly weakens security of the default CSP, because it is now more permissive. In a future release, in which breaking backwards compatibility is acceptable, I would like to make the CSP secure by default.

## Testing

I installed with the changes and performed cross-origin operations.

## Documentation Updates

I do not believe any are required.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -555,7 +555,7 @@ kubecostFrontend:
     server:
       - "'Access-Control-Allow-Origin' '*' always;"
       - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
-      - Content-Security-Policy "default-src 'self' 'unsafe-inline' api.userway.org cdn.userway.org api-js.mixpanel.com keyper.kubecost.com; frame-ancestors 'none'; font-src 'self' data:;";
+      - "Content-Security-Policy \"default-src 'self' 'unsafe-inline' http: https: ws: data:; frame-ancestors 'none';\";"
       - Cache-Control "must-revalidate";
 
 # Kubecost Metrics deploys a separate pod which will emit kubernetes specific metrics required


cost-analyzer/values.yaml
```diff
@@ -555,7 +555,7 @@ kubecostFrontend:
     server:
       - "'Access-Control-Allow-Origin' '*' always;"
       - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
-      - Content-Security-Policy "default-src 'self' 'unsafe-inline' api.userway.org cdn.userway.org api-js.mixpanel.com keyper.kubecost.com; frame-ancestors 'none'; font-src 'self' data:;";
+      - "Content-Security-Policy \"default-src 'self' 'unsafe-inline' http: https: ws: data:; frame-ancestors 'none';\";"
```
thomasvn, line 558: Looks good! Just double checking it looks like you've dropped `font-src` in the new line? That was intentional?

thomasvn (approved):

cost-analyzer/values.yaml
nealormsbee: That was intentional! I will explain.

You can limit the places that different kinds of content is allowed to originate from, and then fall back to `default-src` in the absence of a rule for a specific type of content.

Previously, the rule for `default-src` was to only allow inline content, content from `'self'` (the domain where the UI is hosted), and a specific subset of known domains (userway, mixpanel). We had some fonts that were embedded as `data://` urls, so I set a specific rule for those.

_Now_, I'm trying to derisk breaking the application by making the overall policy more permissive. One issue that's already occurred is that we couldn't use the context-switcher because we'd need to allow connecting to any host via `connect-src`.  Another is that an image was added to the product using a `data://` url, which we should disallow somehow, but we don't have dev tooling in place for that. So, in development, the image rendered normally. Then, in-product, it fails to render because the CSP does not allow `data://` for images.

The approach I'm taking here is to make the CSP pretty darn permissive so that we don't break the application in other unforeseen ways. We could try to keep it as tightly scoped as possible with e.g.

```
Content-Security-Policy: "font-src: 'self' data:; connect-src: http: https:; img-src: 'self' data:; default-src: ...;"
```

And that would be more secure. My worry is that we'd run into more cases where:

1. This breaks the product in ways I did not foresee (like with context switching)
2. This increases the chance that we break the product in the future, due to not having a good way to prevent e.g. embedding images as `data://` urls.

---

It's a good question. May be worth reconsidering, given that we need to be enhancing our security posture...I'll see if I can rework to be safe for the app without giving up too much ground security-wise.

PR #4051, opened 2025-05-05
remove pvc from test

JesseGoodier:
edit chart test to remove pvc requirement for eks due to current failures.
NA on release notes.
test: https://github.com/kubecost/cost-analyzer-helm-chart/actions/runs/14846038124

issue: 
image

I deployed a replicated cluster and found that there are no SC. which we really don't need for this test anyway.

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -1,4 +1,4 @@
-name: Chart build
+name: Chart build and test
 
 on:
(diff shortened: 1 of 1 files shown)


thomasvn (approved):
Looks good! Could you post a screenshot of the errors you're seeing into this PR? Just so we can go back to reference them if needed?

PR #4060, opened 2025-05-09
Add unified agent access needs to cluster role

Sean-Holcomb:
## Commentary for Reviewers
This change updates the Cluster Role for Cost-Analyzer to support the access needs of the unified agent
These changes were made using these permissions:
```
- verbs:
      - get
      - watch
      - list
    apiGroups:
      - ''
      - extensions
      - apps
      - batch
    resources:
      - namespaces
      - replicationcontrollers
      - services
      - nodes
      - nodes/spec
      - pods
      - jobs
      - cronjobs
      - persistentvolumes
      - persistentvolumeclaims
      - deployments
      - replicasets
      - daemonsets
  - verbs:
      - get
      - list
    apiGroups:
      - ''
    resources:
      - services/proxy
      - pods/proxy
      - nodes/proxy
      - nodes/stats
      - nodes/metrics
```

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
@@ -39,10 +39,7 @@ rules:
     resources:
       - configmaps
       - nodes
-      - nodes/spec
       - pods
-      - jobs
-      - cronjobs
       - events
       - services
       - resourcequotas
@@ -52,46 +49,39 @@ rules:
       - persistentvolumes
       - namespaces
       - endpoints
-      - deployments
-      - replicasets
-      - daemonsets
     verbs:
       - get
       - list
       - watch
   - apiGroups:
       - ''
+    verbs:
+      - get
+      - list
     resources:
       - services/proxy
       - pods/proxy
       - nodes/proxy
-      - nodes/stats
-      - nodes/metrics
+  - apiGroups:
+      - apps
+    resources:
+      - statefulsets
+      - deployments
+      - daemonsets
+      - replicasets
     verbs:
       - get
       - list
+      - watch
   - apiGroups:
-      - extensions
-      - apps
       - batch
     resources:
-      - namespaces
-      - replicationcontrollers
-      - services
-      - nodes
-      - nodes/spec
-      - pods
-      - jobs
       - cronjobs
-      - persistentvolumes
-      - persistentvolumeclaims
-      - deployments
-      - replicasets
-      - daemonsets
+      - jobs
     verbs:
       - get
-      - watch
       - list
+      - watch
   - apiGroups:
       - autoscaling
     resources:
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
@@ -59,8 +59,6 @@ rules:
       - get
       - list
     resources:
-      - services/proxy
-      - pods/proxy
       - nodes/proxy
   - apiGroups:
       - apps
--- a/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
@@ -39,7 +39,10 @@ rules:
     resources:
       - configmaps
       - nodes
+      - nodes/spec
       - pods
+      - jobs
+      - cronjobs
       - events
       - services
       - resourcequotas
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -39,7 +39,10 @@ rules:
     resources:
       - configmaps
       - nodes
+      - nodes/spec
```
chipzoller: This isn't a valid resource to my knowledge. `spec` is not a subresource. `status`, however, is.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -39,7 +39,10 @@ rules:
...
       - pods
+      - jobs
+      - cronjobs
```
chipzoller: These two API resources are in the `batch` API group so this change does nothing.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -49,30 +52,46 @@ rules:
       - persistentvolumes
       - namespaces
       - endpoints
+      - deployments
+      - replicasets
+      - daemonsets
```
chipzoller: These three resources are in the `apps` API group so this also does nothing.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -49,30 +52,46 @@ rules:
...
+      - ''
     resources:
-      - statefulsets
-      - deployments
-      - daemonsets
-      - replicasets
+      - services/proxy
+      - pods/proxy
+      - nodes/proxy
+      - nodes/stats
```
chipzoller: `/stats` is not a subresource to the Node resource. `/status` is.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -49,30 +52,46 @@ rules:
...
+      - nodes/metrics
```
chipzoller: This is also not a valid subresource to my knowledge.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -49,30 +52,46 @@ rules:
...
     verbs:
       - get
       - list
-      - watch
   - apiGroups:
+      - extensions
```
chipzoller: The `extensions` API group is no longer needed.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -49,30 +52,46 @@ rules:
...
+      - apps
       - batch
     resources:
-      - cronjobs
+      - namespaces
+      - replicationcontrollers
```
chipzoller: We don't need to support this antiquated API resource moving forward.

chipzoller (commented):
I know this wasn't exactly in scope for your changes, but we should take this opportunity to make at least this ClusterRole correct per each API resource and not throwing things together hoping they work.

mbolt35:
@chipzoller Good to know about all of these mistakes, this was forwarded to us from the Cloudy team as their cluster role requirements. Since access to the node metrics directly was throwing 403s, I asked them for their resource access and did not verify before sean pushed this up. 

I think that really all we need is `nodes/proxy` added, since that is what we were specifically missing that caused the 403s. 

I will check to see the reasoning behind:
```
- services/proxy
- pods/proxy
```

chipzoller:
I see. In that case it's probably just `node/proxy` especially if the code is getting metrics from nodes through the API server (not going direct). That is the main purpose of that subresource.

mbolt35:
@chipzoller We'll get this trimmed down to just the new subresource early next week - thanks again for the help.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
Sean-Holcomb: Can you expand on this a little bit. Is it no longer needed in general or no longer needed for Kubecost?

Sean-Holcomb:
@chipzoller thanks for the feedback. Just changing this to git and list for the proxy sub resources. If we do end up needing more we we can add them.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
```diff
@@ -53,6 +53,15 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - ''
+    verbs:
+      - get
+      - list
+    resources:
+      - services/proxy
```
chipzoller: Are we certain this is even a valid subresource for the Service API? In my searches I couldn't confirm this. Similar with `/proxy` for Pods.

chipzoller (commented):
I would check and confirm the `/proxy` subresource is valid for Pods and Services. Based on getting metrics from nodes via the API server (the typical need for `node/proxy`), this should be the only thing needed, but maybe there's more to it.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
chipzoller: This was removed as I see it.

cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
Sean-Holcomb: Honestly the documentation is very obtuse on the subresource thing as I am sure you know. We can cut those out too

chipzoller: Came back to say `service/proxy` is valid. But, that said, we should be positive if we need it or not.

chipzoller (approved):
In discussion with Sean, it seems this now works as intended.

PR #4063, opened 2025-05-12
[KCM-4051] Clickhouse by default for 3.0

ameijer:
## Release Notes Headline

switch default to DBV3 (Clickhouse) 

## Commentary for Reviewers

this makes clickhouse an opt-out feature rather than an opt-in

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-4051

## User Impact and Risks

switches the underlying database. impact is high but this is why we are doing it in 3.0

## Testing

Tested in nightly envs


cpetersen5 (approved):
✅ 🎉

PR #4064, opened 2025-05-13
cluster controller custom-secret fix

JesseGoodier:
- Update cluster controller helm settings due to conflicting values. Action required to adjust configs if using the API key from secondary clustesrs
- Add config validation
- Add force pod restart when api key changes

## Release Notes Headline

- Update cluster controller helm settings due to conflicting values. Action required to adjust configs if using the API key from secondary clustesrs
- Add config validation
- Add force pod restart when api key changes

## Commentary for Reviewers

Tested many different configs with the new clusterController.kubecostAPIKeyCheck 

## Links to Issues or Tickets

KCM-3986

## User Impact and Risks

Could have a logic error, but I tested many paths

## Testing

```sh
helm upgrade kubecost-helm-chart ../cost-analyzer --set clusterController.enabled=true --set clusterController.createClusterControllerSecret=false 
```

logs:

```
kl deployments/kubecost-helm-chart-cluster-controller
2025-05-13T15:05:30Z INF Determined to be running in a cluster. Using in-cluster K8s config.
2025-05-13T15:05:30Z INF Recommendation service CheckAvailable() GET finished apiKey=mMH1pTXXXX status=200 url=http://kubecost...
```

Using old method:
image

Bad secret:
image

- works with valid custom existing secret

## Documentation Updates

N/A


cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -222,10 +221,11 @@ spec:
         {{- if .Values.clusterController.labels }}
         {{- toYaml .Values.clusterController.labels | nindent 8 }}
         {{- end }}
-      {{- with .Values.global.podAnnotations}}
       annotations:
+      {{- with .Values.global.podAnnotations}}
         {{- toYaml . | nindent 8 }}
       {{- end }}
+        checksum/configs: {{ include "configsChecksum" . }}
```
thomasvn, line 227: Good catch!

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -294,7 +294,7 @@ spec:
           {{- else }}
           value: http://{{ $serviceName }}.{{ .Release.Namespace }}:{{ .Values.service.targetPort | default 9090 }}/model
           {{- end }}
-        {{- if .Values.clusterController.kubecostAPIKey }}
+        {{- if or .Values.clusterController.kubecostAPIKey (eq .Values.clusterController.createClusterControllerSecret false) }}
```
jessegoodier: clusterController.kubecostAPIKeyCheck will check if the secret exists and has the correct key

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
jessegoodier: Thanks. See updated testing in notes above and added some details

thomasvn:
@jessegoodier One concern that may be out of scope for this PR. For a while now, I think we have been using `.Values.clusterController.secretName` mistakenly for two different purposes.

One is for the k8s Secret in this PR ([ref](https://github.com/kubecost/cost-analyzer-helm-chart/blob/v2.7/cost-analyzer/templates/kubecost-cluster-controller-secret-template.yaml)).

Another is for a k8s Secret we ask the user to create with a `service-key.json` with node autoscaling permissions ([docs ref](https://www.ibm.com/docs/en/kubecost/self-hosted/2.x?topic=configuration-cluster-controller)):

https://github.com/kubecost/cost-analyzer-helm-chart/blob/7d4a7b99367a039a82a9c31c4c41728514c0e08c/cost-analyzer/templates/kubecost-cluster-controller-template.yaml#L336-L344

cost-analyzer/templates/_helpers.tpl
```diff
@@ -157,6 +157,26 @@ will result in failure. Users are asked to select one of the two presently-avail
...
+{{/*
+Cluster Controller API Key check
+*/}}
+{{- define "clusterController.kubecostAPIKeyCheck" -}}
+  {{- if .Values.clusterController.enabled }}
+    {{- if and .Values.clusterController.kubecostAPIKey ((eq .Values.clusterController.createClusterControllerSecret false)) }}
+      {{- fail "\n\nConfiguration error:\nWhen setting clusterController.kubecostAPIKey, clusterController.createClusterControllerSecret must be true." -}}
+    {{- end -}}
+    {{- if and (not .Values.clusterController.createClusterControllerSecret) .Values.clusterController.secretName }}
+      {{- if not (lookup "v1" "Secret" .Release.Namespace .Values.clusterController.secretName) }}
```
thomasvn: Nice idea adding some checks in! If we are using the `lookup` function here, do we need to also first check for `.Values.global.platforms.cicd.enabled=true`?

cost-analyzer/templates/_helpers.tpl
```diff
@@ -157,6 +157,26 @@ will result in failure. Users are asked to select one of the two presently-avail
...
 
+{{/*
+Cluster Controller API Key check
+*/}}
+{{- define "clusterController.kubecostAPIKeyCheck" -}}
+  {{- if .Values.clusterController.enabled }}
+    {{- if and .Values.clusterController.kubecostAPIKey ((eq .Values.clusterController.createClusterControllerSecret false)) }}
+      {{- fail "\n\nConfiguration error:\nWhen setting clusterController.kubecostAPIKey, clusterController.createClusterControllerSecret must be true." -}}
+    {{- end -}}
+    {{- if and (not .Values.clusterController.createClusterControllerSecret) .Values.clusterController.secretName }}
```
thomasvn: One other concern with this check, is that it expects a certain value inside `.Values.clusterController.secretName`. I'm good to approve for now, but should we create a Jira to track/resolve the conflicting usage of `.Values.clusterController.secretName` (details in a different PR comment).

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
thomasvn: Sorry if I'm confused about this. Why do we have the `(eq .Values.clusterController.createClusterControllerSecret false)` check here?

Regardless of whether the secret was created by the user or the Helm chart, don't we need to extract the `kubecostAPIKey` from the secret and set it as an env var?

cost-analyzer/templates/_helpers.tpl
jessegoodier: fixed, moved to the secret reference in the deployment template.

cost-analyzer/templates/_helpers.tpl
jessegoodier: this was a huge logic error. Can't thank you enough for thinking this through.

Given that we have very few users, I'm taking the opportunity to get it right now. Added a fail if using old keys pairs.

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
jessegoodier: fixed!

cost-analyzer/values.yaml
```diff
@@ -1176,10 +1176,22 @@ clusterController:
...
+  usePrimaryApiForRightSizing: false
+  primaryKubecostURL: "" # URL for secondary clusters to connect to primary Kubecost (ex: https://kubecost.myorganization.com)
+  # There are two methods available to configure the API key, either with a secret created externally or via helm
+  # When creating the secret externally, the secret must use kubecostAPIKey as the file name.
+  # Example: kubectl create secret generic kubecost-primary-api-secret --from-literal=kubecostAPIKey=<your-api-key>
+  createPrimaryApiSecret: false # disable if you want to use your own secret or use a secrets manager.
+  primaryApiSecretName: kubecost-primary-api-secret
+  primaryApiKey: "" # If set, createPrimaryApiSecret must be true
+  # The following are used for cloud service provider access, this is not required for in-cluster optimizations
+  createClusterControllerSecret: true
```
thomasvn, line 1192: I don't see that this is used anywhere else in the chart. Is this safe to just delete altogether?

cost-analyzer/templates/kubecost-cluster-controller-api-secret-template.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and .Values.clusterController.enabled .Values.clusterController.createPrimaryApiSecret }}
+{{- if and .Values.clusterController.createPrimaryApiSecret (lt (len .Values.clusterController.primaryApiKey) 5) }}
+{{- fail "\n\nWhen clusterController.createPrimaryApiSecret is true, primaryApiKey must be set." }}
```
thomasvn, line 3: Updating the error message here to be in line with the logic

```suggestion
{{- fail "\n\nWhen clusterController.createPrimaryApiSecret is true, primaryApiKey must be set and longer than 5 characters" }}
```

cost-analyzer/values.yaml
```diff
@@ -1176,10 +1176,22 @@ clusterController:
...
   affinity: {}
   nodeSelector: {}
-  primaryKubecostURL: ""  # URL for secondary clusters to connect to primary Kubecost (ex: https://kubecost.myorganization.com)
-  kubecostAPIKey: ""  # API Key for secondary clusters to authenticate with primary
-  createClusterControllerSecret: true  # disable if you want to use your own secret manager
+  # When using Cluster Controller from a secondary/agent cluster, you will need to set the following to allow
+  # the secondary to query the Primary cluster API for container right-sizing information.
+  # This method should also be used for the Kubecost Primary when using single-sign-on to use an api key (service account).
+  # These setting are not needed when running Kubecost on a single cluster without single-sign-on.
+  usePrimaryApiForRightSizing: false
```
thomasvn, line 1183: Is there a need for this config? Is it redundant with `primaryKubecostURL`? If `primaryKubecostURL` exists then we know we want to use it for rightSizing?

cost-analyzer/templates/kubecost-cluster-controller-template.yaml
```diff
@@ -282,22 +282,33 @@ spec:
...
+            {{- if .Values.clusterController.usePrimaryApiForRightSizing }}
+              {{- if not (lookup "v1" "Secret" .Release.Namespace .Values.clusterController.primaryApiSecretName) }}
+                {{- fail (printf "\n\nConfiguration error:\nSecret '%s' specified in clusterController.primaryApiSecretName does not exist in namespace %s" .Values.clusterController.primaryApiSecretName .Release.Namespace) -}}
+              {{- end -}}
+              {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.clusterController.primaryApiSecretName -}}
+              {{- if not (index $secret.data "kubecostAPIKey") }}
+                {{- fail (printf "\n\nConfiguration error:\nSecret '%s' specified in clusterController.primaryApiSecretName does not contain kubecostAPIKey" .Values.clusterController.primaryApiSecretName) -}}
+              {{- end -}}
+            {{- end -}}
+          {{- end }}
```
thomasvn, line 307: This looks like something that belongs in the `_helpers.tpl` file?

thomasvn (commented):
This generally looks good! However there are a lot of configuration changes here that would be introduced in v2.8.x, then likely removed by v3.0.

jessegoodier:
3.0 will fix this.

PR #4066, opened 2025-05-13
Add a Github Action that will require PRs to have a label for the target version of the product

nealormsbee:
## Release Notes Headline
N/A

## Commentary for Reviewers
Add a Github Action that will require PRs to have a label for the target version of the product.

## Links to Issues or Tickets

N/A

## User Impact and Risks
The impact will be to repo contributors. All pull requests will be required to set a label of the format `v$MAJOR.$MINOR`, denoting which major/minor release the PR is targeting. At least one of these must be set for all PRs.

## Testing

N/A

## Documentation Updates
N/A

PR #4074, opened 2025-05-14
Runner debug wip

cpetersen5:
## Release Notes Headline
N/A

## Commentary for Reviewers
This PR changes the runner for the `Check Labels` job from the unavailable `Kubecost-Linux-Small-x86` runner to the `ubuntu-latest` runner. The `Kubecost-Linux-Small-x86` runner is not configured for the `cost-analyzer-helm-chart` repo. 

## Links to Issues or Tickets
N/A - Troubleshooting https://github.com/kubecost/cost-analyzer-helm-chart/pull/4066/files#r2089165762 and https://kubecost.slack.com/archives/C044PFZN7SN/p1747216862744319. 

## User Impact and Risks
N/A

## Testing
- Successfully ran the `Check Labels` job on the `ubuntu-latest` runner when I opened the PR.

## Documentation Updates
N/A

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1723,10 +1723,10 @@ kubecostAggregator:
 
   resources:
     limits:
-      memory: 5Gi
+      memory: 2Gi
     requests:
       cpu: 1000m
-      memory: 5Gi
+      memory: 1Gi
 
   readinessProbe:
     enabled: true
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -1721,10 +1721,12 @@ kubecostAggregator:
     storageClass: ""  # default storage class
     storageRequest: 128Gi
 
-  resources: {}
-    # requests:
-    #   cpu: 1000m
-    #   memory: 1Gi
+  resources:
+    limits:
+      memory: 5Gi
+    requests:
+      cpu: 1000m
+      memory: 5Gi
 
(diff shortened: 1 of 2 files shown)


ameijer (approved):

cost-analyzer/values.yaml
```diff
@@ -1721,10 +1721,12 @@ kubecostAggregator:
     storageClass: ""  # default storage class
     storageRequest: 128Gi
 
-  resources: {}
-    # requests:
-    #   cpu: 1000m
-    #   memory: 1Gi
+  resources:
+    limits:
+      memory: 2Gi
```
jessegoodier, line 1726: I am not sure we want to set this by default- was this intentional?
Users will have to add a higher limit if they have a request that is greater than this limit or the sts will fail to create.

cpetersen5: Hey @jessegoodier, this change was intentional. Clickhouse is incredibly memory hungry and we were seeing repeated build failures on the aggregator pods due to memory constraints. We can play with these settings though.

thomasvn: @cpetersen5 If ClickHouse is very memory hungry, what's the intended effect of setting a Limit of `2Gi`? Is it to force garbage collection?

Regardless, I'm wary of setting a default memory limit on Aggregator that will be automatically inherited by all 3.0 users ... we should create a TODO to revisit this configuration.

PR #4076, opened 2025-05-16
fix a number of DAST scan alerts

nickcurie:
## Release Notes Headline
Improve security of basic kubecost install

## Commentary for Reviewers
Improves the default kubecost install to be more secure. All configs under `nginxHeaders.server` are intended for users to overwrite with more secure configs. As an example, setting `"Content-Security-Policy \"default-src 'self' data: api.userway.com cdn.userway.com api.mixpanel.com; connect-src http: https:; style-src 'self' 'unsafe-inline'; frame-ancestors 'none'; form-action 'self';\";"` will allow connections from all domains, so users should set this to a value that will allow for the use of context switching to known domains.

## Links to Issues or Tickets
Closes https://apptio.atlassian.net/browse/KCM-3837

## User Impact and Risks
All configs under `nginxHeaders.server` are intended for users to overwrite with more secure configs. As an example, setting `"Content-Security-Policy \"default-src 'self' data: api.userway.com cdn.userway.com api.mixpanel.com; connect-src http: https:; style-src 'self' 'unsafe-inline'; frame-ancestors 'none'; form-action 'self';\";"` will allow connections from all domains, so users should set this to a value that will allow for the use of context switching to known domains.

## Testing
Ran these changes against a local install of kubecost, observed the UI was functional. Ran ZAP scanner against this install and it did not come back with any alerts.

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -15,6 +15,7 @@ metadata:
     {{- include "cost-analyzer.commonLabels" . | nindent 4 }}
 data:
   nginx.conf: |
+    server_tokens off;
     gzip_static  on;
 
     # Enable gzip encoding for content of the provided types of 50kb and higher.
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -576,8 +576,9 @@ kubecostFrontend:
     server:
       - "'Access-Control-Allow-Origin' '*' always;"
       - "'Access-Control-Allow-Methods' 'GET, PUT, POST, DELETE, OPTIONS' always;"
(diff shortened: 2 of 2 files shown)


nickcurie:
This PR is reliant on some front-end changes, as we cannot remove the `unsafe-inline' piece of the CSP header (and therefore resolve some alerts) without breaking the UI. 
[2025-05-16-ZAP-Report-.pdf](https://github.com/user-attachments/files/20259442/2025-05-16-ZAP-Report-.pdf)

nickcurie:
`unsafe-inline` is still present in this PR, but has been moved to just `style-src` to reduce the security surface area. Since we are still waiting on front end changes to remove this, the thought was we can merge this PR as-is and come back later to remove `unsafe-inline`
[2025-05-21-ZAP-Report-.pdf](https://github.com/user-attachments/files/20371769/2025-05-21-ZAP-Report-.pdf)

ameijer (approved):

PR #4084, opened 2025-05-27
update applyRetroactively

JesseGoodier:
Force applyRetroactively to true per product team discussion.

## Release Notes Headline

NA, was not released previously. use ECP docs for guidance.

## Testing

always true:

```sh
helm upgrade ecp ~/git/cost-analyzer-helm-chart/cost-analyzer -f helmValues-kubecost-ecp.yaml --dry-run=server |grep ENTERPRISE_CUSTOM_PRICING_APPLY_RETROACTIVELY -A1
            - name: ENTERPRISE_CUSTOM_PRICING_APPLY_RETROACTIVELY
              value: "true"
```


chipzoller (commented):
What if we just eliminate the value from the values file entirely rather than leave it with a warning?

jessegoodier:
> What if we just eliminate the value from the values file entirely rather than leave it with a warning?

works for me!

chipzoller (approved):

PR #4090, opened 2025-05-29
custom json of instance types mounted in aggregator statefulset to be populated into the DB

avrodrigues5:
## Release Notes Headline

Provides the ability for the user to mount a custom JSON file containing instance types to be used in the recommendation.

## Commentary for Reviewers

This address our problem in custom provider environment to provide recommendation.

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-4113
https://apptio.atlassian.net/browse/KCM-3920

## User Impact and Risks

None

## Testing
Tested if with my cluster 

## Documentation Updates
Maybe the whole EPIC in future.


cost-analyzer/values-instance-types.yaml
```diff
@@ -0,0 +1,7 @@
+instanceTypes:
```
nikovacevic, line 1: @chipzoller this setting provides a way for users to configure a set of custom instance types for node group sizing. Do you like `instanceTypes` or would you prefer `customInstanceTypes`?

nikovacevic (approved):
This looks good to me. Let's let Chip scan the values here to see if the format looks right. I don't necessarily anticipate a lot of users going this route, so it may be OK as-is. (It's useful for testing, regardless, Chip.)

cost-analyzer/values-instance-types.yaml
chipzoller: We would just be talking about established, CSP instance types (i.e., names) here, right? This wouldn't apply to non-CSP clusters?

avrodrigues5: > We would just be talking about established, CSP instance types (i.e., names) here, right? This wouldn't apply to non-CSP clusters?

Yes established Cloud service provider but for an on-prem users the provider is defaulted to custom in our product and this provides a way to upload instances types that they want recommended for on-prem providers.

My Idea around using


```
instanceTypes:
  enabled: false
  #  kubectl create configmap -n kubecost custom-instance-types --from-file custom.json
  custom:
    configMapName: custom-instance-types
    location:
      URI: /var/configs/instance-types/custom.json
```

if we want to provide an ability in future for customer to mount aws.json , everything under one top-level configuration key. 

```
instanceTypes:
  enabled: false
  #  kubectl create configmap -n kubecost custom-instance-types --from-file custom.json
  custom:
    configMapName: custom-instance-types
    location:
      URI: /var/configs/instance-types/custom.json
  aws:
    configMapName: aws-instance-types
    location:
      URI: /var/configs/instance-types/aws.json
```

avrodrigues5: address concern with Chip and Niko in DM !

PR #4091, opened 2025-05-30
[KCM-3819] default routing to aggregator

ameijer:
## Release Notes Headline

we no longer serve any endpoints from the cost model, and that container is getting deprecated as of 3.0. this PR then by default routes to aggregator

## Commentary for Reviewers

cost model won't even get deployed anymore. nice thing about this too is that new api endpoints won't need helm chart PRs anymore. yes this will break a lot of stuff on nightly but don't hate the player, hate the game. 

## Links to Issues or Tickets
https://apptio.atlassian.net/browse/KCM-3819

## User Impact and Risks

High impact. impacts are that old endpoints like getCOnfigs that the FE needs are no longer there. we will update the FE to not need those, or subsequent PRs will return skeletonized versions 

## Testing

Mnaully tested

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -1663,7 +1663,7 @@ for more information
   {{- else if .Values.imageVersion }}
     {{ .Values.kubecostModel.image }}:{{ .Values.imageVersion }}
   {{- else if eq "development" .Chart.AppVersion }}
-    gcr.io/kubecost1/cost-model-nightly:latest
+    gcr.io/guestbook-227502/agent:latest
   {{- else }}
     {{ .Values.kubecostModel.image }}:prod-{{ $.Chart.AppVersion }}
   {{- end }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -738,7 +738,7 @@ spec:
               mountPath: /var/configs/auth
             {{- end }}
           env:
-          {{- if or .Values.kubecostModel.federatedStorageConfigSecret .Values.kubecostModel.federatedStorageConfig }}
+            {{- if or .Values.kubecostModel.federatedStorageConfigSecret .Values.kubecostModel.federatedStorageConfig }}
             - name: EXPORT_BUCKET_CONFIG_FILE
               value: /var/configs/etl/federated/federated-store.yaml
             {{- end }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -631,6 +631,8 @@ spec:
           {{- toYaml .Values.global.containerSecuritycontext | nindent 12 }}
           {{- end }}
           volumeMounts:
+            - name: tmp
+              mountPath: /tmp
             - name: persistent-configs
               mountPath: /var/configs
             {{- if .Values.extraVolumeMounts }}
--- a/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-deployment-template.yaml
@@ -736,8 +736,29 @@ spec:
               mountPath: /var/configs/auth
             {{- end }}
           env:
+          {{- if or .Values.kubecostModel.federatedStorageConfigSecret .Values.kubecostModel.federatedStorageConfig }}
+            - name: EXPORT_BUCKET_CONFIG_FILE
+              value: /var/configs/etl/federated/federated-store.yaml
+            {{- end }}
+            {{- if .Values.kubecostModel.minuteMetricsEnabled }}
+            - name: MINUTE_METRICS_ENABLED
+              value: "true"
+            {{- end }}
+            {{- if .Values.kubecostModel.opencostSourceEnabled }}
+            - name: OPENCOST_SOURCE_ENABLED
+              value: "true"
+            {{- end }}
+            {{- if .Values.kubecostModel.collectorDataSourceEnabled }}
+            - name: COLLECTOR_DATA_SOURCE_ENABLED
+              value: "true"
+            {{- end }}
+            {{- if .Values.kubecostModel.cloudyEmitterEnabled }}
+            - name: CLOUDY_EMITTER_ENABLED
+              value: "true"
+            {{- end }}
             - name: CONTAINER_IMAGE_TAG
               value: {{ include "cost-model.imagetag" . }}
+
             {{- if .Values.global.grafana }}
             - name: GRAFANA_ENABLED
               value: "{{ template "cost-analyzer.grafanaEnabled" . }}"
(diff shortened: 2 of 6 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -53,13 +53,15 @@ data:
...
-{{- end }}
+        {{- if .Values.kubecostFrontend.useDefaultFqdn }}
+        server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}.svc.cluster.local:9004;
+        {{- else }}
+        {{- if (.Values.kubecostFrontend.aggregator).fqdn }}
+        server {{ .Values.kubecostFrontend.aggregator.fqdn }};
+        {{- else }}
+        server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}:9004;
+        {{- end }}
+        {{- end }}
```
thomasvn: Minor, but I think more readable?

```suggestion
        {{- if .Values.kubecostFrontend.useDefaultFqdn }}
        server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}.svc.cluster.local:9004;
        {{- else if (.Values.kubecostFrontend.aggregator).fqdn }}
        server {{ .Values.kubecostFrontend.aggregator.fqdn }};
        {{- else }}
        server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}:9004;
        {{- end }}
```

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -109,17 +111,6 @@ data:
...
-        {{- if .Values.kubecostFrontend.useDefaultFqdn }}
-        server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}.svc.cluster.local:9004;
-        {{- else }}
-        {{- if (.Values.kubecostFrontend.aggregator).fqdn }}
-        server {{ .Values.kubecostFrontend.aggregator.fqdn }};
-        {{- else }}
-        server {{ .Release.Name }}-aggregator.{{ .Release.Namespace }}:9004;
-        {{- end }}
-        {{- end }}
-    }
```
thomasvn, line 122: Another minor thing for readability.

Kind of confusing that we are deleting this `upstream`, but keeping the `upstream model`? Everything routes to model which routes to aggregator? Would it be simpler if we try to use the `/aggregator` route everywhere in this config?

And then `upstream model` should just reroute to `/aggregator`?

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -389,1235 +379,126 @@ data:
...
         location /model/authrbac {
-            proxy_pass http://aggregator/isAdminAuthenticated;
+            proxy_pass http://model/isAdminAuthenticated;
         }
     {{- end }}
     {{- end }}
 
 
 {{- if and (not .Values.agent) (not .Values.cloudAgent) (not (eq (include "aggregator.deployMethod" .) "disabled")) }}
-    # TODO make aggregator route the default, start special-casing
```
thomasvn, line 420: Thanks for deleting all this. Looks like all the lines below simply route specific `/model/` endpoints to the Aggregator pod.

thomasvn:
A couple minor things that I think could really enhance the readability of this config. If this is urgent though, happy to approve and get this merged asap.

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
ameijer: addressed, I think?

cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
ameijer: the remaining endpoints here actually route to the cloud cost pod, which is why they need to stay

cost-analyzer/templates/_helpers.tpl
```diff
@@ -1663,7 +1663,7 @@ for more information
   {{- else if .Values.imageVersion }}
     {{ .Values.kubecostModel.image }}:{{ .Values.imageVersion }}
   {{- else if eq "development" .Chart.AppVersion }}
-    gcr.io/kubecost1/cost-model-nightly:latest
+    gcr.io/guestbook-227502/agent:latest
```
thomasvn, line 1666: Have we decided this is where we'll be pushing changes to the agent in the near term?

ameijer: I have sort of made that decision unilaterally just to get us off the ground here. This will likely need to move to icr before 3.0 anyway

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -736,8 +738,29 @@ spec:
               mountPath: /var/configs/auth
             {{- end }}
           env:
+          {{- if or .Values.kubecostModel.federatedStorageConfigSecret .Values.kubecostModel.federatedStorageConfig }}
```
thomasvn: Nit

```suggestion
            {{- if or .Values.kubecostModel.federatedStorageConfigSecret .Values.kubecostModel.federatedStorageConfig }}
```

cost-analyzer/templates/_helpers.tpl
thomasvn: Good with me!

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -753,56 +776,10 @@ spec:
...
-              value: {{ .Values.groupFiltersConfigmapName }}
-            {{- end }}
-            {{- if .Values.pricingConfigmapName }}
-            - name: PRICING_CONFIGMAP_NAME
-              value: {{ .Values.pricingConfigmapName }}
-            {{- end }}
-            {{- if .Values.metricsConfigmapName }}
-            - name: METRICS_CONFIGMAP_NAME
-              value: {{ .Values.metricsConfigmapName }}
-            {{- end }}
```
thomasvn, line 803: The agent _may_ still be using these env vars. Just in case let's keep these in the Helm chart?

https://github.com/opencost/opencost/blob/629d9b8c6c4361891a9172553174c3b23603f830/pkg/costmodel/router.go#L555-L556

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -818,23 +795,12 @@ spec:
...
             - name: CONFIG_PATH
               value: /var/configs/
-            - name: DB_PATH
-              value: /var/db/
-            - name: CLUSTER_PROFILE
-            {{- if .Values.kubecostProductConfigs }}
-              value: {{ .Values.kubecostProductConfigs.clusterProfile | default "production" }}
-            {{- else }}
-              value: production
-            {{- end }}
```
thomasvn, line 828: I _think_ this is safe to remove, but it is still referenced in the OpenCost agent

https://github.com/opencost/opencost/blob/629d9b8c6c4361891a9172553174c3b23603f830/pkg/costmodel/clusterinfo.go#L21

This env var is fed to the `clusterInfo` endpoint, which I know we are trying to remove dependence on.

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -844,57 +810,12 @@ spec:
...
-                  name: {{ .Values.global.prometheus.queryServiceBasicAuthSecretName }}
-                  key: PASSWORD
-            {{- end }}
-            {{- if .Values.global.prometheus.queryServiceBearerTokenSecretName }}
-            - name: DB_BEARER_TOKEN
-              valueFrom:
-                secretKeyRef:
-                  name: {{ .Values.global.prometheus.queryServiceBearerTokenSecretName }}
-                  key: TOKEN
-            {{- end }}
```
thomasvn, line 865: If there's a possibility that users are "allowed" to run the 3.0 agent using Prom as its datasource, this may want to be left in for now?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -844,57 +810,12 @@ spec:
...
             {{- if .Values.global.prometheus.insecureSkipVerify }}
             - name: INSECURE_SKIP_VERIFY
               value: {{ (quote .Values.global.prometheus.insecureSkipVerify) }}
             {{- end }}
-            {{- if .Values.global.prometheus.kubeRBACProxy }}
-            - name: KUBE_RBAC_PROXY_ENABLED
-              value: {{ (quote .Values.global.prometheus.kubeRBACProxy) }}
-            {{- end }}
```
thomasvn, line 873: Ditto here?

cost-analyzer/templates/cost-analyzer-deployment-template.yaml
```diff
@@ -844,57 +810,12 @@ spec:
...
-                  key: AWS_ACCESS_KEY_ID
-            - name: AWS_SECRET_ACCESS_KEY
-              valueFrom:
-                secretKeyRef:
-                  name: {{ .Values.pricingCsv.location.csvAccessCredentials }}
-                  key: AWS_SECRET_ACCESS_KEY
-            {{- end }}
-            {{- end }}
-            {{- end }}
-            {{- end }}
```
thomasvn, line 897: OpenCost does still have references to "CSV Provider".

However, I believe because Kubecost 2.8 will have its own notion of Enterprise Custom CSV Pricing which happens all on the Aggregator, this is safe to remove from the Kubecost Helm chart?

thomasvn:
Looks good, thanks for so much cleanup here @ameijer! There are a few env vars here which I might leave in to be safe, but mostly looks good to go! Lmk if you there's anything you disagree with.

thomasvn (approved):
Going to approve for now! This should be ok to merge as is, and if we find that the 3.0 agent _does_ require some env vars added back in, we can do that.

ameijer:
thanks thomas. I think these are good open questions. going to resolve them simply because that is what is required to merge this, but let's revisit this continually over the next few months

PR #4094, opened 2025-06-02
[KCM-3819] return only version info in installInfo response

ameijer:
## Release Notes Headline

return helm chart info in installinfo. this is the bare minimum of what the FE needs to check API version compatibility. We will get this information from release time variables, rather than reading a container

## Commentary for Reviewers

should help the FE boot

## Links to Issues or Tickets
https://apptio.atlassian.net/browse/KCM-3819

## User Impact and Risks

minimal, should restore some functionality to the FE 

## Testing

tested via curl, FE will adjust to accommodate

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -585,9 +585,8 @@ data:
                 }
             ';
         }
-    }
 
-    location /model/installInfo {
+        location /model/installInfo {
             default_type 'application/json';
             return 200 '\n
                 {
@@ -595,6 +594,8 @@ data:
                 }
             ';
         }
+    }
+
 
 {{- end }}
 {{- end }}
--- a/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
@@ -587,5 +587,14 @@ data:
         }
     }
 
+    location /model/installInfo {
+            default_type 'application/json';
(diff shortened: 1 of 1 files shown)


cost-analyzer/templates/cost-analyzer-frontend-config-map-template.yaml
```diff
@@ -587,5 +587,14 @@ data:
...
     }
 
+    location /model/installInfo {
+            default_type 'application/json';
+            return 200 '\n
+                {
+                    "version": "{{ .Chart.Version }}"
+                }
+            ';
+        }
```
thomasvn: Looks good! Minor indentation/style change? I've found that when this Nginx conf is set to return JSON, it does a poor job with newlines and indentation. I say we let the client-side prettify the JSON.

```suggestion
    location /model/installInfo {
        default_type 'application/json';
        return 200 '{"version": "{{ .Chart.Version }}"}';
    }
```

thomasvn:
This looks good! One last thing before we merge, could you test this template change with the following command and record your results?

```sh
$ helm template ./cost-analyzer --debug
```

I want to make sure that this addition gets templated into the frontend configmap.

thomasvn:
When I tested this, it looks like `/model/installInfo` is being templated outside of the `server {}` code block. I think we may need to move this Location up into the server block?

I ran the following commands to get this [result.txt](https://github.com/user-attachments/files/20559648/result.txt):

```sh
cd cost-analyzer
helm template . --show-only templates/cost-analyzer-frontend-config-map-template.yaml
```

thomasvn (approved):
Tested this, looks good! Thank you

PR #4095, opened 2025-06-02
Cleanup and remove old files

thomasvn:
## Release Notes Headline

Removes the `kubecost.yaml` file from the repo, as it was no longer actively maintained.

## Commentary for Reviewers

- `Chart.yaml` now references `0.0.0-dev` as its version. This is to clarify the fact that the version of this Chart is on the develop branch. Kubecost's release scripts will automatically update this `Chart.yaml` version during release ([ref](https://github.com/kubecost/release-scripts/blob/c86d21acc44524969b5f694576b5dca1a35e204e/.github/actions/update-helm-chart/action.yaml#L42-L49)).
- Remove `kubecost.yaml`. It is no longer receiving periodic updates. It was last updated for v2.3.3
- Remove `kustomization.yaml`. It did not serve a purpose specifically for this repo, and was only an example of how you can post-process the `kubecost.yaml` file if needed.
- Remove `historical-builds` directory.

## Links to Issues or Tickets

- https://apptio.atlassian.net/browse/KCM-3699

## User Impact and Risks

There is a low likelihood that users are still leveraging the `kubecost.yaml` file, as it has not received frequent updates. Risk remains low, because the file can be generated from the user's end by running `helm template`.

## Testing

N/A

## Documentation Updates

N/A


cost-analyzer/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: development
 description: Kubecost Helm chart - monitor your cloud costs!
 name: cost-analyzer
-version: 2.3.3
+version: 0.0.0-dev  # This is a placeholder used on the develop branch. If installing Kubecost, please reference a release branch.
```
thomasvn, line 5: This field requires semVer, which is why I used `0.0.0-dev`. Open to any better suggestions!

chipzoller (approved):

PR #4097, opened 2025-06-04
provide a way to configure FE defaults via helm

ameijer:
## Release Notes Headline
Provide defaults to the FE via helm 

## Commentary for Reviewers
re-implements the getconfigs endpoint to provide reasonable defaults for the FE, and allows the customer to set those 

## User Impact and Risks
None, the FE provides defaults if something goes wrong 

## Testing

tested in nightly


JesseGoodier (approved):

jessegoodier:
I'm looking at uncommenting all values. this is an interesting start to that.

PR #4098, opened 2025-06-04
Document networkCosts.prometheusScrape config option

shreyas44:
## Release Notes Headline

Document `networkCosts.prometheusScrape` config option

I think a better change would be to merge this with `networkCosts.enable` in `cost-analyzer/templates/cost-analyzer-network-costs-service-template.yaml`. Is there a reason they're separate?


JesseGoodier (approved):
thanks for this!

PR #4103, opened 2025-06-12
Fix scope of if statement on cloud cost

Sean-Holcomb:
## Commentary for Reviewers
The scope of the if statement which cloudCost.enabled controlled was incorrect in that disabling it only disabled the check on the existence of the secret. I have fixed the scope so that it completely turns off the deployment, and added a helper that checks for cloud cost configuration.

## Testing
Tested using `helm template`


Sean-Holcomb:
Outdated https://github.com/kubecost/cost-analyzer-helm-chart/pull/4137

PR #4109, opened 2025-06-18
Add and mount values configmap to aggregator

kaelanspatel:
## Commentary for Reviewers

Mounts a configmap containing user-configured helm values to the aggregator. This replaces the `HELM_VALUES` env variable that used to be present in the cost-analyzer, so as to serve a `/helmValues` endpoint.

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-3879

## User Impact and Risks

This potentially exposes values that would have been auto-added to secrets as values in this configmap. From my understanding, this is no different from the `HELM_VALUES` variable and is the desired state from the ask.

## Testing

Tested on new cluster install by upgrading using local helm chart and though exec into container.
Tested through helm template.
Tested in conjunction with new helm endpoint to encode/send via GET.

### DIFF
--- a/kubecost/templates/aggregator/aggregator-statefulset.yaml
+++ b/kubecost/templates/aggregator/aggregator-statefulset.yaml
@@ -76,6 +76,9 @@ spec:
         - name: aggregator-staging
           emptyDir:
             sizeLimit: {{ .Values.aggregator.stagingEmptyDirSizeLimit }}
+        - name: user-helm-values
+          configMap:
+            name: helm-values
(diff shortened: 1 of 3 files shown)


jessegoodier:
if secrets are used, they should be in externalSecret type tools. The risk here is acceptable.

JesseGoodier (approved):

thomasvn:
@kaelanspatel Could you cherry-pick this into v3.0 branch please?

PR #4110, opened 2025-06-19
[KCM-4014] revamp the license file to support v2 licensing

ameijer:
## Release Notes Headline

Fix how we create and use license delivery in the chart 

## Commentary for Reviewers

we weren't actually mounting the provided license in the aggregator statefulset. this fixes that and makes us read the helm values better

## Testing

tested via manual install testing, confirmed license delivered in this way was read


JesseGoodier (approved):
you even got the configsChecksum. 

thanks for this.

PR #4128, opened 2025-06-30
fix: render prometheus recording rules

smauermann:
## Release Notes Headline

Restore the creation of the bundled prometheus recording rules when enabled in the values file (`.Values.prometheusRule.enabled`).

## Commentary for Reviewers

Rendering of the prometheus recording rules was broken with this commit https://github.com/kubecost/cost-analyzer-helm-chart/commit/a1ace4405f68e150b28ef05ff5d49c814334c439. 

## Links to Issues or Tickets

N/A

## User Impact and Risks

No risks, the intended functionality was restored.

## Testing

I've tested the changes by rendering out the templates and confirming the recording rule gets rendered according to the value of `.Values.prometheusRule.enabled`.

## Documentation Updates

N/A

### DIFF
--- a/cost-analyzer/templates/cost-analyzer-prometheusrule-template.yaml
+++ b/cost-analyzer/templates/cost-analyzer-prometheusrule-template.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.prometheus }}
 {{- if .Values.prometheus.serverFiles }}
-{{- if .Values.prometheus.serverFiles.rules }}
+{{- if (index .Values.prometheus.serverFiles "recording_rules.yml") }}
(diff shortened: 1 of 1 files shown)


jessegoodier:
@smauermann Thank you for your thoughtful changes.

We are removing prometheus from the next version of Kubecost. If you would like to fix this in 2.x, you can target the change to the v2.8 branch.

3.0 is targeted for late August, but subject to change based on external factors.

smauermann:
Hi @jessegoodier, thanks for taking a look! I have opened a new PR targeting the v2.8 branch. https://github.com/kubecost/cost-analyzer-helm-chart/pull/4135

PR #4137, opened 2025-07-02
Sth/architecture changes

Sean-Holcomb:
## Commentary for Reviewers

This PR makes changes that align the template structure of the chart with the architecture outlined in 
https://ibm-my.sharepoint.com/:w:/p/kai_wombacher/EZW3dpeHPYJJq-AVSgDdTwIBF0As4kxeYyZts2bsh6ZaCA?e=EJ4BZw

Changes of note:
- Remove definitions and control which imply multiple deployment methods
- Put templates in directories which mirror component structure
- Use of finops-agent as sub chart
  - Move definition of cluster id to global value so that it can be applied to sub charts
  - rework export bucket for compatibility with sub charts
- ban the use of the words "ETL" and "Federated"


thomasvn (commented):
We probably want to use this as an opportunity to remove `.Values.kubecostAggregator.deployMethod` as well?

cost-analyzer/templates/federated-store-secret.yaml
```diff
@@ -1,12 +0,0 @@
-{{- if .Values.kubecostModel.federatedStorageConfig -}}
```
thomasvn, line 1: Why are we deleting the `federated-store` Secret? The Aggregator and CloudCosts containers still need this right?

thomasvn: Oh, or did you intend to replace this with the `.Values.kubecost.exportBucket`?

Sean-Holcomb: yes something like that. this secret is shared use with the finops-agent subchart so I am trying to move us away from current configuration paradigm to something that works a little bit better

cost-analyzer/templates/_helpers.tpl
```diff
@@ -72,238 +36,114 @@ Kubecost 2.0 preconditions
...
-{{/*
-Actions Storage source contents check. Either the Secret must be specified or the YAML, not both.
-*/}}
-{{- define "actionsStorageSourceCheck" -}}
-  {{- if ((.Values.kubecostProductConfigs).actions).enabled -}}
-  {{- if and ((.Values.kubecostProductConfigs).actions).storageConfigSecret ((.Values.kubecostProductConfigs).actions).storageConfig -}}
-    {{- fail "\nkubecostProductConfigs.actions.storageConfigSecret and kubecostProductConfigs.actions.storageConfig are mutually exclusive. Please specify only one." -}}
-  {{- end -}}
-  {{- end -}}
-{{- end -}}
```
thomasvn: @Sean-Holcomb This is new code that @biancaburtoiu has introduced for our 3.0 MultiCluster Actions feature. Can we please keep this function?

thomasvn:
Just to confirm, we expect users to be specifying exactly one of the two?

- `.Values.federatedStorage.configYAML`. This would be their YAML secret directly in Helm Values
- Or `.Values.global.federatedStorage.existingSecret`

cost-analyzer/values.yaml
```diff
@@ -6,6 +6,17 @@ global:
...
   zone: ""  # cluster.local
 
+  clusterId: cluster-one  # globally unique cluster name
+
+  ## The name of the Secret containing a bucket config for federated storage.
+  ## The contents should be stored under a key named storage-config.yaml by default.
+  ## Ref: https://www.ibm.com/docs/en/kubecost/self-hosted/2.x?topic=cluster-long-term-federated-storage-configuration
+#  federatedStorage:
+#    existingSecret: federated-storage-config
+#    fileName: federated-store.yaml # set this value to change what the file name for the yaml config is
```
thomasvn: This is a minor suggestion, and can be addressed in a future PR. My preference is for all configs to be explicitly defined in the `values.yaml` file. No configs that are commented out.

Should we instead define `existingSecretName` here? Its default value is an empty string, which in Helm is considered an empty value. Only when a user defines a string here will we assume that a secret exists.

```suggestion
  federatedStorage:
    existingSecretName: ""
    fileName: federated-store.yaml # set this value to change what the file name for the yaml config is
```

cost-analyzer/templates/_helpers.tpl
```diff
@@ -72,238 +36,114 @@ Kubecost 2.0 preconditions
...
-      {{- fail "\n\nWhen using multi-cluster Prometheus, you must specify a unique `(((((.Values.prometheus).server).global).external_labels).cluster_id)` for each cluster.\nNote this must be a globally unique identifier.\n" -}}
+    {{- printf "\n\nIn Kubecost 3.0, `.Values.prometheus.server.global.external_labels.cluster_id` is no longer required.\nWhen it is set, it is used for backwards compatibility. \nSee TODO for more information.\n" -}}
+  {{- end }}
+  {{- if (.Values.kubecostProductConfigs).clusterName }}
+    {{- printf "\n\nIn Kubecost 3.0, `.Values.prometheus.server.global.external_labels.cluster_id` is no longer required.\nWhen it is set, it is used for backwards compatibility. \nSee TODO for more information.\n" -}}
+  {{- end }}
+  {{- if not .Values.global.clusterId }}
+    {{- fail "\n\nIn Kubecost 3.0, `.Values.global.clusterId` is required to be set"}}
+  {{- end }}
+  {{- if or (.Values.global.federatedStorage).existingSecret ((.Values.federatedStorage).secret).config }}
```
thomasvn, line 98: Are you checking the right config here? Should it instead be `.Values.federatedStorage.configYAML`?

cost-analyzer/templates/_helpers.tpl
```diff
@@ -72,238 +36,114 @@ Kubecost 2.0 preconditions
...
-{{-  if .Capabilities.APIVersions.Has "v1/Secret" }}
-  {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.kubecostModel.federatedStorageConfigSecret }}
-  {{- if or (not $secret) (not (index $secret.data "federated-store.yaml")) }}
-    {{- fail (printf "The federated storage config secret '%s' does not exist or does not contain the expected key 'federated-store.yaml'" .Values.kubecostModel.federatedStorageConfigSecret) }}
-  {{- end }}
-{{- end -}}
-{{- end -}}
-{{- end -}}
+{{- define "kubecost.federatedStorage.source.check" -}}
+  {{- if and ((.Values.global).federatedStorage).existingSecret ((.Values.federatedStorage).secret).config -}}
```
thomasvn, line 128: Again here, do you mean to check `.Values.federatedStorage.configYAML` instead?

JesseGoodier (approved):
Thanks for all the effort here. this is not easy.

will continue to refine the backwards compatibility.

for now, we will need to use:
```sh
helm template kubecost ../cost-analyzer  \
  --set global.clusterId=xxx
```

federated store via yaml works. 
But we will need to test BYO secrets.

cost-analyzer/templates/_helpers.tpl
jessegoodier: added back

cost-analyzer/templates/_helpers.tpl
jessegoodier: we will need to address all checks for v3. good callout

PR #4144, opened 2025-07-13
Adding a Service Account to Forecasting Deployment 

heysashy:
## Release Notes Headline
Adding a service account for forecasting deployment 

## Commentary for Reviewers
An Option to add a service account to forecasting deployment. 

Service accounts should be added to every Kubernetes deployment to implement the principle of least privilege, ensuring each application only has access to the specific resources it needs rather than using the shared default service account with potentially broad permissions. This approach provides better security isolation by limiting the blast radius if a container is compromised, preventing attackers from accessing resources beyond what that specific application requires. Additionally, dedicated service accounts enable proper audit trails and compliance by making it possible to track which application performed specific actions within the cluster. 

## Links to Issues or Tickets
N/A 

## User Impact and Risks
No impact, this is an optional field to add.

### DIFF
--- a/cost-analyzer/values.yaml
+++ b/cost-analyzer/values.yaml
@@ -848,7 +848,10 @@ forecasting:
     repository: gcr.io/kubecost1/kubecost-modeling
     tag: v0.1.27
   imagePullPolicy: IfNotPresent
-
+  
+  # Set Service Account For Forecast # optional 
+  # serviceAccountName: ""
+  
   # Resource specification block for the forecasting container.
   resources:
     requests:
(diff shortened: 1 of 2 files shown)


raymondsquared:
Thanks for raising this PR!

Relying on the default service account in k8s isn't a good idea for several reasons and some organisations actually blocked the usage of using default service account on their policy checker.

cost-analyzer/values.yaml
```diff
@@ -849,6 +849,9 @@ forecasting:
     tag: v0.1.27
   imagePullPolicy: IfNotPresent
 
+  # Set Service Account For Forecast # optional
+  # serviceAccountName: ""
```
```suggestion
  serviceAccountName: ""
```

JesseGoodier (approved):

jessegoodier:
sorry we didn't merge this sooner.

Now there are too many changes, I'm just going to do a new PR. Thanks for this @heysashy

PR #4168, opened 2025-07-29
3.0 fixes for federated store/aggregator services

JesseGoodier:
## Release Notes Headline

1. Allow using old kubecostModel.federated store secrets
2. Add servicemonitor for clickhouse
3. Fix various templates

## Commentary for Reviewers

tested on both RBAC and non-rbac envrionments

## Testing

See config for clickhouse exporter on saml-keyclock in nigthly


kubecost/templates/aggregator/aggregator-servicemonitor.yaml
```diff
@@ -28,4 +28,40 @@ spec:
   selector:
     matchLabels:
     {{- include "kubecost.aggregator.commonLabels" . | nindent 6 }}
+    {{- if .Values.aggregator.service.labels }}
+    {{- toYaml .Values.aggregator.service.labels | nindent 6 }}
+    {{- end }}
+---
+apiVersion: monitoring.coreos.com/v1
```
jessegoodier, line 35: we may change this before GA. but I don't see a reason to not enable both when enabling one for now.

kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -517,6 +520,18 @@ spec:
...
+            - name: ETL_HOURLY_STORE_DURATION_HOURS
+              value: "49"
+            - name: FEDERATED_CLUSTER
+              value: "true"
+            - name: FEDERATED_PRIMARY_CLUSTER
+              value: "true"
+            - name: ETL_BUCKET_CONFIG
+              value: /var/configs/federated-store.yaml
+            - name: FEDERATED_STORE_CONFIG
+              value: /var/configs/federated-store.yaml
```
ameijer: these can be removed

kubecost/templates/aggregator/aggregator-service.yaml
```diff
@@ -16,12 +16,16 @@ spec:
...
+      targetPort: 9004  
+    {{- if or .Values.saml.enabled .Values.oidc.enabled}}    
+    - name: tcp-api-rbac
       port: 9008
       targetPort: 9008
+    {{- else }}
     {{- end }}
+    - name: db-metrics
+      port: 9363
+      targetPort: 9363
```
ameijer, line 28: should we wrap this in a conditional?

kubecost/templates/aggregator/aggregator-statefulset.yaml
jessegoodier: done

kubecost/templates/aggregator/aggregator-service.yaml
jessegoodier: for now, I think we should add the exporter always. up to the user to scrape it or not.
and we may want to merge (these metrics) this with the existing.
and by merge, I mean cherry-pick what we find to be important.

ameijer (approved):

PR #4175, opened 2025-08-05
add heartbeat flag and retention

mittal-ishaan:
## Release Notes Headline
Add heartbeat enabled flag and config for setting heartbeat retention

## User Impact and Risks
Users will be able to disable heartbeat and set retention for it

## Testing
used helm template cmd to check if the envs were getting populated in the output manifest

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -1033,7 +1033,11 @@ aggregator:
 ## retention: deletes data beyond specified retention period
 diagnostics:
   enabled: true
-  retention: 90d
+  retention: 30d
+
+heartbeat:
+  enabled: true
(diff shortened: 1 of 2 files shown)


kubecost/values.yaml
```diff
@@ -1033,7 +1033,11 @@ aggregator:
 ## retention: deletes data beyond specified retention period
 diagnostics:
   enabled: true
-  retention: 90d
+  retention: 30d
+
+heartbeat:
+  enabled: true
+  retention: 7d
```
ameijer:
```suggestion
  retention: 30d
``

ameijer (approved):

PR #4180, opened 2025-08-07
Atm/kcm 4354

ameijer:
## Release Notes Headline

fixes for cluster controller v2

## Commentary for Reviewers

some small fixes related to expected keys in configmaps

### DIFF
--- a/kubecost/templates/aggregator/aggregator-statefulset.yaml
+++ b/kubecost/templates/aggregator/aggregator-statefulset.yaml
@@ -485,6 +485,12 @@ spec:
             - name: DIAGNOSTICS_RETENTION
               value: {{ .Values.diagnostics.retention | quote }}
             {{- end }}
+            {{- if .Values.heartbeat.enabled }}
+            - name: HEARTBEAT_ENABLED
+              value: "true"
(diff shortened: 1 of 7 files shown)


kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -399,7 +399,7 @@ spec:
               mountPath: /var/configs/actions/storage
             - name: federated-storage-config
               mountPath: /var/configs/actions/storage/actions-store.yaml
-              subPath: federated-store.yaml
+              subPath: actions-store.yaml
```
nickcurie, line 402: I don't really get what we're doing here. We're mounting the actions store to the federated storage? Are both the actions-store.yaml and federated-store.yaml the same file but renamed?

ameijer: that is exactly what is going on here! we want to make it easy for people to only give 1 secret for both allocations AND actions

nickcurie: Ok makes sense!

nickcurie (approved):
LGTM

PR #4185, opened 2025-08-11
fix federated-store

JesseGoodier:
Added toYaml and comment on backward compatibility.

I will say that the lack of consistency here is not ideal (config: vs federatedStorageConfig) but this isn't critical to me.

@thomasvn do you have concerns?

tested. format is working properly for all 3 options.

```yaml
kubecostModel:
  federatedStorageConfig: |-
    type: S3
    config:
      bucket: demo-data
      endpoint: s3.amazonaws.com
      region: us-west-2
```
```yaml
federatedStorage:
  config: |-
    type: S3
    config:
      bucket: demo-data
      endpoint: s3.amazonaws.com
      region: us-west-2
```
```yaml
global:
  federatedStorage:
    config: |-
      type: S3
      config:
        bucket: demo-data
        endpoint: s3.amazonaws.com
        region: us-west-2
```

### DIFF
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -548,21 +548,24 @@ federated storage config helpers
   {{- end }}
 {{- end -}}
 
+    {{/*
+    NOTE: added kubecostModel for backward compatibility
+    */}}
 {{- define "kubecost.federatedStorage.config" }}
   {{- if (.Values.kubecostModel).federatedStorageConfig -}}
-    {{- (.Values.kubecostModel).federatedStorageConfig -}}
-  {{- else if (.Values.federatedStorage).config }}
-    {{- (.Values.federatedStorage).config -}}
+    {{- toYaml (.Values.kubecostModel).federatedStorageConfig -}}
+  {{- else if (.Values.federatedStorage).config -}}
+    {{- toYaml (.Values.federatedStorage).config -}}
   {{- else if (.Values.global.federatedStorage).federatedStorageConfig -}}
-    {{- (.Values.global.federatedStorage).federatedStorageConfig -}}
-  {{ else }}
+    {{- toYaml (.Values.global.federatedStorage).federatedStorageConfig -}}
+  {{- else -}}
     {{/*
     TODO:Default federated storage config 
     for single cluster environments
     */}}
-    {{- printf "localClusterBucket" }}
-  {{- end }}
-{{- end }}
+    {{- printf "localClusterBucket" -}}
+  {{- end -}}
+{{- end -}}
 
 {{- define "kubecost.federatedStorage.fileName" -}}
 {{ default "federated-store.yaml" (.Values.global.federatedStorage).fileName }}
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -548,16 +548,16 @@ federated storage config helpers
   {{- end }}
 {{- end -}}
 
-    {{/*
-    NOTE: added kubecostModel for backward compatibility
-    */}}
-{{- define "kubecost.federatedStorage.config" }}
+{{/*
+NOTE: added kubecostModel for backward compatibility
+*/}}
+{{- define "kubecost.federatedStorage.config-b64" }}
   {{- if (.Values.kubecostModel).federatedStorageConfig -}}
-    {{- toYaml (.Values.kubecostModel).federatedStorageConfig -}}
+    {{- (.Values.kubecostModel).federatedStorageConfig | b64enc -}}
   {{- else if (.Values.federatedStorage).config -}}
-    {{- toYaml (.Values.federatedStorage).config -}}
-  {{- else if (.Values.global.federatedStorage).federatedStorageConfig -}}
-    {{- toYaml (.Values.global.federatedStorage).federatedStorageConfig -}}
+    {{- (.Values.federatedStorage).config | b64enc -}}
+  {{- else if (.Values.global.federatedStorage).config -}}
+    {{- (.Values.global.federatedStorage).config | b64enc -}}
   {{- else -}}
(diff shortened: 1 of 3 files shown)


kubecost/templates/_helpers.tpl
```diff
@@ -548,21 +548,24 @@ federated storage config helpers
...
 
+    {{/*
+    NOTE: added kubecostModel for backward compatibility
+    */}}
 {{- define "kubecost.federatedStorage.config" }}
   {{- if (.Values.kubecostModel).federatedStorageConfig -}}
-    {{- (.Values.kubecostModel).federatedStorageConfig -}}
-  {{- else if (.Values.federatedStorage).config }}
-    {{- (.Values.federatedStorage).config -}}
+    {{- toYaml (.Values.kubecostModel).federatedStorageConfig -}}
```
thomasvn: Are you sure here? In your PR description you say you test by doing the following:

```yaml
kubecostModel:
  federatedStorageConfig:
    type: S3
    config:
      bucket: xxx
```

But I thought that historically this storage config was passed as a string, not an object? https://github.com/kubecost/kubecost/blob/33a39985a5c6641d339a29f177440efaaf2b795c/cost-analyzer/values.yaml#L633-L670

kubecost/templates/_helpers.tpl
```diff
@@ -548,21 +548,24 @@ federated storage config helpers
...
+  {{- else if (.Values.federatedStorage).config -}}
+    {{- toYaml (.Values.federatedStorage).config -}}
```
thomasvn: Ditto here. It looks like our current values file is expecting `.Values.federatedStorage.config` to be a string?

https://github.com/kubecost/kubecost/blob/87ca5777d1e494e8244c061b7c63e5e7c64d1b76/kubecost/values.yaml#L535

thomasvn:
Potentially out of scope for this PR. It seems the current `values.yaml` in the `develop` branch has multiple ways to set a storage config. This is kind of confusing, can we prioritize just one primary method?

- `.Values.global.federatedStorage`
- `.Values.finops-agent.federatedStorage`
- `.Values.federatedStorage`

Can we fix the `values.yaml` and templates to prioritize referencing just one of these? If it's too much work for this PR maybe we file a task in Jira for this?

jessegoodier:
> Potentially out of scope for this PR. It seems the current `values.yaml` in the `develop` branch has multiple ways to set a storage config. This is kind of confusing, can we prioritize just one primary method?
> 
> * `.Values.global.federatedStorage`
> * `.Values.finops-agent.federatedStorage`
> * `.Values.federatedStorage`
> 
> Can we fix the `values.yaml` and templates to prioritize referencing just one of these? If it's too much work for this PR maybe we file a task in Jira for this?

100% this. I think we need to figure out what the agent config should be and then narrow down as much as we can.

kubecost/templates/_helpers.tpl
jessegoodier: you are right! 
okay, fixed everything I think.

jessegoodier:
updated description with the new tests. 

We will have to narrow down the values when we integrate the agent and single cluster mode.

kubecost/templates/_helpers.tpl
```diff
@@ -548,21 +548,24 @@ federated storage config helpers
   {{- end }}
 {{- end -}}
 
-{{- define "kubecost.federatedStorage.config" }}
+{{/*
+NOTE: added kubecostModel for backward compatibility
+*/}}
+{{- define "kubecost.federatedStorage.config-b64" }}
```
jessegoodier, line 554: I'm not married to this change with including the type in the name. But it does make it obvious that it should be opaque.

kubecost/templates/_helpers.tpl
```diff
@@ -548,21 +548,24 @@ federated storage config helpers
...
-  {{- else if (.Values.federatedStorage).config }}
-    {{- (.Values.federatedStorage).config -}}
-  {{- else if (.Values.global.federatedStorage).federatedStorageConfig -}}
-    {{- (.Values.global.federatedStorage).federatedStorageConfig -}}
-  {{ else }}
+    {{- (.Values.kubecostModel).federatedStorageConfig | b64enc -}}
+  {{- else if (.Values.federatedStorage).config -}}
+    {{- (.Values.federatedStorage).config | b64enc -}}
+  {{- else if (.Values.global.federatedStorage).config -}}
+    {{- (.Values.global.federatedStorage).config | b64enc -}}
```
thomasvn, line 560: Did you update this from `federatedStorageConfig` to `config`? I'm fine with the config change, but could you also make sure to do the following:

- Remove `.Values.global.federatedStorage.federatedStorageConfig` from the values file. Make sure to update the comments on it too.
- Remove any existing references to `.Values.global.federatedStorage.federatedStorageConfig`

kubecost/values.yaml
```diff
@@ -12,10 +12,21 @@ global:
...
+    # federatedStorage:
+    #   config: |-
+    #     type: S3
+    #     config:
+    #       bucket: ibm-kubecost-unified-agent-demo-data
+    #       endpoint: s3.amazonaws.com
+    #       region: us-west-2
+    #       access_key: xxx
+    #       secret_key: xxx
+    # see federatedStorage: below for more examples
```
thomasvn: Small comment update, then this should be good to go.

```suggestion
    # A federated storage config can be provided via an existingSecretName or via a string in helm values.
    # If both are provided, the string in helm values take precedence.
    # Example:
    #   config: |-
    #     type: S3
    #     config:
    #       bucket: ibm-kubecost-unified-agent-demo-data
    #       endpoint: s3.amazonaws.com
    #       region: us-west-2
    #       access_key: xxx
    #       secret_key: xxx
```

thomasvn (approved):

PR #4186, opened 2025-08-12
TODO: fix comments on federated store

JesseGoodier:
https://github.com/kubecost/kubecost/pull/4185

I tested:
```
global:
  federatedStorage:
    config: |-
 ```
creates correctly with the `config`  key uncommented for federatedStorage on the root


jessegoodier:
fixed by #4194

PR #4189, opened 2025-08-13
remove hard-coded aggregator resource limits

JesseGoodier:
reverts hard-coded limits to reduce risks during upgrades to 3.0.

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -953,9 +953,9 @@ aggregator:
     limits: {}
       # memory: ""
       # cpu: ""
-    requests: {}
-      # memory: ""
-      # cpu: ""
+    requests:
+      memory: 3Gi
+      cpu: 100m
   readinessProbe:
     enabled: true
     initialDelaySeconds: 10
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -949,14 +949,13 @@ aggregator:
   aggregatorDbStorage:
     storageClass: ""  # default storage class
     storageRequest: 128Gi
-
   resources:
-    limits:
-      memory: 5Gi
-    requests:
-      cpu: 2000m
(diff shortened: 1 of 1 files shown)


kubecost/values.yaml
```diff
@@ -949,14 +949,13 @@ aggregator:
   aggregatorDbStorage:
     storageClass: ""  # default storage class
     storageRequest: 128Gi
-
   resources:
-    limits:
-      memory: 5Gi
-    requests:
-      cpu: 2000m
-      memory: 5Gi
```
thomasvn: I'm in agreement with reverting the `limits`. I think we should leave the `requests` as is? What do you think?

thomasvn:
@cpetersen5 I believe you were the one to originally set the following.

```yaml
kubecostAggregator:
  resources:
    limits:
      memory: 5Gi
```

Can you explain again why you did this? Did it help in the environments you observed? I'm in agreement with Jesse on this PR that this is probably a configuration that will not work if applied as a default value to a wide range of Kubecost user environments. For large users I feel this would likely OOM their Aggregator.

cpetersen5:
> @cpetersen5 I believe you were the one to originally set the following.
> 
> ```yaml
> kubecostAggregator:
>   resources:
>     limits:
>       memory: 5Gi
> ```
> 
> Can you explain again why you did this? Did it help in the environments you observed? I'm in agreement with Jesse on this PR that this is probably a configuration that will not work if applied as a default value to a wide range of Kubecost user environments. For large users I feel this would likely OOM their Aggregator.

@thomasvn - I'm in agreement that the limits should be removed. We ended up attributing the OOM issues to a bug that was fixed at end of June/beginning of July, if I remember correctly.

cpetersen5 (approved):

PR #4192, opened 2025-08-15
expose if turbonomics is enabled

aligon:
## Release Notes Headline

N/A

## Commentary for Reviewers

Exposes if the turbonomic integration is enabled via `/productConfigs`

## Testing

Used `helm template` with test values.yaml files with turbo enabled, disabled, and missing. Validated `frontend-configmap.yaml > data > nginx.conf` had the correct return for `location /model/productConfigs`


kubecost/templates/frontend/frontend-configmap.yaml
```diff
@@ -505,6 +505,7 @@ data:
                 "carbonEstimatesEnabled": "{{ include "kubecost.carbonEstimates.enabled" . }}",
                 "clusterControllerEnabled": "{{ include "kubecost.clusterController.enabled" . }}",
                 "forecastingEnabled": "{{ include "kubecost.forecasting.enabled" . }}",
+                "turbonomicEnabled": "{{include "kubecost.turbonomic.enabled" .}}",
```
nikovacevic: Only nit is that the spacing differs from the other lines

nikovacevic (approved):

PR #4206, opened 2025-08-25
fix collections cache setting

JesseGoodier:
Add aggregator.collectoins to values.yaml

testing:

```
k exec -it sts/kubecost-aggregator -- env |ag cache
COLLECTIONS_MEMORY_CACHE_ENABLED=true
```

and false works as expected.


thomasvn (approved):
Thanks for cleaning up!

PR #4214, opened 2025-08-28
[KCM-4506] Cluster controller: Use federated store as default

biancaburtoiu:
## Release Notes Headline

Use federated store secret as default if no storage configs defined for cluster controller v2.

## Commentary for Reviewers

Mount the federated storage secret as a volume and use it if no storage configs are defined for cluster controller v2, otherwise use the storage config secret.

## Links to Issues or Tickets
Closes https://apptio.atlassian.net/browse/KCM-4506

## User Impact and Risks

N/A

## Testing

- helm template

## Documentation Updates

N/A

### DIFF
--- a/kubecost/templates/aggregator/aggregator-statefulset.yaml
+++ b/kubecost/templates/aggregator/aggregator-statefulset.yaml
@@ -229,7 +229,7 @@ spec:
         - name: actions-storage-config
           secret:
             defaultMode: 420
-            secretName: {{ ((.Values.kubecostProductConfigs).actions).storageConfigSecret | default "actions-store" }}
+            secretName: {{ include "kubecost.actions.secretName" . }}
           {{- end }}
         {{- end }}
       {{- end }}
--- a/kubecost/templates/cluster-controller/cluster-controller-actions-bucket.yaml
+++ b/kubecost/templates/cluster-controller/cluster-controller-actions-bucket.yaml
@@ -1,5 +1,5 @@
 {{- if (.Values.clusterController).enabled }}
-{{- if not (.Values.clusterController).storageConfigSecret }}
+{{- if (.Values.clusterController).storageConfig }}
 apiVersion: v1
 kind: Secret
 type: Opaque
--- a/kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
+++ b/kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
@@ -78,6 +78,18 @@ spec:
         volumeMounts:
         - name: cluster-controller-keys
           mountPath: /var/keys
+        {{- if (not (.Values.clusterController).legacyMode) }}
+        {{- if and (not (.Values.clusterController).storageConfigSecret) (not (.Values.clusterController).storageConfig) }}
+        - name: federated-storage-config
+          mountPath: /var/configs/actions-store.yaml
+          subPath: {{ include "kubecost.federatedStorage.fileName" . }}
+          readOnly: true
+        {{- else }}
+        - name: actions-bucket-config
+          mountPath: /var/configs
+          readOnly: true
+        {{- end }}
+        {{- end }}
         env:
         - name: POD_NAME
           valueFrom:
@@ -136,12 +148,6 @@ spec:
         ports:
         - name: http-server
           containerPort: 9731
-        {{- if (not (.Values.clusterController).legacyMode) }}
-        volumeMounts:
-        - name: actions-bucket-config
-          mountPath: /var/configs
-          readOnly: true
-        {{- end }}
         resources:
           {{- toYaml .Values.clusterController.resources | nindent 12 }}
       {{- if .Values.imagePullSecrets }}
@@ -174,8 +180,15 @@ spec:
           # and their validity.
           optional: true
       {{- if not (.Values.clusterController).legacyMode }}
+      {{- if and (not (.Values.clusterController).storageConfigSecret) (not (.Values.clusterController).storageConfig) }}
+      - name: federated-storage-config
+        secret:
+          defaultMode: 420
+          secretName: {{ include "kubecost.federatedStorage.secretName" . }}
+      {{- else }}
       - name: actions-bucket-config
         secret:
           secretName: {{ include "kubecost.clusterController.actionsBucketConfigSecretName" . }}
+      {{- end }}
       {{- end }}
 {{- end }}
--- a/kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
+++ b/kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
@@ -80,6 +80,8 @@ spec:
           mountPath: /var/keys
         {{- if (not (.Values.clusterController).legacyMode) }}
         {{- if and (not (.Values.clusterController).storageConfigSecret) (not (.Values.clusterController).storageConfig) }}
+        - name: config-storage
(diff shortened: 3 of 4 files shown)


kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -229,7 +229,7 @@ spec:
         - name: actions-storage-config
           secret:
             defaultMode: 420
-            secretName: {{ ((.Values.kubecostProductConfigs).actions).storageConfigSecret | default "actions-store" }}
+            secretName: {{ include "kubecost.actions.secretName" . }}
```
biancaburtoiu, line 232: Updated to match the definition in L222, the same secretName should be referenced in both cases

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -136,12 +148,6 @@ spec:
         ports:
         - name: http-server
           containerPort: 9731
-        {{- if (not (.Values.clusterController).legacyMode) }}
-        volumeMounts:
-        - name: actions-bucket-config
-          mountPath: /var/configs
-          readOnly: true
-        {{- end }}
```
biancaburtoiu, line 144: This was a duplicate `volumeMounts` definition, I've moved this in the existing `volumeMounts` block above

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -174,8 +180,15 @@ spec:
           # and their validity.
           optional: true
       {{- if not (.Values.clusterController).legacyMode }}
+      {{- if and (not (.Values.clusterController).storageConfigSecret) (not (.Values.clusterController).storageConfig) }}
+      - name: federated-storage-config
+        secret:
+          defaultMode: 420
+          secretName: {{ include "kubecost.federatedStorage.secretName" . }}
+      {{- else }}
```
biancaburtoiu, line 188: We mount the federated storage Secret as a volume if neither of the cluster controller-specific `storageConfigSecret` and `storageConfig` are defined.

kubecost/values.yaml
```diff
@@ -1227,18 +1227,6 @@ clusterController:
   #         "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
   #         "client_x509_cert_url": ""
   #       }
-  # # Actions can reference configuration at a global or cluster-specific level
-  # config:
```
biancaburtoiu, line 1231: This config block shouldn't be needed in the cluster controller values IIRC?

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -78,6 +78,20 @@ spec:
...
+        - name: federated-storage-config
+          mountPath: /var/configs/actions-store.yaml
+          subPath: {{ include "kubecost.federatedStorage.fileName" . }}
+          readOnly: true
+        {{- else }}
+        - name: actions-bucket-config
+          mountPath: /var/configs
+          readOnly: true
+        {{- end }}
+        {{- end }}
```
biancaburtoiu: We mount the federated storage Secret as a volume if neither of the cluster controller-specific `storageConfigSecret` and `storageConfig` are defined. 

I believe we need the base directory `/var/configs` available if we want to mount the federated secret at `/var/configs/actions-store.yaml` (`/var/configs` isn't defined anywhere else in the cluster controller context if I'm not mistaken) so I added an empty volume `config-storage` (defined in `volumes` below) at that path.

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -174,8 +182,17 @@ spec:
           # and their validity.
           optional: true
       {{- if not (.Values.clusterController).legacyMode }}
+      {{- if and (not (.Values.clusterController).storageConfigSecret) (not (.Values.clusterController).storageConfig) }}
+      - name: config-storage
+        emptyDir: { }
+      - name: federated-storage-config
+        secret:
+          defaultMode: 420
+          secretName: {{ include "kubecost.federatedStorage.secretName" . }}
```
biancaburtoiu, line 191: We assume that this secret will exist; we assume that the cluster controller is installed alongside aggregator and so the `kubecost.federatedStorage.secretName` (ie. `.values.global.federatedStorage.existingSecret` or the config to create the secret) will already have been defined.

kubecost/templates/cluster-controller/cluster-controller-actions-bucket.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if (.Values.clusterController).enabled }}
-{{- if not (.Values.clusterController).storageConfigSecret }}
+{{- if (.Values.clusterController).storageConfig }}
```
biancaburtoiu, line 2: Updated this to only create the Secret if the `storageConfig` exists; on the previous condition, if neither `storageConfigSecret` **nor** `storageConfig` were to exist, this would still create the Secret but with nil contents

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -105,7 +119,7 @@ spec:
         - name: LOG_LEVEL
           value: {{ .Values.clusterController.logLevel | default "info" }}
         - name: ACTIONS_BUCKET_CONFIG
-          value: /var/configs/actions-store.yaml
+          value: /var/configs/actions/storage/actions-store.yaml
```
biancaburtoiu, line 122: Updated this path for consistency with the aggregator filesystem

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -78,6 +78,20 @@ spec:
...
         - name: cluster-controller-keys
           mountPath: /var/keys
+        {{- if not (.Values.clusterController).legacyMode }}
+        {{- if and (not (.Values.clusterController).storageConfigSecret) (not (.Values.clusterController).storageConfig) }}
+        - name: actions-storage
+          mountPath: /var/configs/actions/storage
+        - name: federated-storage-config
+          mountPath: /var/configs/actions/storage/actions-store.yaml
+          subPath: {{ include "kubecost.federatedStorage.fileName" . }}
+          readOnly: true
```
ameijer, line 88: love it

kubecost/values.yaml
ameijer: yes, you are right, since these go into aggregator. can you make sure they are defined in kubecostProductCOnfigs then? with all the other actions stuff?

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
ameijer: yes, safe assumption, and ok to explode if this is not the case

ameijer (approved):
LGTM

kubecost/values.yaml
biancaburtoiu: correct, this lives in kubecostProductConfigs, see https://github.com/kubecost/kubecost/blob/8c72e3774bf864bacbcbb37bc20e8c7e5d731f88/kubecost/values.yaml#L1298-L1365

JesseGoodier (approved):
thank you!

PR #4216, opened 2025-08-28
consistent service account templates

JesseGoodier:
I have done basic testing that the aggregator rolebinding works and service accounts I expect to either be created (or not created) are working.

But we need more testing before merging.

Jira [KCM-4382]

[KCM-4382]: https://apptio.atlassian.net/browse/KCM-4382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jessegoodier:
AI (clearly) assisted on this one. but it is a logical layout and allows all pods to have dedicated service accounts.

I'm not married to this, but we should consider supporting a dedicated SA per pod.

biancaburtoiu:
leaving a note for future reference: may need to revisit the two new files I've added in https://github.com/kubecost/kubecost/pull/4225 as to whether they should use `template` or `include` on the `kubecost.aggregator.serviceAccountName` references
- `kubecost/templates/aggregator/aggregator-role.yaml`
- `kubecost/templates/aggregator/aggregator-role-binding.yaml`

kubecost/templates/_helpers.tpl
```diff
@@ -195,6 +195,85 @@ Create the name of the service account
...
+{{- end -}}
+
+{{/*
+Check if component service account should be created
+Usage: {{ include "kubecost.componentServiceAccountCreate" (dict "component" "aggregator" "context" .) }}
+*/}}
+{{- define "kubecost.componentServiceAccountCreate" -}}
+{{- $component := .component -}}
+{{- $context := .context -}}
+{{- $componentConfig := index $context.Values.serviceAccount.components $component -}}
```
Sean-Holcomb, line 222: These functions could just take a "SAConfig" as an argument rather than relying on all the SAConfigs being in a centralized list

Sean-Holcomb (commented):
I agree with the premise of this PR however I am unsure about the centralization of the SA configs as it works against the modularity of the components. For example if cluster-controller were to move to a sub chart. These changes would make that more difficult.

PR #4217, opened 2025-08-29
example configs

JesseGoodier:
providing basic configs in-repo

### DIFF
--- a/examples/agentOnly/helmValues-kubecost-aws.yaml
+++ b/examples/agentOnly/helmValues-kubecost-aws.yaml
@@ -1,0 +1,28 @@
+global:
+  ## clusterId is used to identify the cluster in the FinOps Agent
+  ## when using the same cluster names in different regions, we suggest appending the region to the clusterId
+  clusterId: globally-unique-cluster-id
+  federatedStorage:
+    ## Use the existingSecret when a access key and secret is used to access the federated storage.
+    ## the 'config' can be used when using assumed roles.
+    existingSecret: ""
+    config: |-
+      type: S3
+      config:
+        bucket: kubecost-federated-storage
+        endpoint: s3.amazonaws.com
+        region: us-west-2
+clusterController:
+  enabled: true  # optional, automated rightsizing of containers
+networkCosts:
+  enabled: true  # optional, daemonset improves cost allocation the the correct pod
+aggregator:
+  enabled: false
+frontend:
+  enabled: false
+cloudCost:
+  enabled: false
+forecasting:
+  enabled: false
+finops-agent:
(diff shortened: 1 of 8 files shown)


examples/agentOnly/helmValues-kubecost-aws.yaml
```diff
@@ -0,0 +1,28 @@
+global:
+  ## clusterId is used to identify the cluster in the FinOps Agent
+  ## when using the same cluster names in different regions, we suggest appending the region to the clusterId
+  clusterId: globally-unique-cluster-id
+  federatedStorage:
+    ## Use the existingSecret when a access key and secret is used to access the federated storage.
+    ## the 'config' can be used when using assumed roles.
```
thomasvn: I think the comment about assumed roles can lead to user confusion. It may be easier just to keep it simple like this.

```suggestion
    # A federated storage config can be provided via an existingSecret or via a string in helm values.
    # If both are provided, the string in helm values take precedence.
```

examples/federatedStorage/azure/federated-store.yaml
```diff
@@ -0,0 +1,9 @@
+type: AZURE
+config:
+  storage_account: "kubecost-multi-cluster-storage"
+  container: ""
+  ## keys are not required when using assumed roles
+  storage_account_key: ""
+  ## endpoint is only needed if primary blob service endpoint domain is not blob.core.windows.net
+  # endpoint: "blob.core.chinacloudapi.cn"
+  # max_retries: 0
```
thomasvn: Remove this section? It's probably not relevant for a quickstart example? They can refer to the full docs if they need this config.

```suggestion
```

examples/federatedStorage/azure/federated-store.yaml
```diff
@@ -0,0 +1,9 @@
+type: AZURE
+config:
+  storage_account: "kubecost-multi-cluster-storage"
```
thomasvn:
```suggestion
  storage_account: kubecost-multi-cluster-storage
```

Minor detail. All your other federated-store.yaml configs don't have quotes here

thomasvn:
Looks good to me, only minor feedback on the docs.

My main concern is that these examples/docs may quickly go out of date. We will either need to continuously maintain them, or eventually delete them to reduce any user confusion here. For now, I'm good to merge.

thomasvn (approved):

jessegoodier:
thanks for the suggested edits- agree on all points.

I believe many users will benefit from this small set of common configurations and hope that outweighs any duplication in work.

happy to delete this folder if we find a better solution.

PR #4225, opened 2025-08-29
[KCM-4382] Log collection: Set up Role/RoleBinding for aggregator SA

biancaburtoiu:
## Release Notes Headline

N/A

## Commentary for Reviewers

We need the service account linked with the `aggregator` to be able to read pod logs for bug report capture. This PR:
- sets up "log collection" Role with `pods/log` read rule 
- sets up RoleBinding to associate log collection Role with the kubecost aggregator service account

## Links to Issues or Tickets

Part of https://apptio.atlassian.net/browse/KCM-4382

## User Impact and Risks

N/A

## Testing

N/A

## Documentation Updates

N/A

### DIFF
--- a/kubecost/templates/aggregator/aggregator-role-binding.yaml
+++ b/kubecost/templates/aggregator/aggregator-role-binding.yaml
@@ -14,5 +14,4 @@ subjects:
   - kind: ServiceAccount
     name: {{ template "kubecost.aggregator.serviceAccountName" . }}
     namespace: {{ .Release.Namespace }}
----
 {{- end }}
--- a/kubecost/templates/aggregator/aggregator-role-binding.yaml
+++ b/kubecost/templates/aggregator/aggregator-role-binding.yaml
@@ -1,0 +1,18 @@
+{{- if (.Values.reporting).logCollection }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ template "kubecost.aggregator.serviceAccountName" . }}-log-collection
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kubecost.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "kubecost.aggregator.serviceAccountName" . }}-log-collection
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "kubecost.aggregator.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+---
+{{- end }}
--- a/kubecost/templates/aggregator/aggregator-role.yaml
+++ b/kubecost/templates/aggregator/aggregator-role.yaml
@@ -1,0 +1,19 @@
+{{- if (.Values.reporting).logCollection -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ .Release.Namespace }}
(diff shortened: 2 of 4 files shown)


kubecost/templates/kubecost-role-binding-template.yaml
biancaburtoiu, line 1: The Role and RoleBinding were defined sometime in the past, specifically for log collection, but tied to the general "kubecost.serviceAccountName". Since we want this log collection related piece to exist in the aggregator, i've moved these definitions in the aggregator templates and made them point to the aggregator service account "kubecost.aggregator.serviceAccount".

kubecost/templates/kubecost-role-template.yaml
```diff
@@ -1,31 +1,3 @@
...
-    - watch
-- apiGroups:
-    - ''
-  resources:
-    - configmaps
-  verbs:
-    - get
-    - list
-    - watch
-    - update
```
biancaburtoiu, line 26: I omitted this rule from the new Role definition I've set up in `aggregator-role.yaml`. I don't see how we would need it for log collection... could anyone help me double-check this?

kubecost/templates/kubecost-role-binding-template.yaml
biancaburtoiu: Another consequence of making this change is I effectively revoke the pod logs read permissions from the general "kubecost.serviceAccountName" SA. I'm saying this because the `logCollection` value in the helm chart was default true, so this Role (and RoleBinding) were created more often than not. 

Now, only the aggregator SA would have them. Is that acceptable?

kubecost/templates/aggregator/aggregator-role.yaml
```diff
@@ -0,0 +1,18 @@
...
+rules:
+  - apiGroups:
+      - ''
+    resources:
+      - "pods/log"
+    verbs:
+      - get
+      - list
+      - watch
+{{- end }}
```
biancaburtoiu, line 18: Would also appreciate feedback on the filenames used for these two new files

kubecost/templates/aggregator/aggregator-role-binding.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    {{- include "kubecost.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "kubecost.aggregator.serviceAccountName" . }}-log-collection
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "kubecost.aggregator.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
```
biancaburtoiu, line 16: @jessegoodier As long as the template definition for "kubecost.aggregator.serviceAccountName" stays valid, I believe these changes should not clash in any way with your work in https://github.com/kubecost/kubecost/pull/4216

JesseGoodier (approved):

kubecost/templates/kubecost-role-binding-template.yaml
jessegoodier: yes, this is good. we can narrow down when needed in the future.

kubecost/templates/aggregator/aggregator-role.yaml
jessegoodier: as in aggregator-role.yaml? works for me.

kubecost/templates/kubecost-role-template.yaml
jessegoodier: agree

JesseGoodier (approved):

PR #4246, opened 2025-09-10
Fix Athena config from values with proper authorizer types

Sean-Holcomb:
## Commentary for Reviewers

Authorizer types for Athena Cloud Cost config were incorrect. Updated with the correct values


thomasvn (approved):
Tentatively approving because this is an update in the error message.

@Sean-Holcomb have we tested this `cloud-integration.json` schema? I don't see it documented yet in the public documentation the error message links to. https://www.ibm.com/docs/en/kubecost/self-hosted/2.x?topic=installation-cloud-billing-integrations

PR #4247, opened 2025-09-10
Add Backwards Compatible Cloud-Integration Secret Name Logic

HMetcalfeW:
## Testing

<todo: Hunter still to test>


thomasvn:
I believe this PR is safe to close. PR https://github.com/kubecost/kubecost/pull/4290 throws an error if `.Values.kubecostProductConfigs.cloudIntegrationSecret` is set.

PR #4258, opened 2025-09-14
Cleanup `values.yaml` for 3.0 release

thomasvn:
## Release Notes Headline

Cleanup `values.yaml` for 3.0 release

## Commentary for Reviewers

- Explicitly sets default values or empty, instead of leaving the config commented out.
- Removes `.Values.upgrade.tov3`. This was used as a precaution in 2.0. However, because 3.0 has a new install/upgrade command, we can be more certain that users are explicitly wanting to move up to 3.0.
  - Previously `helm install kubecost --repo https://kubecost.github.io/cost-analyzer/ cost-analyzer`
  - Now `helm install kubecost --repo https://kubecost.github.io/kubecost/ kubecost`
- Removes legacy config `.Values.imageVersion`
- Fixes frontend endpoints `/model/hideOrphanedResources` and `/model/hideDiagnostics`. Deduplicates multiple configs which did this.
- Removes `.Values.kubecostProductConfigs.metricsConfigs`. The primary metric emitter is now the agent.
- Removes nearly all cloud configurations under `.Values.kubecostProductConfigs`
- Removes `.Values.kubecostProductConfigs.labelMappingConfigs`

## Links to Issues or Tickets

N/A

## User Impact and Risks

N/A

## Testing

Checked the difference before and after this PR's changes. Validated that the only difference was in [this configmap](https://github.com/kubecost/kubecost/blob/f71a93b038bfd4e5059da3af9cb3b610e4860d13/kubecost/templates/aggregator/aggregator-values-configmap.yaml#L10) which prints all Helm values to a JSON file. [before.yaml.txt](https://github.com/user-attachments/files/22317632/before.yaml.txt). [after.yaml.txt](https://github.com/user-attachments/files/22317633/after.yaml.txt). [diff.txt](https://github.com/user-attachments/files/22317634/diff.txt).

```sh
$ helm template ./kubecost > after.yaml
$ git checkout develop
$ helm template ./kubecost > before.yaml
```

## Documentation Updates

N/A


kubecost/templates/NOTES.txt
```diff
@@ -11,7 +11,7 @@
 
 Kubecost {{ .Chart.Version }} has been successfully installed.
 
-Kubecost 3.x is a major upgrade from previous versions and includes major new features including a brand new API Backend. Please review the following documentation to ensure a smooth transition: TODO
+Kubecost 3.x is a major upgrade from previous versions and includes major new features including a brand new API Backend.
```
chipzoller: Maybe also state breaking changes and point users to read release notes.

chipzoller (approved):

kubecost/values-openshift-cluster-prometheus.yaml
```diff
@@ -13,21 +13,3 @@ serviceMonitor:
   enabled: true
   aggregatorMetrics:
     enabled: false
-
```
mittal-ishaan, line 16: Any reason to remove. I think this is still needed

kubecost/templates/NOTES.txt
thomasvn: Ok! I went ahead and added a sentence referencing the release notes https://github.com/kubecost/kubecost/releases.

I'm going to go ahead and merge, but we can tweak the wording later this week if needed!

PR #4271, opened 2025-09-16
[KCM-4572] Forecasting: use /model/forecast instead of /forecasting, update path

biancaburtoiu:
## Release Notes Headline

N/A

## Commentary for Reviewers

Bug fix 

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-4572

## User Impact and Risks

N/A

## Testing

- installed chart in local kind cluster, port forwarded 9090 and validated that the /model/forecast paths were available

## Documentation Updates

N/A

### DIFF
--- a/kubecost/templates/frontend/frontend-configmap.yaml
+++ b/kubecost/templates/frontend/frontend-configmap.yaml
@@ -475,10 +475,10 @@ data:
         }
 
         {{- if .Values.forecasting.enabled }}
-        location /forecasting {
+        location /model/forecast/ {
             default_type 'application/json';
             proxy_read_timeout          300;
-            proxy_pass http://forecasting/;
+            proxy_pass http://forecasting/forecast/;
             proxy_redirect off;
             proxy_set_header Connection "";
             proxy_set_header  X-Real-IP  $remote_addr;


kubecost/templates/frontend/frontend-configmap.yaml
biancaburtoiu, line 1: We want the routes called by the UI:
- /model/forecast/allocation?aggregate=namespace&predictionWindow=30d
- /model/forecast/assets?aggregate=type&predictionWindow=30d&filter%5Bjoin%5D=%2B

to point to the following routes supported by the `forecasting` Service (as per https://github.com/kubecost/kubecost-modeling):
- /forecast/allocation 
- /forecast/assets 

The configuration below should support that

biancaburtoiu: For context:
- there exists a `forecasting` Deployment in the kubecost install, as well as a Service surfacing port 5000 on the pods controlled by the Deployment
- `proxy_pass` points to the `forecasting` upstream server, which is defined further above in the nginx config to direct to the server address of the forecasting Service, on port 5000

nik-kc: Makes total sense, thanks for the context!

nik-kc (approved):

PR #4275, opened 2025-09-16
[KCM-4575] Network costs: Add ClusterRole/ClusterRoleBinding

biancaburtoiu:
## Release Notes Headline

N/A

## Commentary for Reviewers

Bug fix: add `ClusterRole` / `ClusterRoleBinding` to allow necessary permissions required by the network costs daemonset. 

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-4575

## User Impact and Risks

N/A

## Testing

- helm template

## Documentation Updates

N/A


kubecost/templates/kubecost-network-costs-role-template.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: {{ template "kubecost.serviceAccountName" . }}-network-costs-cluster-role
+  labels:
+    {{- include "kubecost.commonLabels" . | nindent 4 }}
+rules:
+  - apiGroups:
+      - ''
+    resources:
+      - nodes
+      - pods
+      - services
```
biancaburtoiu: As per k8s docs (https://kubernetes.io/docs/reference/access-authn-authz/rbac/):
> A Role always sets permissions within a particular [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces); when you create a Role, you have to specify the namespace it belongs in.
ClusterRole, by contrast, is a non-namespaced resource. The resources have different names (Role and ClusterRole) because a Kubernetes object always has to be either namespaced or not namespaced; it can't be both.

`nodes` are cluster-scoped resources, so they need a ClusterRole instead of a simple Role. `pods` and `services`, however, are namespace-scoped, so we could use a separate Role for them. 
I decided to put them all in the same, single ClusterRole as I assumed that the network costs DaemonSet will need to look at pods and services across the entire cluster. Also, in the past (before [this commit](https://github.com/kubecost/kubecost/commit/c051df2b34eb2b38ec7ef44f76cbb7d7bdbf1a60)) we used to a have a `ClusterRole` with all permissions tied in it, which is why I decided to put everything in a single ClusterRole. 

If this needs to be split, please let me know

kubecost/templates/kubecost-network-costs-role-template.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if (.Values.networkCosts).enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
biancaburtoiu: This ClusterRole is **only created if network costs are enabled**

kubecost/templates/kubecost-network-costs-cluster-role-binding-template.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+    {{- include "kubecost.commonLabels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "kubecost.serviceAccountName" . }}-network-costs-cluster-role
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "kubecost.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
```
biancaburtoiu: This ClusterRoleBinding ties the Role to the `"kubecost.serviceAccountName"`. This service account has a number of other usages:
- in the frontend Deployment
- in the aggregator StatefulSet, in the absence of an aggregator-specific SA
- in the cloud cost Deployment, in the absence of a cloud cost-specific SA

We may want to consider the idea of setting up a separate, network costs daemonset-specific SA? In the "worst case scenario", where network costs are enabled and this service account ends up being used for the other workloads, we would end up granting **read permissions on the pods/services/nodes resources** to all these workloads. Is that a big concern?

chipzoller: Yes, this should have its own SA as per Kubernetes best practices.

chipzoller:
I haven't read through this thoroughly, but what changed here to required net new resources be created?

kubecost/templates/kubecost-network-costs-cluster-role-binding-template.yaml
thomasvn: Thanks for digging into `"kubecost.serviceAccountName"`! I agree that the network costs daemonset should have its own SA.

thomasvn: I think Jesse would agree too, because he started a draft PR to separate out all ServiceAccounts https://github.com/kubecost/kubecost/pull/4216

thomasvn:
> I haven't read through this thoroughly, but what changed here to required net new resources be created?

@chipzoller I think it's because in v2.8 we were always creating a `ClusterRole` with sufficient permissions for all our pods ([ref](https://github.com/kubecost/kubecost/blob/v2.8/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml)). In v3.0 we do not create a ClusterRole unless you have the ClusterController enabled ([ref](https://github.com/search?q=repo%3Akubecost%2Fkubecost%20ClusterRole&type=code)).

kubecost/templates/network-costs/_helpers.tpl
```diff
@@ -35,3 +35,11 @@ app: {{ template "kubecost.networkCosts.name" . }}
...
 app: {{ template "kubecost.networkCosts.name" . }}
 {{- end }}
+
+{{- define "kubecost.networkCosts.serviceAccountName" -}}
+{{- if .Values.networkCosts.serviceAccountName -}}
+    {{ .Values.networkCosts.serviceAccountName }}
+{{- else -}}
+    {{ template "kubecost.serviceAccountName" . }}
+{{- end -}}
+{{- end -}}
```
biancaburtoiu: This allows users to specify a `.values.networkCosts.serviceAccountName` name of a SA to be used specifically for the network costs daemonset. If that is not specified, we fall back to using the default `"kubecost.serviceAccountName"`

biancaburtoiu: Can I confirm that the intended behavior upon defining `.values.networkCosts.serviceAccountName` is to use an _existing_ SA previously created by the user? That's what I thought, but I found [this file](https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/cloud-cost/cloud-cost-service-account.yaml) and got confused -- for cloud costs, it seems that we _create_ a SA if create is true _and_ the SA name is supplied?

biancaburtoiu: Alternatively, do we want to _always_ create a new, separate SA, specific for network costs, as we do for the cluster controller? https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/cluster-controller/cluster-controller-service-account.yaml

chipzoller: Yes, we should not be recycling ServiceAccounts between separate components.

kubecost/templates/kubecost-network-costs-cluster-role-binding-template.yaml
thomasvn, line 1: Can we move this new template into the `network-costs` dir?

kubecost/templates/kubecost-network-costs-cluster-role-template.yaml
thomasvn, line 1: Can we move this new template into the `network-costs` dir?

kubecost/templates/network-costs/_helpers.tpl
thomasvn: @biancaburtoiu I think that we should follow the patterns established by cluster controller!

- Always create a new and separate SA specific for network costs
- Generate a default name for that SA https://github.com/kubecost/kubecost/blob/8ea7f9581a8274281b18403c64d7bb694179c242/kubecost/templates/cluster-controller/_helpers.tpl#L25-L27
- No need to offer the config `.Values.networkCosts.serviceAccountName`.

kubecost/templates/kubecost-network-costs-cluster-role-binding-template.yaml
biancaburtoiu: Done

kubecost/templates/kubecost-network-costs-cluster-role-template.yaml
biancaburtoiu: Done

kubecost/templates/network-costs/_helpers.tpl
biancaburtoiu: Thank you for your guidance @thomasvn, much appreciated - i've applied all your feedback in the latest commits

kubecost/templates/network-costs/network-costs-cluster-role.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    {{- include "kubecost.commonLabels" . | nindent 4 }}
+rules:
+  - apiGroups:
+      - ''
+    resources:
+      - nodes
+      - pods
+      - services
+      - endpoints
```
biancaburtoiu, line 15: I cross-checked these resources against:
- the original ClusterRole config which was known WAI for the network costs daemonset: https://github.com/kubecost/kubecost/blob/v2.8/cost-analyzer/templates/cost-analyzer-cluster-role-template.yaml
- analysis run by Bob against the contents of the https://github.com/kubecost/network-costs-rs repository 

Should any rules be missing from this ClusterRole, please let me know. Otherwise we will need to double check these rules are sufficient in a real install

kubecost/templates/kubecost-network-costs-role-template.yaml
chipzoller: No, you've got it right.

chipzoller (approved):
LGTM

PR #4282, opened 2025-09-18
Remove the `.Values.kubecost` root config.

thomasvn:
## Release Notes Headline

N/A

## Commentary for Reviewers

- The `.Values.kubecost` root config was originally created as a shared config for all pods which deployed the `cost-model` container (aggregator, cloudcost, localstore). Although this abstraction can be understood by Kubecost internally, users are not aware of what pods will be deploying the `cost-model` container and is confusing for them to understand.
- This PR deletes `.Values.kubecost`. Anything previously under this config was moved elsewhere.
- I've validated none of our nightly environments currently use configs under `.Values.kubecost` ([here](https://github.com/kubecost/integration-ci-cd/tree/main/values)).

## Links to Issues or Tickets

N/A

## User Impact and Risks

No risk. These configs have not yet been released.

## Testing

I took a `diff` of before and after my changes with default configurations. [before.yaml.txt](https://github.com/user-attachments/files/22399543/before.yaml.txt). [after.yaml.txt](https://github.com/user-attachments/files/22399545/after.yaml.txt). [diff.txt](https://github.com/user-attachments/files/22399547/diff.txt).

```sh
$ helm template ./kubecost > after.yaml
$ git checkout develop
$ helm template ./kubecost > before.yaml
```

The `diff.txt` only shows changes to the following:

- Configmap which prints out all `values.yaml`
- Checksums
- Updated ImagePullPolicy to `IfNotPresent` as default in multiple places
- New key `"tenMinuteDataRetention": "36",` in `/model/productConfigs`

## Documentation Updates

N/A


kubecost/templates/_helpers.tpl
```diff
@@ -63,10 +63,12 @@ RBAC exclusivity check: make sure either simple RBAC or RBAC Teams is configured
...
 
 {{/*
-Print a warning if PV is enabled AND EKS is detected AND the EBS-CSI driver is not installed. This can be disabled is the installer's RBAC does not allow kube-system access
+Print a warning if PV is enabled AND EKS is detected AND the EBS-CSI driver is not installed.
+Skip the check if CI/CD is enabled and skipSanityChecks is set. Argo CD, for
+example, does not support templating a chart which uses the lookup function.
 */}}
 {{- define "kubecost.eksStorage.check" }}
-{{ if (((.Values.kubecost).eksStorage).check).enabled }}
+{{- if not (and .Values.global.platforms.cicd.enabled .Values.global.platforms.cicd.skipSanityChecks) }}
```
thomasvn, line 71: Delete `eksStorage.check.enabled`, because we already have the guard `skipSanityChecks`. This is the existing guard we use in templates before we perform a Helm `lookup()`

kubecost/templates/frontend/frontend-configmap.yaml
```diff
@@ -506,8 +506,9 @@ data:
                 "forecastingEnabled": "{{ include "kubecost.forecasting.enabled" . }}",
                 "turbonomicEnabled": "{{ include "kubecost.turbonomic.enabled" . }}",
                 "chartVersion": "DEVELOP_BRANCH",
-                "hourlyDataRetention": "{{ (.Values.kubecost.retention1h) }}",
-                "dailyDataRetention": "{{ (.Values.kubecost.retention1d) }}",
+                "tenMinuteDataRetention": "{{ (.Values.aggregator.retention10m) }}",
```
thomasvn, line 509: Added a new key here, because why not!

ameijer (commented):
please fix integration tests @ kubecost-integration-tests/.github/workflows/build-stack.yaml after you merge this, and kubecost-deploy configs for dev @ kubecost-deploy/argocd/apps/dev/kubecost/values.yaml, this is going to break beta 3.0 deployments

thomasvn:
@ameijer Thank you for that catch! I didn't check those repos. I'll create PRs for both of those repos now, before merging this one.

thomasvn:
@ameijer I've updated `kubecost-integration-tests/.github/workflows/build-stack.yaml` in the linked PR. However I don't see any usage of `.Values.kubecost` in `kubecost-deploy/argocd/apps/dev/kubecost/values.yaml`.

Can you validate? Otherwise is this ready to approve and go in?

ameijer:
I had one question on the integration tests PR. I'll approve to unblock though

ameijer (approved):

Sean-Holcomb (commented):
I know I missed the boat a little on this one a little. The main purpose of this which is removing `.Values.kubecost` is fine since  it purpose was to be a place where shared configs can go with the hope that it could replace `.Values.kubecostProductConfigs`. Which is something which did not happen because of compatibility concerns. However I would say that changing the signatures on the helper functions is by removing the kubecost prefix is less helpful as that servered the purpose of uniquely identifying those helpers from helpers which may exist in child charts and uniquely identifying them for any parent charts. Not a big or pressing issue at the moment and fortunately changing helper signatures (unlike values) is something we can do at anytime with much lower risk of a breaking change.

thomasvn:
@Sean-Holcomb Are you saying that changing the helper function

```
# FROM
{{- define "aggregator.imageRegistry" -}}

# TO
{{- define "kubecost.aggregator.imageRegistry" -}}
```

has implications on subcharts? If so, I didn't know that, that's my bad! Let me know if there's any follow up work here to make sure we get it fixed.

Sean-Holcomb:
It looks like I misread the PR I was under the impression you went from
```
# FROM
{{- define "kubecost.aggregator.imageRegistry" -}}

# TO
{{- define "aggregator.imageRegistry" -}}
```

PR #4301, opened 2025-09-22
Add more detailed failure messages on `.Values.federatedETL`

thomasvn:
## Release Notes Headline

N/A

## Commentary for Reviewers

- We received feedback that the previous error message on `.Values.federatedETL` was vague.
- This adds specific error messages for each case
- This PR also specifically adds `.Values.finops-agent.enabled=true`. By default, it was already evaluating to true ([ref](https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies)).

## Links to Issues or Tickets

N/A

## User Impact and Risks

N/A

## Testing

```yaml
# values-test.yaml
federatedETL:
  federatedCluster: true
```

```sh
$ helm template ./kubecost -f values-test.yaml
Error: execution error at (kubecost/templates/NOTES.txt:2:4): `.Values.federatedETL.federatedCluster` is no longer supported. Please use `.Values.finops-agent.enabled=true` instead.
```

```sh
# Ran the following commands. No diff in defaults.
$ helm template ./kubecost --debug > after.yaml
$ git checkout develop
$ helm template ./kubecost --debug > before.yaml
```

## Documentation Updates

N/A

### DIFF
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -12,13 +12,13 @@ Kubecost 3.0 preconditions
     {{ fail "`.Values.kubecostModel.federatedStorageConfigSecret` is no longer supported. Please use `.Values.global.federatedStorage.existingSecret` instead." }}
   {{- end -}}
   {{- if (.Values.federatedETL).federatedCluster -}}
-    {{ fail "`.Values.federatedETL.federatedCluster` is no longer supported. Please use `.Values.finops-agent.enabled=true` instead." }}
+    {{ fail "`.Values.federatedETL.federatedCluster` is no longer supported. Please use `.Values.finops-agent.enabled=true` instead if you want to push this cluster's metrics to the federated storage." }}
   {{- end -}}
   {{- if (.Values.federatedETL).agentOnly -}}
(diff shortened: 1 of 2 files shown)


HMetcalfeW (approved):
LGTM. One comment, `(.Values.federatedETL)` are you wrapping these values for ease of readability?

thomasvn:
@HMetcalfeW The parenthesis help to make sure we don't dereference a potentially null value!

kubecost/templates/_helpers.tpl
```diff
@@ -11,8 +11,17 @@ Kubecost 3.0 preconditions
   {{- if (.Values.kubecostModel).federatedStorageConfigSecret -}}
     {{ fail "`.Values.kubecostModel.federatedStorageConfigSecret` is no longer supported. Please use `.Values.global.federatedStorage.existingSecret` instead." }}
   {{- end -}}
+  {{- if (.Values.federatedETL).federatedCluster -}}
+    {{ fail "`.Values.federatedETL.federatedCluster` is no longer supported. Please use `.Values.finops-agent.enabled=true` instead." }}
```
thomasvn: Before merging, I may want to update this error message one more time. It needs to be clear that `.Values.finops-agent.enabled=true` only if they set `federatedCluster=true`

PR #4303, opened 2025-09-23
[KCM-3969] Cluster controller: Use port 9732 for non-legacy mode

biancaburtoiu:
## Release Notes Headline
N/A

## Commentary for Reviewers
Update the cluster controller Service to surface port 9732 if not in legacy mode. 

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-3969

## User Impact and Risks

N/A

## Testing

- helm template

## Documentation Updates

N/A

### DIFF
--- a/kubecost/templates/cluster-controller/cluster-controller-service.yaml
+++ b/kubecost/templates/cluster-controller/cluster-controller-service.yaml
@@ -11,8 +11,13 @@ spec:
   ports:
     - name: http
       protocol: TCP
+      {{- if not (.Values.clusterController).legacyMode }}
+      port: 9732
+      targetPort: 9732
+      {{- else }}
       port: 9731
       targetPort: 9731
+      {{- end }}
   selector:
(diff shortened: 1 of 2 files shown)


kubecost/templates/cluster-controller/cluster-controller-service.yaml
```diff
@@ -11,8 +11,13 @@ spec:
   ports:
     - name: http
       protocol: TCP
+      {{- if not (.Values.clusterController).legacyMode }}
+      port: 9732
+      targetPort: 9732
+      {{- else }}
       port: 9731
       targetPort: 9731
+      {{- end }}
```
biancaburtoiu, line 20: I don't think that the Service is used in the normal user flow, but we do want to use it in our integration tests to allow calling the force sync API: https://github.com/kubecost/kubecost-integration-tests/pull/298/files

ameijer (approved):
good looks

PR #4323, opened 2025-09-29
bump aggregator.persistentConfigStorage.storageRequest to 10Gi from 1Gi

cpetersen5:
## Release Notes Headline

 When deploying the Kubecost Helm chart to a cluster using a non-standard storage class, the aggrator fails to come up because we're allocating 1Gi of storage for the AggregatorConfig PVC, but most non-default storage classes have a higher minimum. This results in the aggregator failing to come up, per the logs in this bug ticket: https://apptio.atlassian.net/browse/KCM-4659

## Commentary for Reviewers

 Bumps default Helm values for `.Values.aggregator.persistentConfigStorage.storageRequest` from 1Gi to 10Gi. Note: we use 50Gi in the `kc30rc` namespace. 

## Links to Issues or Tickets

 N/A

## User Impact and Risks

 Limited - this change should help customers running non-standard default storage classs from experiencing deployment issues. 

## Documentation Updates

 N/A

### DIFF
--- a/kubecost/templates/NOTES.txt
+++ b/kubecost/templates/NOTES.txt
@@ -8,6 +8,7 @@
 {{- include "kubecost.caCertsSecretConfig.check" . -}}
 {{- include "kubecost.actionsStorageSourceCheck" . -}}
 {{- include "kubecost.localStoreClusterIdCheck" . -}}
+{{- include "kubecost.aggregatorConfigStorageWarning" . -}}
 ================================================================================
 KUBECOST INSTALLATION SUCCESSFUL
 ================================================================================
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -594,3 +594,12 @@ NOTE: added kubecostModel for backward compatibility
 {{ printf "\n\nWARNING: The clusterId is set to the default value of 'cluster-one'. This is not recommended if you intend to use multi-cluster federation in the future. Please set a globally unique .Values.global.clusterId\n\n" }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+NOTE: Added storage request warning for aggregator config storage
+*/}}
+{{- define "kubecost.aggregatorConfigStorageWarning" -}}
+{{- if and .Values.aggregator.enabled (lt (.Values.aggregator.aggregatorDbStorage.storageRequest | regexFind "[0-9]+" | int) 50) -}}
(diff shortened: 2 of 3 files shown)


thomasvn:
@cpetersen5 Thanks for taking initiative on this one. Are you sure this is the right approach?

The `aggregator.persistentConfigStorage` PV doesn't hold the database, it holds JSON/YAML config files that the Aggregator reads in. 50Gi would almost certainly be overprovisioning. Is there something better that we can do here? For example a useful warning/error message if we detect a higher performance StorageClass is in use?

cpetersen5:
> @cpetersen5 Thanks for taking initiative on this one. Are you sure this is the right approach?
> 
> The `aggregator.persistentConfigStorage` PV doesn't hold the database, it holds JSON/YAML config files that the Aggregator reads in. 50Gi would almost certainly be overprovisioning. Is there something better that we can do here? For example a useful warning/error message if we detect a higher performance StorageClass is in use?

@thomasvn I added a warning if the storage requests for Aggregator's config is less than 50Gi and set the default value in the Helm chart to 10Gi instead of 50Gi, like in the `kc30` namespace. How do you feel about this approach?

@jessegoodier - Do you have a strong opinion on this? I noticed you updated the configuration for the `kc30` namespace, not sure if you have any additional context to add.

cpetersen5:
@thomasvn @jessegoodier - Wanted to circle back on this. Do either of you have any feedback on this approach to handling non-standard storage classes gracefully? cc: @peatey

thomasvn:
Hey @cpetersen5 thanks for continuing to work on this. I really appreciate it.

I'm torn here. It would be great to provide warnings/instructions to users whose PVs are not being provisioned (due to a StorageClass requirement). I'm not sure what the right approach is though.

Increasing `.Values.aggregator.persistentConfigsStorage.storageRequest` to anything greater than 1Gi is definitely overprovisioning. And I am hesitant to overprovision a PV for **all users** as the new default experience. Additionally this error **only occurs** on specialized StorageClasses which _requires_ a minimum amount of storage when creating a PV.

Do you have ideas here on other approaches we could take that are less intrusive? What do you think about closing this PR/ticket for now, and returning to this if this continues to be a problem for many users?

cpetersen5:
> Hey @cpetersen5 thanks for continuing to work on this. I really appreciate it.
> 
> I'm torn here. It would be great to provide warnings/instructions to users whose PVs are not being provisioned (due to a StorageClass requirement). I'm not sure what the right approach is though.
> 
> Increasing `.Values.aggregator.persistentConfigsStorage.storageRequest` to anything greater than 1Gi is definitely overprovisioning. And I am hesitant to overprovision a PV for **all users** as the new default experience. Additionally this error **only occurs** on specialized StorageClasses which _requires_ a minimum amount of storage when creating a PV.
> 
> Do you have ideas here on other approaches we could take that are less intrusive? What do you think about closing this PR/ticket for now, and returning to this if this continues to be a problem for many users?

Thanks, Thomas! I agree it makes sense to close this out. Perhaps we track customers who experience this issue and document the storage classes they're using. If we find most of our customers are having the same issue with a handful of the same storage classes we can add templating around them.

PR #4337, opened 2025-10-07
Refactor Legacy Actions

thomasvn:
## Release Notes Headline

N/A

## Commentary for Reviewers

This removes the special `./crds` directory. It instead turns the CRD into a template.

The issue with the `./crds` directory, is that Helm automatically applies this CRD before templating the rest of the chart ([docs ref](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you)). This caused problems with the OpenShift Operator, and is probably unnecessary given this CRD is now part of a legacy feature.

The only drawback of putting a CRD in a template, is that there will be issues if the CustomResource gets templated before the CustomResourceDefinition.

TODO: Still investigating whether more should be refactored.


thomasvn:
Closing for now. Not immediately needed.

PR #4359, opened 2025-10-14
Readme updates

JesseGoodier:
## Commentary for Reviewers

Adding comments based on user concerns. 

## Testing

For the systemProxy, I tested the env vars are set. We do not have a way to easily test if this works, will need a user with and old-school firewall to confirm.

```yaml
finops-agent:
  extraEnvVars:
  - name: HTTP_PROXY
    value: httpProxyUrl
  - name: http_proxy
    value: httpProxyUrl
  - name: HTTPS_PROXY
    value: httpsProxyUrl
  - name: https_proxy
    value: httpsProxyUrl
  - name: NO_PROXY
    value: noProxy
  - name: no_proxy
    value: noProxy
```

image

## Documentation Updates

NA

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -602,6 +602,15 @@ finops-agent:
     collectorDataSource:
       enabled: true
       scrapeInterval: 30s
+      ## Retention values are for the agent's local store only and do not affect Kubecost history.
+      ## These can be increased, but will require more memory.
+      retention10m: 6
+      retention1h: 3
+      retention1d: 2
+    exporter:
+      ## Emission interval is the interval at which the agent will emit metrics to the object storage.
+      ## This can be increased to reduce the number of object (versions) written to the object storage when versioning is enabled.
+      emissionInterval: 1m
     cloudability:
       ## Cloudability requires an API key to send data to the cloudability platform
       enabled: false
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -949,7 +949,7 @@ aggregator:
     storageClass: ""  # default storage class. NFS is not supported.
     storageRequest: 128Gi
   resources:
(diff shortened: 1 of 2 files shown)


kubecost/values.yaml
```diff
@@ -400,8 +400,25 @@ teams:
 ## Adds the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables to all
 ## containers. Typically used in environments that have firewall rules which
 ## prevent kubecost from accessing cloud provider resources.
+## While this is here to assist for backward compatibility- it is not officially supported. Consider using a service mesh for proxy setup.
```
thomasvn: Are we sure we want to go so far as to say "it is not officially supported"?

kubecost/values.yaml
```diff
@@ -558,52 +575,36 @@ localStore:
...
   agent:
     collectorDataSource:
       enabled: true
       scrapeInterval: 30s
-      networkPort: 3001
-      retention10m: 6
-      retention1h: 3
-      retention1d: 2
-    exporter:
-      emissionInterval: 1m
```
thomasvn, line 575: What do you think about leaving these configs in? I think they provide good context.

kubecost/values.yaml
jessegoodier: can we say it is supported if we have no way to test?

kubecost/values.yaml
jessegoodier: SG, will add back. 

goal was to reduce confusion and only have users change these when there's a reason.

kubecost/values.yaml
jessegoodier: softer wording :)

thomasvn (approved):
Nice! Thanks for the clarifying comments!

PR #4360, opened 2025-10-14
pre-commit formatting

JesseGoodier:


.pre-commit-config.yaml
thomasvn, line 1: Just checking my understanding here, as I haven't used pre-commit hooks before:

- Using pre-commit hooks are optional?
- To participate, you must have the `pre-commit` tool downloaded?
- If downloaded, this config file defines checks to be run for every local `git commit` that you run?

PR #4364, opened 2025-10-15
Correctly handle init Container block in cloud cost deployment template

mittal-ishaan:
## Release Notes Headline
Correctly handle init Container block in cloud cost deployment template

## Links to Issues or Tickets
NA

## Testing
Helm template for different scenarios `helm template . > template.yaml`
1. UpdateCAtrust diabled and plugin diabled
2. only UpdateCAtrust enabled
3. Only plugin enabled
4. Both enabled
All 4 times I got the desired and valid template.yaml


JesseGoodier (approved):

thomasvn (approved):
This looks correct to me. Thanks! Could you update the PR details with the commands/output when testing this?

PR #4371, opened 2025-10-23
fix healthz logs

JesseGoodier:
## Release Notes Headline

disabled healthz logs

The frontend healthz endpoint needs to be outside the saml block.

## User Impact and Risks

less logs, no impact

## Testing

```
k exec -it deployments/saml-okta-frontend -- curl 0:9090/healthz
healthy
```

## Documentation Updates
NA

### DIFF
--- a/kubecost/templates/frontend/frontend-configmap.yaml
+++ b/kubecost/templates/frontend/frontend-configmap.yaml
@@ -124,7 +124,11 @@ data:
         location = /custom_504.html {
             internal;
         }
-
+        location /healthz {
+            add_header 'Content-Type' 'text/plain';
+            return 200 "healthy\n";
+            access_log off;
+        }
 {{- if or .Values.saml.enabled .Values.oidc.enabled }}
 
         add_header Cache-Control "max-age=0";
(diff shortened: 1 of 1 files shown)


kubecost/templates/frontend/frontend-configmap.yaml
```diff
@@ -124,7 +124,11 @@ data:
         location = /custom_504.html {
             internal;
         }
-
+        location /healthz {
+            add_header 'Content-Type' 'text/plain';
+            return 200 "healthy\n";
+            access_log off;
+        }
```
thomasvn, line 132:
```suggestion
        # Needs to be served outside of the auth middleware
        location /healthz {
            add_header 'Content-Type' 'text/plain';
            return 200 "healthy\n";
            access_log off;
        }

```

Recommend keeping the comment. But overall this looks good!

thomasvn (approved):

kubecost/templates/frontend/frontend-configmap.yaml
jessegoodier: We don't need the comment, it is outside the scope of middleware.

jessegoodier:
I thought the comment was in the saml/oidc block. all good.

PR #4381, opened 2025-10-27
remove hyphen from finops-agent helm key

JesseGoodier:
## Commentary for Reviewers

After an internal discussion, we all agreed that the hyphen in finops-agent is going to be problematic and we are early enough where the user impact is minimal.

## Links to Issues or Tickets

[KCM-4816]

## User Impact and Risks

The Kubecost 3.0 helm chart added a new helm section for finops-agent using a `-` in the key. Any user that has defined values using that key will need to change it to `finopsagent:` before upgrading to 3.0.3. 

> note that we intentionally are not using camelCase due to the key being used in k8s objects where uppercase letters are not valid.

## Testing

fail if using old finops-agent key:
```
helm template temp ../kubecost --set global.clusterId="test1" \ 
 --set finops-agent.agent.cloudability.enabled=false
```
```
Error: execution error at (kubecost/templates/NOTES.txt:11:4): 
The helm values for finops-agent have been updated.
Please change the finops-agent: key in your helm values to finopsagent:
```

Correctly passing value to sub chart:
```
helm template temp ../kubecost --set global.clusterId="test1" \
  --set finopsagent.agent.cloudability.enabled=true \
  |grep CLOUDABILITY_EMITTER_ENABLED -A1

```
```
            - name: CLOUDABILITY_EMITTER_ENABLED
              value: "true"
```

The above var is disabled by default:
```
helm template temp ../kubecost --set global.clusterId="test1" \ 
 |grep CLOUDABILITY_EMITTER_ENABLED -A1
```
```
            - name: CLOUDABILITY_EMITTER_ENABLED
              value: "false"
```

install is successful 
```
helm upgrade --install kc30 ../kubecost -n kc30 --create-namespace
Release "kc30" has been upgraded. Happy Helming!
NAME: kc30
LAST DEPLOYED: Tue Oct 28 08:36:29 2025
NAMESPACE: kc30
STATUS: deployed
REVISION: 2
NOTES:
WARNING: The clusterId is set to the default value of 'cluster-one'. This is not recommended if you intend to use multi-cluster federation in the future. Please set a globally unique .Values.global.clusterId

================================================================================
KUBECOST INSTALLATION SUCCESSFUL
```

## Documentation Updates

Need to do a find a replace on docs.

[KCM-4816]: https://apptio.atlassian.net/browse/KCM-4816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -12,4 +12,4 @@ dependencies:
   - name: finops-agent
     version: "v1.0.3"
     repository: https://kubecost.github.io/finops-agent-chart
-    condition: finops-agent.enabled
+    condition: finopsAgent.enabled
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -10,6 +10,7 @@ annotations:
       url: https://www.kubecost.com
 dependencies:
   - name: finops-agent
+    alias: finopsAgent
     version: "v1.0.3"
     repository: https://kubecost.github.io/finops-agent-chart
     condition: finopsAgent.enabled
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -10,7 +10,7 @@ annotations:
       url: https://www.kubecost.com
 dependencies:
   - name: finops-agent
-    alias: finopsAgent
+    alias: finopsagent
     version: "v1.0.3"
     repository: https://kubecost.github.io/finops-agent-chart
-    condition: finopsAgent.enabled
+    condition: finopsagent.enabled
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -404,8 +404,8 @@ teams:
 ## Ref: https://www.oreilly.com/library/view/security-with-go/9781788627917/5ea6a02b-3d96-44b1-ad3c-6ab60fcbbe4f.xhtml
 ##
 ## You must also set the finops-agent systemProxy values independently.
-## Please merge these values below with the finops-agent values for your release.
-# finops-agent:
+## Please merge these values below with the finopsAgent: values for your release.
+# finopsAgent:
 #   extraEnvVars:
 #   - name: HTTP_PROXY
 #     value: httpProxyUrl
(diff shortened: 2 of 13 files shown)


kubecost/values.yaml
```diff
@@ -575,10 +575,10 @@ localStore:
       # This can be used to allow the agent history to be uploaded to cloud storage when upgrading to Enterprise.
       helm.sh/resource-policy: keep
 
-## Below are a subset of values for the finops-agent sub-chart that are most commonly changed.
-## See https://github.com/kubecost/finops-agent-chart/
+## Below are a subset of values for the finopsAgent sub-chart that are most commonly changed.
+## See https://github.com/kubecost/finopsAgent-chart/
```
HMetcalfeW: We can keep the hyphen here

thomasvn:
Nice! We should also add a check to the `_helpers.tpl` to `fail` users if they had configured the old `.Values.finops-agent`, and tell them to use the new configuration. Something like:

```
{{- if index .Values "finops-agent" -}}
  {{- fail "..." }}
{{- end -}}
```

jessegoodier:
> Nice! We should also add a check to the `_helpers.tpl` to `fail` users if they had configured the old `.Values.finops-agent`, and tell them to use the new configuration. Something like:
> 
> ```
> {{- if index .Values "finops-agent" -}}
>   {{- fail "..." }}
> {{- end -}}
> ```

thanks for that, done!

HMetcalfeW (approved):
LGTM; great change

kubecost/Chart.yaml
```diff
@@ -10,6 +10,7 @@ annotations:
       url: https://www.kubecost.com
 dependencies:
   - name: finops-agent
+    alias: finopsagent
```
thomasvn, line 13: Thoughts on making it camelCase (i.e. `finopsAgent`)?

Our other service configs are currently camel cased (e.g. `networkCosts`, `clusterController`, `cloudCost`, `localStore`).

thomasvn: Oops I'm sorry. I just read this comment in your PR details:

> note that we intentionally are not using camelCase due to the key being used in k8s objects where uppercase letters are not valid.

thomasvn: Could you help me understand this a little more? ^^ Why has this not caused a problem for networkCosts, clusterController, cloudCost, or localStore?

thomasvn: Ok! After catching up on an internal Slack thread I understand the issue now

```
ServiceAccount "kc30-finopsAgent" is invalid: metadata.name: Invalid value: "kc30-finopsAgent": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters,
```

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

thomasvn (approved):

PR #4397, opened 2025-10-29
285-cleanup

JesseGoodier:
## cleanup 2.8.5

- fixes grafana dashboards/plugin registering
- Removes a warning about the duplicate port

in single pod installs:

```
helm install kubecost284 -n kubecost284 kubecost2x/cost-analyzer --version 2.8.4 --create-namespace
```

I1031 15:15:05.103453   50753 warnings.go:110] "Warning: spec.template.spec.containers[3].ports[0]: duplicate port name \"tcp-api\" with spec.template.spec.containers[2].ports[0], services and probes that select ports by name will use spec.template.spec.containers[2].ports[0]"

## Testing

Tested logs and grafana functionality.


cost-analyzer/templates/_helpers.tpl
```diff
@@ -1335,7 +1335,7 @@ Begin Kubecost 2.0 templates
   {{- end }}
   args: ["cloud-cost"]
   ports:
-    - name: tcp-api
+    - name: cloud-cost
```
thomasvn, line 1338: Just curious why you're doing this? Not a big deal! I don't think the name of the port matters.

jessegoodier: I should have added more detail for that.

in single pod installs:

```
helm install kubecost284 -n kubecost284 kubecost2x/cost-analyzer --version 2.8.4 --create-namespace
```

I1031 15:15:05.103453   50753 warnings.go:110] "Warning: spec.template.spec.containers[3].ports[0]: duplicate port name \"tcp-api\" with spec.template.spec.containers[2].ports[0], services and probes that select ports by name will use spec.template.spec.containers[2].ports[0]"

jessegoodier: I also confirmed that the cloud integration secret can still be created from the UI.

mittal-ishaan (approved):

PR #4400, opened 2025-10-30
2.9 fix aggregator volumeMount

JesseGoodier:
## fixes for 2.9

fix path

### DIFF
--- a/cost-analyzer/templates/_helpers.tpl
+++ b/cost-analyzer/templates/_helpers.tpl
@@ -162,7 +162,7 @@ RBAC exclusivity check: make sure either simple RBAC or RBAC Teams is configured
 Federated Storage source contents check. Either the Secret must be specified or the JSON, not both.
 */}}
 {{- define "federatedStorageSourceCheck" -}}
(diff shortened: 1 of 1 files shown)


mittal-ishaan (dismissed):

thomasvn:
Thanks for this catch @jessegoodier! I still see other locations with this issue. For example: https://github.com/kubecost/kubecost/blob/v2.9.2/cost-analyzer/templates/_helpers.tpl#L165

thomasvn (changes requested):
Other locations which need this fix:

- https://github.com/kubecost/kubecost/blob/v2.9.2/cost-analyzer/templates/_helpers.tpl#L1307
- https://github.com/kubecost/kubecost/blob/v2.9.2/cost-analyzer/templates/_helpers.tpl#L1439

jessegoodier:
> Thanks for this catch @jessegoodier! I still see other locations with this issue. For example: https://github.com/kubecost/kubecost/blob/v2.9.2/cost-analyzer/templates/_helpers.tpl#L165

thank you! done.

thomasvn (approved):

PR #4404, opened 2025-11-03
Add option to add budgets and budgets actions via helm

mittal-ishaan:
## Release Notes Headline
Add option to add budgets and budgets actions via helm

## Commentary for Reviewers
Add option to add budgets and budgets actions via helm
Populate the budget configmap name env in aggregator container for aggregator to pick it up and achnowledge  when running

## Links to Issues or Tickets
https://apptio.atlassian.net/browse/KCM-4791

## User Impact and Risks
No risk as they are disabled by default
It enables the user to add budgets and budget actions via helm values. (It adds on the already existing ones and do not overwrite completely)

## Testing
Created a helm install to test it.

## Documentation Updates
Needed, but yet to be done


kubecost/values.yaml
```diff
@@ -138,6 +138,36 @@ global:
...
         #     value: "service1" # corresponds to a value to filter cloud cost aggregate by service data on.
 
+  ## Kubecost Budget and Budget Actions
+  ## Ref: https://www.ibm.com/docs/en/kubecost/self-hosted/3.x?topic=ui-budgets
+  ##
+  budget:
+    enabled: false  # If true, overwrites budget parameters set through UI
+    budgetsConfig:
+      - name: "Example Budget 0"
+        budgetType: "allocations" # allocations, asset, cloud or collections
```
avrodrigues5: Looks like lint is failing! Could you fix that?

kubecost/values.yaml
```diff
@@ -138,6 +138,36 @@ global:
         #   - property: "service"
         #     value: "service1" # corresponds to a value to filter cloud cost aggregate by service data on.
 
+  ## Kubecost Budget and Budget Actions
+  ## Ref: https://www.ibm.com/docs/en/kubecost/self-hosted/3.x?topic=ui-budgets
+  ##
+  budget:
```
avrodrigues5: @chipzoller Should this be commented off in values.yaml?

kubecost/values.yaml
avrodrigues5: @chipzoller @ameijer Do we need an example for each — allocations, assets, cloud, and collection budgets — in the Helm values?

Chip, this is in reference to your request from 1–2 weeks ago to add Helm functionality for budgets and budget actions.

kubecost/values.yaml
chipzoller:
```suggestion
  budgets:
```
Needs to be plural. May need to update elsewhere.

chipzoller: If it functions as a master switch, no, I'd leave it enabled.

chipzoller (commented):
The `budgets` key needs to be under `kubecostProductConfigs` and not global. This is a product configuration.

avrodrigues5 (commented):
Please wait for @chipzoller Approval here. But I am good with this!

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.kubecostProductConfigs.budgets }}
+{{- if .Values.kubecostProductConfigs.budgets.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{default "budgets-configs" .Values.budgetsConfigmapName }}
```
chipzoller: Why does this name a value which isn't specified in the values file?

chipzoller (commented):
I don't understand the necessity of the `enabled` switch here. If this doesn't function as a master switch (where `enabled`=`false` does NOT prevent budgets from being created in the UI), and it seems like it doesn't, why couldn't we just dispense with that and only create the ConfigMap if the array at `budgets.budgetsConfig[]` has any objects?

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
mittal-ishaan: The other configmaps had something similar and their name helm config was not defined in the value file. I thought of keeping it like the rest. I will move it under Budgets in the values file.
Ref:
https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/aggregator/kubecost-alerts-configmap.yaml
https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/aggregator/kubecost-asset-reports-configmap.yaml
https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/aggregator/kubecost-cloud-cost-reports-configmap.yaml
https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/aggregator/kubecost-saved-reports-configmap.yaml

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.kubecostProductConfigs.budgets }}
+{{- if .Values.kubecostProductConfigs.budgets.budgetsConfig }}
+{{- if (gt (len .Values.kubecostProductConfigs.budgets.budgetsConfig) 0) }}
```
chipzoller: These can all be summarized in one line.

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
chipzoller: I don't like that we've done this. We should be keeping all product configs in the same place. And any Helm value should be populated in the main values file, even if commented out. We shouldn't have "hidden" Helm values as the values file should serve as the authoritative config file to the product.

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
mittal-ishaan: thank you for pointing out. Updated it now

kubecost/values.yaml
```diff
@@ -1347,6 +1347,36 @@ kubecostProductConfigs:
     #           profile: "Development"
     #           rightSizingWindow: "24h"
 
+  ## Kubecost Budgets and Budgets Actions
+  ## Ref: https://www.ibm.com/docs/en/kubecost/self-hosted/3.x?topic=ui-budgets
+  ##
```
jessegoodier, line 1358:
```suggestion
  ## Only provide a configMapName if using an existing configMap. Otherwise leave the value as ""
  ##  
```

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if ((.Values.kubecostProductConfigs).budgets).budgetsConfig }}
```
jessegoodier, line 1:
```suggestion
{{- if ((.Values.kubecostProductConfigs).budgets).budgetsConfig }}
{{- if .Values.kubecostProductConfigs.budgets.configmapName }}
{{ fail "\nError: Only provide a .Values.kubecostProductConfigs.budgets.configmapName if using an existing configMap. Otherwise leave the value null."
{{- end }}
```

JesseGoodier (commented):
see my comments. just trying to prevent confusion.
But yours looks good if you prefer to allow customizing the configMap name when the yaml is also passed.

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
mittal-ishaan: I mean there is no harm in setting the configmap name with defined budgetsConfigs yaml that would create a new cm in itself.

> if you prefer to allow customizing the configMap name when the yaml is also passed.

Yes

mittal-ishaan: should we fail though, in this case?

kubecost/templates/aggregator/aggregator-statefulset.yaml
jessegoodier, line 61: should we add to the checksum?

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
jessegoodier: all good, I don't have a strong opinion.

JesseGoodier (approved):

kubecost/templates/aggregator/aggregator-statefulset.yaml
mittal-ishaan: Yes sure 👍
Ideally we would not want that. there is a configwatcher in aggregator that looks for any changes in the configmap and updates accordingly and there would be no need for aggregator restart when it updates

kubecost/templates/aggregator/kubecost-budgets-configmap.yaml
mittal-ishaan: Noted. I will add it to my to do's to correct this.

kubecost/values.yaml
mittal-ishaan: I will create documentation for it and there we can add other examples

kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -454,6 +454,10 @@ spec:
             - name: SMTP_CONFIGMAP_NAME
               value: {{ .Values.smtpConfigmapName }}
             {{- end }}
+            {{- if .Values.kubecostProductConfigs.budgets.configmapName }}
+            - name: BUDGETS_CONFIGMAP_NAME
+              value: {{ .Values.kubecostProductConfigs.budgets.configmapName }}
+            {{- end }}
```
thomasvn, line 460: @mittal-ishaan Is there a potential logic issue here?

By default the values.yaml sets this config to an empty string 

```yaml
kubecostProductConfigs:
  budgets:
    configmapName: ""
```

Therefore if a user doesn't set this config (which seems to be optional), then the `BUDGETS_CONFIGMAP_NAME` env var is never set? Is that the intended behavior?

mittal-ishaan: this was the intended behavour. If the user has its own configmap or they want to defien the name of the configmap on their own, they can define that and this env would be populated. Otherwise we do not set `BUDGETS_CONFIGMAP_NAME` env and the aggregator [defaults to](https://github.com/kubecost/kubecost-cost-model/blob/d68f54257ce36de96f843956b339e01cfac92325/pkg/env/costmodelenv.go#L108-L110) `budgets-configs` configmap name when this env is not set. Which is also what we [default to](https://github.com/kubecost/kubecost/blob/develop/kubecost/templates/aggregator/kubecost-budgets-configmap.yaml) when creating the budgets configmap name.

thomasvn: Great! Thanks for clarifying.

PR #4423, opened 2025-11-11
clean up values and comments for cluster controller

JesseGoodier:
## clean up values and comments for cluster controller

change mount to allow custom pre-configured configmap

chart tests and lint pass.

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -1236,6 +1236,7 @@ clusterController:
 
   rbac:
     # In 3.x, only containerRequestRightSizing and resourceQuotaRightSizing are supported, all other features are in progress.
+    # The below settings allow the cluster controller to perform actions on the cluster.
     containerRequestRightSizing: true
     resourceQuotaRightSizing: true
     clusterRightSizing: false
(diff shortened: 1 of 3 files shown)


jessegoodier:
what is the plan for 3.1 @chipzoller ? 

I think we should uncomment the values at a minimum. 

this one is not urgent.

chipzoller:
I'm not sure what this is supposed to do. Actions should be disabled by default until they are fixed for free and EKS Optimized tiers, which is expected in 3.1. And, from then, they should be enabled by default with only the container request sizing use case enabled.

kubecost/values.yaml
```diff
@@ -1286,8 +1287,29 @@ telemetry:
 kubecostProductConfigs:
   actions:
     # Enable multi-cluster cluster controller actions (v2)
-    enabled: false
-
+    enabled: true
```
jessegoodier:
```suggestion
    enabled: false
```

jessegoodier:
> I'm not sure what this is supposed to do. Actions should be disabled by default until they are fixed for free and EKS Optimized tiers, which is expected in 3.1. And, from then, they should be enabled by default with only the container request sizing use case enabled.

Sounds good. I set the values.yaml back to false. 

This is just a cleanup of comments and uncommenting values.

PR #4427, opened 2025-11-11
Allow turbonomic integration using existing secret 

mittal-ishaan:
## Release Notes Headline
Allow turbonomic integration using existing secret 

## Commentary for Reviewers
Right now users have to provide their turbonomic client secrets and other configs via helm values to get applied. The change allows users to mount an existing secret to the aggregator pod with turbonomic configuration so they now do not need to add it in their values file.
I need to add turbonomic configuration for a customer in our argoCD setup and we do not allow secrets being pushed to github repo
## User Impact and Risks
No major impact as it adding an existing secret is disabled by default

## Testing
Helm teplate looks good

### DIFF
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -501,6 +501,10 @@ kubecost.costEventsAudit.enabled flag for nginx configmap
 {{- end -}}
 {{- end -}}
 
+{{- define "kubecost.turbonomic.secretName" -}}
+{{- default "kubecost-integrations-turbonomic" .Values.global.integrations.turbonomic.secretName -}}
+{{- end -}}
+
 {{- /*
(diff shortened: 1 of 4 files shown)


kubecost/templates/_helpers.tpl
```diff
@@ -501,6 +501,10 @@ kubecost.costEventsAudit.enabled flag for nginx configmap
 {{- end -}}
 {{- end -}}
 
+{{- define "kubecost.turbonomic.secretName" -}}
+{{- default "kubecost-integrations-turbonomic" .Values.global.integrations.turbonomic.secretName -}}
```
ameijer, line 505: please include the release name in the default  name, and truncate to 63 chars, strip any trailing - after truncation

PR #4445, opened 2025-11-18
Fix: frontend.service.type override ignored

Cedric-Lefebvre:
## Release Notes Headline

Fix: frontend.service.type was ignored when set.

## Commentary for Reviewers

The template always rendered `ClusterIP`. The fix makes the chart correctly use the value provided by the user.

## Links to Issues or Tickets

N/A

## User Impact and Risks

Users can now set frontend.service.type and get the expected result.
Very low risk.

## Testing

```
helm template .  --set frontend.service.type=NodePort -s templates/frontend/frontend-service.yaml
```

The output now correctly shows type: `NodePort`.

## Documentation Updates

N/A


JesseGoodier (approved):
Thank you!

PR #4450, opened 2025-11-20
Update values.yaml

nandrews-ibm:
## Release Notes Headline

Removed reference to "bingen" which is internal to Kubecost product team and should not be referenced externally to avoid confusion.


JesseGoodier (approved):
Thanks!

PR #4475, opened 2025-12-04
set default network costs affinity

JesseGoodier:
## set default network costs affinity
The networkCosts daemonset requires specific labels in order to work. This sets a nodeSelector that will only run the pod on nodes with the required label.

## Links to Issues or Tickets

[KCM-4805]

## User Impact and Risks

This will remove the DS on an upgrade, but we cannot think of a reason a user would want to run the DS if they do not have the required labels.

## Testing

```sh
helm template kubecost ./kubecost --set networkCosts.useDefaultAffinity=false| kubeconform -strict -summary
Summary: 41 resources found parsing stdin - Valid: 41, Invalid: 0, Errors: 0, Skipped: 0
```
```sh
helm template kubecost ./kubecost --set networkCosts.useDefaultAffinity=true| kubeconform -strict -summary
Summary: 41 resources found parsing stdin - Valid: 41, Invalid: 0, Errors: 0, Skipped: 0
```
installed on microk8s and it removed the pod:
image

upgraded on cluster with required label and the pods continue to run.

## Documentation Updates

[KCM-4805]: https://apptio.atlassian.net/browse/KCM-4805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jessegoodier:
@thomasvn FYI, in our DM group, Matt was good with this.
allows us to mitigate the log spam with our new defaults and allows users to run it with a simple override.

kubecost/templates/network-costs/network-costs-daemonset.yaml
```diff
@@ -121,9 +121,19 @@ spec:
...
-        {{- toYaml . | nindent 8 }}
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: topology.kubernetes.io/region
+                operator: Exists
+              - key: topology.kubernetes.io/zone
+                operator: Exists
+        {{- else }}
```
thomasvn: Small style change. Indentation to match `if`.

```suggestion
      {{- else }}
```

kubecost/values.yaml
```diff
@@ -814,6 +814,10 @@ networkCosts:
       #      - "20.0.0.0/8"
 
   tolerations: []
+  ## useDefaultAffinity: if true, the affinity will be set to a nodeAffinity that will only schedule
+  ## networkCosts pods to run on nodes that have a topology.kubernetes.io/region label.
+  ## If false, the affinity will be set to the affinity provided in the affinity block.
+  useDefaultAffinity: true
   affinity: {}
```
thomasvn: Instead of creating a new config `useDefaultAffinity`, what do you think about just setting the default right here in the `values.yaml`? Then users have the option to override it or modify it as they please.

```suggestion
  ## The networkCosts pod requires labels "topology.kubernetes.io/region" and
  ## "topology.kubernetes.io/zone" for traffic classification. Set `affinity: {}`
  ## to run the networkCosts pod even in the absence of these labels.
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: topology.kubernetes.io/region
            operator: Exists
          - key: topology.kubernetes.io/zone
            operator: Exists
```

thomasvn:
Good idea with setting a nodeAffinity for this pod requiring labels "topology.kubernetes.io/region" and 'topology.kubernetes.io/zone"! This generally looks good to me, but I had a suggestion about how we apply the default.

jessegoodier:
> Good idea with setting a nodeAffinity for this pod requiring labels "topology.kubernetes.io/region" and 'topology.kubernetes.io/zone"! This generally looks good to me, but I had a suggestion about how we apply the default.

done. i was worried it would make it difficult to null based on my previous memory of this. But it works!

thomasvn (approved):

PR #4499, opened 2025-12-17
Fix localStore.persistentVolume.existingClaim not being respected

MikeFindsThings:
## Release Notes Headline

Fix `localStore.persistentVolume.existingClaim` being ignored during deployment

## Commentary for Reviewers

The deployment template was using Helm's `default` function with arguments in the wrong order, causing it to always use the generated PVC name and ignore the user-specified `existingClaim` value. This prevented customers from upgrading from Kubecost 2.x while preserving their existing `kubecost-cost-analyzer` PVC.

**Key change:** Line 56 in `cluster-storage-deployment.yaml` now uses the existing `kubecost.localStore.pvcName` helper function (defined in `_helpers.tpl:33-41`) which properly handles the `existingClaim` logic with if/else conditions.

The PVC creation template already uses this helper correctly - this PR aligns the deployment template to match.

## Links to Issues or Tickets

N/A - Customer-reported issue during 2.x to 3.x upgrade

## User Impact and Risks

**Impact:**
- **Positive:** Customers can now successfully set `localStore.persistentVolume.existingClaim` to reference existing PVCs during upgrades
- **Critical for:** 2.x to 3.x upgrades where customers want to preserve their existing storage

**Risks:**
- **Low risk:** The change uses an existing, well-tested helper function that's already used by the PVC template
- **No breaking changes:** Default behavior (empty `existingClaim`) remains unchanged
- **Backward compatible:** Existing deployments without `existingClaim` set are unaffected

## Testing

1. **Created test Helm chart** demonstrating the bug:
   - Old method: `default "kubecost-cost-analyzer" "release-local-store"` always returns `"release-local-store"` (incorrect)
   - New method: `include "kubecost.localStore.pvcName"` returns `"kubecost-cost-analyzer"` (correct)

2. **Verified both scenarios:**
   - With `existingClaim="kubecost-cost-analyzer"` - Uses specified PVC name (correct)
   - With `existingClaim=""` (default) - Uses generated name `<release>-local-store` (correct)

3. **Confirmed helper function logic** in `_helpers.tpl` properly handles:
   - `existingClaim` set - use it
   - `fullnameOverride` set - use it
   - Otherwise - use generated name

## Documentation Updates

N/A - This fixes existing documented functionality in `values.yaml:577` which already mentions using `existingClaim` for 2.x upgrades.


thomasvn (approved):
Great find here @MikeFindsThings, and thank you for the detailed PR notes!

PR #4512, opened 2026-01-08
fix: Add persistent storage for cloud-cost and fix Helm template errors

JesseGoodier:
## Summary
This PR addresses a TODO to add persistent storage support for the cloud-cost config.

- **kubecost/templates/cloud-cost/cloud-cost-persistent-configs-pvc.yaml**: New PVC template for cloud-cost persistent configs
- **kubecost/templates/cloud-cost/cloud-cost-deployment.yaml**: Updated to use PVC when enabled, with emptyDir as fallback
- **kubecost/templates/_helpers.tpl**: Added validation check to warn users when persistent storage is disabled
- **kubecost/templates/NOTES.txt**: Added persistent storage validation check to installation notes
- **kubecost/values.yaml**: Added `persistentConfigsStorage` configuration section
- **kubecost/values-storage.yaml**: Added storage-specific configuration

## Benefits
- Prevents data loss on pod restarts for cloud cost configurations
- Maintains backward compatibility with emptyDir fallback
- Provides clear warnings when persistent storage is disabled without cloud integration secret

## Testing
- ✅ Helm template generation successful (2071 lines rendered)
- ✅ Successfully deployed with `helm upgrade kc-develop ./kubecost`
- ✅ PVC properly rendered in templates
- ✅ No template parsing errors

## Checklist
- [x] Changes committed with descriptive message
- [x] Helm template validation passed
- [x] Successfully deployed to test cluster
- [x] Backward compatible (emptyDir fallback maintained)

### DIFF
--- a/kubecost/templates/NOTES.txt
+++ b/kubecost/templates/NOTES.txt
@@ -8,6 +8,7 @@
 {{- include "kubecost.caCertsSecretConfig.check" . -}}
 {{- include "kubecost.actionsStorageSourceCheck" . -}}
 {{- include "kubecost.finopsagentCheck" . -}}
+{{- include "kubecost.cloudCost.persistentStorage.check" . -}}
 ================================================================================
 KUBECOST INSTALLATION SUCCESSFUL
 ================================================================================
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -477,6 +477,25 @@ kubecost.costEventsAudit.enabled flag for nginx configmap
   {{- end }}
 {{- end }}
 
+{{/*
+Check if cloud cost persistent storage is disabled and warn user
+*/}}
+{{- define "kubecost.cloudCost.persistentStorage.check" }}
+{{- if .Values.cloudCost.enabled }}
+{{- if not ((.Values.cloudCost).persistentConfigsStorage).enabled }}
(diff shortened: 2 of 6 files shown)


jessegoodier:
I found this during testing because my UI entered cloud config kept getting removed.

thomasvn:
@jessegoodier Nice, thank you! What's the main motivation for this PR?

According to your comment, it sounds like any UI configured Cloud Integrations are written as JSON to the local filestore? And if there is no PV that Cloud Integration gets wiped? If so, this makes sense!

With respect to CloudCost ETL files, we shouldn't need them on a PV attached to the CloudCosts pod. Those ETL files should either be pushed to a federated storage bucket, or pushed to the local-store pod and its attached PV.

jessegoodier:
> @jessegoodier Nice, thank you! What's the main motivation for this PR?
> 
> 
> 
> According to your comment, it sounds like any UI configured Cloud Integrations are written as JSON to the local filestore? And if there is no PV that Cloud Integration gets wiped? If so, this makes sense!
> 
> 
> 
> With respect to CloudCost ETL files, we shouldn't need them on a PV attached to the CloudCosts pod. Those ETL files should either be pushed to a federated storage bucket, or pushed to the local-store pod and its attached PV.

Just the config from ui getting removed when the pod restarts. 
No other files are in there.

thomasvn (commented):
LGTM, thank you for fixing!

thomasvn (approved):

PR #4535, opened 2026-01-22
Fix diagnostics and heartbeat disable issue

mittal-ishaan:
## Release Notes Headline
Make diagnostics and heartbeat to acknowledge disabled config

## Commentary for Reviewers
When diagnostics or heartbeat is disabled, we do not set its variable to either true or false. the aggregator defaults to diagnostics/heartbeat true if it does not find their env. vars., in turn making it impossible to disable the diagnostics by intended way.

## Testing
helm template


JesseGoodier (approved):

jessegoodier:
added 3.0 just in case

PR #4558, opened 2026-02-02
Add annotations to PVCs

mittal-ishaan:
## Release Notes Headline
Add annotations to PVCs

## Commentary for Reviewers
Add helm config to allow users to add annotations to PVCs.
Also added a default PVC annotation `helm.sh/resource-policy: keep` to prevent them from getting deleted

## Testing
helm template
helm install and confirmed the created PVCs had custom annotations and if no custom annotations were given, it had the default `helm.sh/resource-policy: keep` annotation.

### DIFF
--- a/kubecost/templates/aggregator/aggregator-statefulset.yaml
+++ b/kubecost/templates/aggregator/aggregator-statefulset.yaml
@@ -51,7 +51,7 @@ spec:
       # a need to "mount" ConfigMap files (using the watcher) to a file system;
       # that's what this per-replica Volume is used for.
       name: persistent-configs
-      {{- with .Values.aggregator.aggregatorDbStorage.annotations }}
+      {{- with .Values.aggregator.persistentConfigsStorage.annotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -47,6 +51,10 @@ spec:
       # a need to "mount" ConfigMap files (using the watcher) to a file system;
       # that's what this per-replica Volume is used for.
       name: persistent-configs
+      {{- with .Values.aggregator.aggregatorDbStorage.annotations }}
```
HMetcalfeW: Perhaps to keep the persistent configs portion consistent with storageClass overrides, perhaps we use `.Values.aggregator.persistentConfigsStorage.annotations`

mittal-ishaan: oops, thank you for pointing it out

HMetcalfeW (approved):
LGTM but defer to @ameijer and @jessegoodier for further review.

kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -28,6 +28,10 @@ spec:
     kind: PersistentVolumeClaim
     metadata:
       name: aggregator-db-storage
+      {{- with .Values.aggregator.aggregatorDbStorage.annotations }}
```
ameijer: you should merge this with global annotations

mittal-ishaan: As in use the global annotations or add a new helm values config for global annotation for PVCs?

ameijer (approved):

PR #4562, opened 2026-02-05
FIX: Standardize cloud-cost service account configuration

JesseGoodier:
## Release Notes Headline

Standardize cloud-cost service account configuration to match cluster-controller pattern

## Commentary for Reviewers

This PR aligns the cloud-cost service account configuration with the cluster-controller pattern for consistency across the codebase. Key changes:

1. **Service Account Template**: Updated condition logic to check both `enabled` and `serviceAccount.create` flags, matching cluster-controller behavior
2. **Naming Convention**: Changed to use `kubecost.cloudCost.fullname` helper to include release name prefix (e.g., `my-release-cloud-cost` instead of just `cloud-cost`)
3. **Annotations Logic**: Simplified to use single-level if/else with fallback to global `serviceAccount.annotations`
4. **Helper Function**: Updated `kubecost.cloudCost.serviceAccountName` to use `fullname` instead of `name`
5. **Values Cleanup**: Removed deprecated `cloudCost.serviceAccountName` field in favor of `cloudCost.serviceAccount.name`

## Links to Issues or Tickets

N/A

## User Impact and Risks

**Impact:**
- Service account names will now include the release name prefix (e.g., `my-release-cloud-cost`)
- Users upgrading will see a new service account created with the prefixed name
- The old service account (without prefix) will remain but won't be used

**Risks:**
- **Breaking Change**: Existing deployments using the cloud-cost service account will need to update any references to the new prefixed name
- RBAC configurations referencing the old service account name will need updates
- Workload identity/IAM role bindings tied to the old service account name must be reconfigured

**Mitigation:**
- Users can override with `cloudCost.serviceAccount.name` to maintain the old name if needed
- Clear upgrade notes should be provided in release documentation

## Testing

- Verified template syntax matches cluster-controller pattern
- Confirmed `kubecost.cloudCost.fullname` helper generates correct name with release prefix
- Validated annotation fallback logic works correctly
- Checked that `serviceAccount.name` override still functions as expected
image

## Documentation Updates

Documentation should be updated to reflect:
- New service account naming convention
- Migration steps for existing deployments
- How to use `cloudCost.serviceAccount.name` to override the default name

### DIFF
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -620,7 +620,7 @@ This will block upgrades until the new value is used, which very few would have 
 */}}
 {{- define "kubecost.finopsagentCheck" -}}
 {{- if index .Values "finops-agent" }}
-  {{ fail "\nThe helm values for finops-agent have been updated.\nPlease change the finops-agent: key in your helm values to finopsagent:\nIf you are using helm with a local repo clone, be sure to run\nhelm dependency build ./kubecost\nbefore installing or upgrading.\n" }}
+  {{ fail "\nThe helm values for finops-agent have been updated.\nPlease change the finops-agent: key in your helm values to finopsagent:\nIf you are using helm with a local repo clone, be sure to run\nhelm dependency update kubecost/\nbefore installing or upgrading.\n" }}
 {{- end -}}
 {{- end -}}
 
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -620,7 +620,7 @@ This will block upgrades until the new value is used, which very few would have 
 */}}
 {{- define "kubecost.finopsagentCheck" -}}
 {{- if index .Values "finops-agent" }}
-  {{ fail "\nThe helm values for finops-agent have been updated.\nPlease change the finops-agent: key in your helm values to finopsagent:" }}
+  {{ fail "\nThe helm values for finops-agent have been updated.\nPlease change the finops-agent: key in your helm values to finopsagent:\nIf you are using helm with a local repo clone, be sure to run\nhelm dependency build ./kubecost\nbefore installing or upgrading.\n" }}
 {{- end -}}
 {{- end -}}
 
--- a/kubecost/templates/cloud-cost/cloud-cost-service-account.yaml
+++ b/kubecost/templates/cloud-cost/cloud-cost-service-account.yaml
@@ -1,18 +1,15 @@
-{{- if and .Values.cloudCost.serviceAccount.create (not .Values.cloudCost.serviceAccount.name) }}
+{{- if and (.Values.cloudCost).enabled .Values.cloudCost.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ template "kubecost.cloudCost.serviceAccountName" . }}
+  name: {{ include "kubecost.cloudCost.fullname" . }}
   namespace: {{ .Release.Namespace }}
   labels:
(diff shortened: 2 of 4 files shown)


kubecost/templates/cloud-cost/cloud-cost-service-account.yaml
```diff
@@ -1,13 +1,15 @@
-{{- if and .Values.serviceAccount.create .Values.cloudCost.serviceAccountName }}
```
jessegoodier, line 1: this isn't specific enough.

kubecost/values.yaml
```diff
@@ -1202,9 +1202,15 @@ cloudCost:
...
-  serviceAccountName: ""
+  ## Service account configuration for cloud cost pod
+  serviceAccount:
+    create: true
+    annotations: {}
+    ## Optionally specify an existing service account name to use instead of creating one
+    name: ""
+    ## When using a unique cloud-cost service account, set this to false
+    ## Default is true for backward compatibility
+    useSharedServiceAccount: true
```
jessegoodier, line 1213: are you guys good with this?

kubecost/templates/_helpers.tpl
```diff
@@ -620,7 +620,7 @@ This will block upgrades until the new value is used, which very few would have
 */}}
 {{- define "kubecost.finopsagentCheck" -}}
 {{- if index .Values "finops-agent" }}
-  {{ fail "\nThe helm values for finops-agent have been updated.\nPlease change the finops-agent: key in your helm values to finopsagent:" }}
+  {{ fail "\nThe helm values for finops-agent have been updated.\nPlease change the finops-agent: key in your helm values to finopsagent:\nIf you are using helm with a local repo clone, be sure to run\nhelm dependency update kubecost/\nbefore installing or upgrading.\n" }}
```
thomasvn, line 623: Just curious is there a reason you're adding this specific error message here? Have you seen this occur in the wild?

If users really are using a local repo clone, we should probably blast this log in many more places than just the check for `.Values.finops-agent`!

thomasvn:
Nice thanks for tackling this! Can you clarify what the motivation is here?

Currently both the Aggregator and CloudCost pods share the same [ServiceAccount](https://github.com/kubecost/kubecost/blob/fadc0401bf1c9bcebb759dc7c210db8dd20e14e7/kubecost/templates/kubecost-service-account.yaml#L1-L6) which attach to a common [Role](https://github.com/kubecost/kubecost/blob/fadc0401bf1c9bcebb759dc7c210db8dd20e14e7/kubecost/templates/kubecost-role-template.yaml#L1-L3). Is the goal to create clean separation between these two ServiceAccounts? Sorry for my confusion.

kubecost/templates/_helpers.tpl
jessegoodier: no, just to help our team. or anyone else that is adding features.

jessegoodier:
> Nice thanks for tackling this! Can you clarify what the motivation is here?
> 
> Currently both the Aggregator and CloudCost pods share the same [ServiceAccount](https://github.com/kubecost/kubecost/blob/fadc0401bf1c9bcebb759dc7c210db8dd20e14e7/kubecost/templates/kubecost-service-account.yaml#L1-L6) which attach to a common [Role](https://github.com/kubecost/kubecost/blob/fadc0401bf1c9bcebb759dc7c210db8dd20e14e7/kubecost/templates/kubecost-role-template.yaml#L1-L3). Is the goal to create clean separation between these two ServiceAccounts? Sorry for my confusion.

I needed it for a role for cloud cost that was cross account and the bucket in our account.
So aggregator role to access our bucket
cloud cost role to cross account cloud billing

mittal-ishaan (approved):

PR #4571, opened 2026-02-10
Revert "Add annotations to PVCs"

mittal-ishaan:
Reverts kubecost/kubecost#4558

Statefulset does not allow PVC annotatiosn to be updated


thomasvn (approved):
Could you also share in the PR details the exact k8s error you were running into?

mittal-ishaan:
sure @thomasvn Creating a new PR, with a workaround for it rather than reverting. Will add description there.

PR #4595, opened 2026-02-27
Implement an httproute

ameijer:
## Release Notes Headline

add httproute support to kubecost chart 

## Commentary for Reviewers
ingress-nginx is moribund. we need to be ready for folks starting to use gateway API. also had bob build a test 

## Links to Issues or Tickets
N/A

## User Impact and Risks
disabled by default; impact minimal. We will test this in integration tests to ensure it WAI 

## Testing
integration tested


kubecost/templates/httproute-template.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.httpRoute.enabled }}
+{{- if not .Values.httpRoute.parentRefs }}
+{{- fail "httpRoute.enabled is true but httpRoute.parentRefs is empty. Set at least one parentRef with a name (e.g. parentRefs: [{ name: my-gateway }])." }}
+{{- end }}
+{{- if not (index .Values.httpRoute.parentRefs 0).name }}
+{{- fail "httpRoute.enabled is true but the first parentRef has no name. Set parentRefs[0].name to your Gateway name." }}
+{{- end }}
```
thomasvn, line 7: I don't think we would ever hit this `fail` condition, because the `values.yaml` currently defines a default `parentRefs` value? Should we comment that out in the `values.yaml`?

```yaml
  parentRefs:
    - name: gateway
```

thomasvn: Aside from this one minor thing, overall LGTM!

ameijer: so, leaving the default value in matches our existing behavior more closely where we deploy an ingress object with the default ingress class, even if you don't have a controller listening on that class. The guard here is that if someone overrides the array but forgets to set name

thomasvn: Understood, approving!

thomasvn (approved):

PR #4614, opened 2026-03-10
Feat: add cluster controller rate limiter parameters in helm

ErikTorres1998:
## Release Notes Headline

Add rate limiter configuration to cluster controller Helm values

## Commentary for Reviewers

Added `rateLimiter.eventsPerSecond` and `rateLimiter.eventBurst` parameters under `clusterController` in values.yaml, wired to `RATE_LIMITER_EVENTS_PER_SECOND` and `RATE_LIMITER_EVENT_BURST` environment variables in the deployment spec. Both parameters are optional.

## Links to Issues or Tickets
https://apptio.atlassian.net/browse/KCM-5396

## User Impact and Risks

Users can now configure cluster controller rate limiting via Helm values. Low risk - parameters are optional and backward compatible.

## Testing

Verified Helm template syntax and conditional logic for environment variables.

## Documentation Updates

N/A

### DIFF
--- a/kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
+++ b/kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
@@ -145,6 +145,16 @@ spec:
         - name: CC_KUBESCALER_DEFAULT_RESIZE_ALL
           value: {{ .Values.clusterController.kubescaler.defaultResizeAll | default "false" | quote }}
         {{- end }}
+        {{- if .Values.clusterController.rateLimiter }}
+        {{- if .Values.clusterController.rateLimiter.eventsPerSecond }}
+        - name: RATE_LIMITER_EVENTS_PER_SECOND
+          value: {{ .Values.clusterController.rateLimiter.eventsPerSecond | quote }}
+        {{- end }}
+        {{- if .Values.clusterController.rateLimiter.eventBurst }}
+        - name: RATE_LIMITER_EVENT_BURST
+          value: {{ .Values.clusterController.rateLimiter.eventBurst | quote }}
+        {{- end }}
+        {{- end }}
         {{- if .Values.clusterController.extraEnv -}}
         {{ toYaml .Values.clusterController.extraEnv | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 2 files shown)


jessegoodier:
edited title, feel free to change back if this is not accurate

kubecost/values.yaml
```diff
@@ -1323,6 +1323,9 @@ clusterController:
   extraEnv: []
   # - name: EXTRA_ENV_VAR
   #   value: "extra_env_var_value"
+  rateLimiter:
+    eventsPerSecond: 0.1
+    eventBurst: 10
```
biancaburtoiu: I would add these further down below, after `clusterController.rbac` and also prepend a brief explanatory comment. Something along the lines of
```yaml
# Controls the rate at which the cluster controller executes operations against the k8s API.
```

They should also be commented out as the code already uses defaults if the envvars aren't explicitly set.

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -145,6 +145,16 @@ spec:
         - name: CC_KUBESCALER_DEFAULT_RESIZE_ALL
           value: {{ .Values.clusterController.kubescaler.defaultResizeAll | default "false" | quote }}
         {{- end }}
+        {{- if .Values.clusterController.rateLimiter }}
```
biancaburtoiu, line 148: were you able to double check e.g. in an. install w/ a local chart that if we do set these parameters, they are correctly populated in the cluster controller?

biancaburtoiu (changes requested):

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
ErikTorres1998: I verified the template rendering locally with helm template and confirmed the environment variables are correctly generated in the deployment spec. I haven't tested with an actual cluster install yet to verify the running pod. The template logic looks correct would you like me to do a full cluster install test or is the template verification sufficient for review?

biancaburtoiu: I think it's ok to merge and test afterwards using the nightly chart and a custom cluster controller image (since we haven't released a new CC version since your changes have been merged)

biancaburtoiu (approved):
LGTM, thank you!

PR #4639, opened 2026-04-07
Update agent to 1.0.16

JesseGoodier:
## Update agent to 1.0.16
Update chart to use latest Finops agent.

## Commentary for Reviewers

Do we want to CP this in 3.1?

## User Impact and Risks

Remove PV for Finops Agent by default. It was not used previously, due to a bug in the path. The cost of the PV generally outsweighs the benefit it provides.

## Testing

Using internally for weeks, update to 1.0.16 works as expected.

## Documentation Updates

updated values and example storage configs.

### DIFF
--- a/kubecost/values-storage.yaml
+++ b/kubecost/values-storage.yaml
@@ -15,12 +15,14 @@ localStore:
     ## Note that setting enabled to false means all data will be lost out on pod restart unless using federated storage.
     enabled: true
 
-## short-term (<2 days) metrics are stored by the finops agent.
+## The PV for the Finops Agent is only used when the object-store is unavailable.
+## It is generally not needed unless the object store connectivity is poor.
+## https://github.com/kubecost/finops-agent-chart/tree/main/charts/finops-agent#persistence for more detail
 finopsagent:
   persistence:
     storageClass: ""
     size: 8Gi
-    enabled: true
+    enabled: false
 
 aggregator:
   ## The aggregator pod has 2 PVCs and will need to be in the same availability zone unless using multi-zone storage.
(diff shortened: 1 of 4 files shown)


thomasvn:
Looks good, thank you for this update @jessegoodier!

A few minor change requests though. I was running some tests today with FinOps Agent v1.0.8 and v1.0.16 (latest). I've found that Kubecost does not write anything to the existing PV. All WAL files are written directly to the configured federated storage bucket.

When Persistence for the FinOps Agent is enabled it creates the volumeMount `data` which has a mountPath only used by Cloudability https://github.com/kubecost/finops-agent-chart/blob/v1.0.8/charts/finops-agent/templates/deployment.yaml#L383-L386. Therefore, I think we can do further cleanup and remove `.Values.finopsagent.persistence` entirely from this Helm chart!

kubecost/values-storage.yaml
```diff
@@ -15,12 +15,14 @@ localStore:
...
-## short-term (<2 days) metrics are stored by the finops agent.
+## The PV for the Finops Agent is only used when the object-store is unavailable.
+## It is generally not needed unless the object store connectivity is poor.
+## https://github.com/kubecost/finops-agent-chart/tree/main/charts/finops-agent#persistence for more detail
 finopsagent:
   persistence:
     storageClass: ""
     size: 8Gi
-    enabled: true
+    enabled: false
```
thomasvn, line 25: I think we can remove this section entirely!

```suggestion
```

kubecost/values.yaml
thomasvn, line 628: Thoughts on deleting this section entirely? In v1.0.16 persistence is also going to be disabled by default https://github.com/kubecost/finops-agent-chart/blob/v1.0.16/charts/finops-agent/values.yaml#L613-L619

jessegoodier:
totally forgot about this one. going to close and add to open #4649

kubecost/values.yaml
jessegoodier: agreed, confusing otherwise

PR #4643, opened 2026-04-24
Stop regenerating kubecost-rbac-secret on every render

cpetersen5:


jessegoodier:
I didn't see this until after I did some work on my PR: https://github.com/kubecost/kubecost/pull/4645

Good callout, copied some of this into mine. 

Mine has a few differences:

1. saml only config without rbac also has the randAlphaNum
2. add check for ci cd flag 
3. completely removes the secrets from the configsChecksum to simplify OpenShift support.

PR #4645, opened 2026-04-27
ci friendly sso and rbac secrets

JesseGoodier:
Generated comprehensive PR description for commit e4976ea681fbac06e15dad54ef6b092c5d28d9e7.

The PR description includes:

**Release Notes Headline:**
Fixed SSO authentication secrets to prevent rotation on every Helm upgrade and added explicit secret configuration requirements for CI/CD deployments.

**Technical Detail**

When oidc.rbac or saml.rbac is enabled, helm chart creates a secret named
"kubecost-rbac-secret" containing a randAlphaNum used for HMAC signing key of RBAC tokens.
This causes CI tools t reconcile the changes and restart pods due to the previous inclusion of that secret in the `configsChecksum`

The above behavior is also true for any SAML configuration, regardless of rbac being enabled.

**Key Changes:**
1. New `kubecost.authSecretValue` helper function that intelligently resolves secrets by:
   - Using explicit values when provided
   - Looking up existing cluster secrets to preserve them
   - Generating random values only for new installations
   - Enforcing explicit values in CI/CD mode

2. New `kubecost.rbacAuthSecretValue` wrapper for RBAC configurations

3. Updated secret templates to use the new helpers

4. Enhanced values.yaml documentation with CI/CD requirements

5.  Remove the secret from the checkSum and add notes to the values file explaining the limitation. This was done to simplify OpenShift support. The secret is rarely if ever changed, so this should not cause any support issues.

**User Impact:**
- Eliminates authentication failures after Helm upgrades
- Preserves existing secrets automatically
- Requires explicit authSecret values for CI/CD deployments (when skipSanityChecks=false)

**Testing Recommendations:**

When not skipping sanity checks, fail and give advice on what needs to be set:

Correct config:

```
helm upgrade kc-jesse -n kc-jesse ./kubecost \
  --create-namespace \
  -f ./temp/helmValues-saml.yaml \
  --set global.platforms.cicd.enabled=true \
  --set saml.authSecret=`openssl rand -hex 32`
```

Incorrect:

```
helm upgrade kc-jesse -n kc-jesse ./kubecost \
  --create-namespace \
  -f ./temp/helmValues-saml.yaml \
  --set global.platforms.cicd.enabled=true
--- 
Error: UPGRADE FAILED: execution error at (kubecost/templates/kubecost-saml-secret-template.yaml:13:19): 
Error: .Values.saml.authSecret must be set when global.platforms.cicd.enabled=true and global.platforms.cicd.skipSanityChecks=false.
```

The description is ready to use for creating a pull request.


cpetersen5 (approved):

kubecost/values.yaml
```diff
@@ -334,6 +335,7 @@ oidc:
   enabled: false
   clientID: ""  # Application client_id parameter obtained from provider. Used to make requests to server.
   clientSecret: ""  # Application/client client_secret parameter obtained from provider. Used to make requests to server.
+  authSecret: ""  # Value of the RBAC auth secret used with OIDC. If this is changed after initial deployment, a manual restart of the aggregator pod will be required. Set this explicitly when using CI/CD or GitOps templating. Example: `openssl rand -hex 32`
```
thomasvn, line 338: I believe only the SAML protocol requires the `authSecret`


```suggestion
```

thomasvn:
@jessegoodier @cpetersen5 Let's please hold off on merging this PR. I like the idea, but I think this can be much more cleanly handled by a backend change as opposed to a Helm change. https://github.com/kubecost/kubecost-cost-model/blob/v3.1.6/pkg/auth/samlauth.go#L34-L53

I'll put this on my TODO to investigate shortly.

jessegoodier:
holding off, Thomas has a more permanent fix.

jessegoodier:
fixing in aggregator code in 3.2

PR #4646, opened 2026-04-29
Actions RSA signing config

kaelanspatel:
## Release Notes Headline
Allow free and enterprise trial users to configure user-generated RSA signing key secrets for the aggregator and cluster controller, allowing for limited actions functionality.

## Commentary for Reviewers
Implements `.Values.kubecostProductConfigs.actions.signing.rsa` and `.Values.clusterController.signing.rsa` blocks, allowing the aggregator and cluster controller to be put into a mode which switches actions signing to use user-configured and generated RSA key secrets instead of the existing functionality, where actions could only be signed by a v2 enterprise license.

## Links to Issues or Tickets
https://apptio.atlassian.net/browse/KCM-5541

## User Impact and Risks
Low risk. Functionality is supported on the relevant backend images, and secrets are generated by the user; this PR only switches the `ACTIONS_USER_RSA_KEY` variable and mounts the secrets with the configured names to the appropriate component. 

## Testing
Tested via `helm template`, then used to deploy a testing environment, verifying correct mounting and ability of components to access mounted secret data.

## Documentation Updates
Forthcoming.


kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -132,6 +132,11 @@ spec:
             - key: smtp.json
               path: smtp.json
         {{- end }}
+        {{- if ((((.Values.kubecostProductConfigs).actions).signing).rsa).enabled }}
+        - name: rsa-private-key
+          secret:
+            secretName: {{ .Values.kubecostProductConfigs.actions.signing.rsa.privateKeySecret }}
```
nik-kc, line 138: I see `publicKeySecret` in `values.yaml`, but where is `privateKeySecret` defined?

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
```diff
@@ -116,6 +120,8 @@ spec:
           value: /var/keys/service-key.json
         - name: CC_LOG_LEVEL
           value: {{ .Values.clusterController.logLevel | default "info" }}
+        - name: ACTIONS_USER_RSA_KEY
```
nik-kc, line 123: Should this env var be gated behind `{{- if ((((.Values.kubecostProductConfigs).actions).signing).rsa).enabled }}` like it is in `aggregator-statefulset.yaml`?

kubecost/values.yaml
```diff
@@ -1372,6 +1372,11 @@ clusterController:
   #         "client_x509_cert_url": ""
   #       }
 
+  signing:
```
nik-kc, line 1390: Can you add comments to the `signing` blocks describing what they do?

kubecost/templates/aggregator/aggregator-statefulset.yaml
kaelanspatel: Oops, it should be under `kubecostProductConfigs` instead of `publicKeySecret`. Fixed.

kubecost/templates/cluster-controller/cluster-controller-deployment.yaml
kaelanspatel: It could be. I just figured it would be clearer to have it always set in the cluster controller, as there are fewer envs and, unlike the aggregator, you implicitly have actions enabled so it's more apparent how the cluster controller expects those actions to be signed, rather than more env bloat that doesn't mean anything unless enabled.

nik-kc: Ah, makes sense! I'm happy with this explanation.

kubecost/values.yaml
```diff
@@ -1477,6 +1487,15 @@ kubecostProductConfigs:
     #         "client_x509_cert_url": ""
     #       }
 
+    # If an enterprise v2 license is configured, actions will sign configuration files pushed to the store with that license
+    # If on free or trial-tier, or if self-management of signing credentials is desired, signing using a user-configured RSA key can be enabled
+
+    # If this is confiugred, the public key must also be configured on the cluster controller under .clusterController.signing.rsa
```
nik-kc: Small typo in "confiugred", otherwise this comment looks great.

nik-kc (approved):

PR #4647, opened 2026-04-30
use automated ingestion concurrency mechanism

ameijer:
## Release Notes Headline

ingestion concurrency now managed automatically 
## Commentary for Reviewers

rather than always requiring a single ingestion thread, we now scale it at 2x the number of cpu cores, which gives good performance without pegging CPU

## Links to Issues or Tickets

Related to https://github.com/kubecost/kubecost-cost-model/pull/4359

### DIFF
--- a/kubecost/templates/aggregator/aggregator-statefulset.yaml
+++ b/kubecost/templates/aggregator/aggregator-statefulset.yaml
@@ -560,8 +560,10 @@ spec:
               value: {{ .Values.aggregator.dbReadThreads | quote }}
             - name: DB_WRITE_THREADS
               value: {{ .Values.aggregator.dbWriteThreads | quote }}
+            {{- if ne .Values.aggregator.dbConcurrentIngestionCount nil }}
             - name: DB_CONCURRENT_INGESTION_COUNT
               value: {{ .Values.aggregator.dbConcurrentIngestionCount | quote }}
+            {{- end }}
             {{- if ne .Values.aggregator.dbMemoryLimit "0GB" }}
             - name: DB_MEMORY_LIMIT
               value: {{ .Values.aggregator.dbMemoryLimit | quote }}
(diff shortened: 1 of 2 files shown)


kubecost/values.yaml
```diff
@@ -1007,8 +1007,8 @@ aggregator:
   # How many threads to use when ingesting Asset/Allocation/CloudCost data
   # from the federated storage bucket. In most cases the default is sufficient,
   # but can be increased if trying to backfill historical data.
-  # default: 1
-  dbConcurrentIngestionCount: 1
+  # default: null (2x the current number of cores available to the process)
+  dbConcurrentIngestionCount: null
```
jessegoodier: how about auto?

kubecost/templates/aggregator/aggregator-statefulset.yaml
```diff
@@ -560,8 +560,10 @@ spec:
               value: {{ .Values.aggregator.dbReadThreads | quote }}
             - name: DB_WRITE_THREADS
               value: {{ .Values.aggregator.dbWriteThreads | quote }}
+            {{- if ne .Values.aggregator.dbConcurrentIngestionCount nil }}
```
thomasvn: Minor, but I'd prefer if the code were this

```suggestion
            {{- if .Values.aggregator.dbConcurrentIngestionCount }}
```

This checks to see if the config exists. I think this logic is used more widely around the chart and is easier to comprehend. Then in the `values.yaml` we can set `dbConcurrentIngestionCount: ""` where an empty string would indicate the config does not exist.

kubecost/values.yaml
ameijer: sure! I like that1

thomasvn (approved):
Looks good, we'll get to testing this in nightly environments.

PR #4663, opened 2026-05-11
KCM-5560 Add kubemodel flag for aggregator

aaliomer-ibm:
Add kubemodel flag (off by default) to enable the flag to be flipped in a future release.
Also, this will allow us to test the kubemodel when deploying with helm for integration tests

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -1082,7 +1082,9 @@ aggregator:
     failureThreshold: 200
 
   ## Set additional environment variables for the aggregator pod
-  extraEnv: []
+  extraEnv:
+    - name: KUBEMODEL_INGESTION_ENABLED
+      value: "false"
   # - name: SOME_VARIABLE
   #   value: "some_value"
 
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -1081,10 +1081,11 @@ aggregator:
     periodSeconds: 10
     failureThreshold: 200
 
+  ## Enable or disable kubemodel ingestion
+  kubemodelIngestionEnabled: false
+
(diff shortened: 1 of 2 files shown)


Sean-Holcomb (commented):
ExtraEnv is for user providing env vars. For this we would want to set the env var from new value in the aggregator block. Then template the value in the aggregators env vars directly.

kubecost/values.yaml
```diff
@@ -1082,7 +1082,9 @@ aggregator:
     failureThreshold: 200
 
   ## Set additional environment variables for the aggregator pod
-  extraEnv: []
+  extraEnv:
```
jessegoodier: would it make sense to just add the variable now instead of using entraEnv?

like this example:

```yaml
            - name: HEARTBEAT_ENABLED
               value: {{ .Values.heartbeat.enabled | quote }}
```

https://github.com/kubecost/kubecost/blob/4fe3bd60778a6812e59dc38f9d618e664daf50d1/kubecost/templates/aggregator/aggregator-statefulset.yaml#L540-L541

JesseGoodier (approved):

jessegoodier:
is this var in cost-model already? if not, what version are we targeting?
we have a customer asking for it in 3.1, but understand if we need to wait for 3.2.
@TrevorKWalker FYI

aaliomer-ibm:
> is this var in cost-model already? if not, what version are we targeting? we have a customer asking for it in 3.1, but understand if we need to wait for 3.2. @TrevorKWalker FYI

not merged yet. we need this variable to be able to test rollback and upgrades as well. hence why it is set to false. i cant speak of the release version

PR #4668, opened 2026-05-15
fix ECP set up via UI

nickcurie:
## Release Notes Headline
N/A

## Commentary for Reviewers
`enterpriseCustomPricing.configMapName` and `enterpriseCustomPricing.location.URI` were pre-populated, which would cause their values to be persisted even if the user only supplied `enterpriseCustomPricing.enabled = true` when intending to configure ECP via the UI.  This would result in mount failures since the user should not need to create a configmap via the k8s API, but the statefulset expects it anyway. This would lead to the aggregator pod to never starting.

## Links to Issues or Tickets
https://apptio.atlassian.net/browse/KCM-5776

## User Impact and Risks
Users will need to do slightly more set up in helm if they want ECP via helm. They will need to provide a configMapName and a location URI. No risks

## Testing
Tested on a replicated cluster. Performed a helm upgrade with the new values.yaml, when ECP is enabled aggregator pod spins up as expected, with no mount errors

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -276,7 +276,7 @@ enterpriseCustomPricing:
   # ConfigMap, for which we provide instructions, also below.
   enabled: false
   # To allow for setting the Enterprise Custom Pricing CSV via the UI, leave
-  # the configMapName and location fields commented out (or unset).
+  # the configMapName and location fields unset.
   #
   # Use the following command to create a ConfigMap from your pricing spec CSV.
   # You may change the ConfigMap name, or file name, as long as you set the
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -283,11 +283,11 @@ enterpriseCustomPricing:
   # correct configMapName and location.URI, respectively.
   #
   #  kubectl create configmap -n kubecost kubecost-enterprise-pricing --from-file pricing.csv
-  #
-  # configMapName: kubecost-enterprise-pricing
+  #             
+  configMapName: "" # e.g. kubecost-enterprise-pricing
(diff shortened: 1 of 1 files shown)


kubecost/values.yaml
```diff
@@ -275,19 +275,19 @@ enterpriseCustomPricing:
...
   #
   #  kubectl create configmap -n kubecost kubecost-enterprise-pricing --from-file pricing.csv
   #
-  configMapName: kubecost-enterprise-pricing
+  # configMapName: kubecost-enterprise-pricing
   # The file name (e.g. pricing.csv) needs to match the file name used to make the ConfigMap (above)
-  location:
-    URI: /var/configs/enterprise-pricing/pricing.csv
+  # location:
+  #   URI: /var/configs/enterprise-pricing/pricing.csv
```
thomasvn, line 291: @nickcurie Instead of commenting them out, what do you think about setting them to empty strings instead? This is the approach we try to use across the `values.yaml` to set an empty value.

```yaml
  configMapName: ""  # e.g. "kubecost-enterprise-pricing"
  location:
    URI: ""  # e.g. "/var/configs/enterprise-pricing/pricing.csv"
```

thomasvn:
Also, this PR now makes the user experience prioritize ECP configured via UI the default experience according to the Helm values. I'm ok with that! Just double checking with @chipzoller as well that this is ok.

nik-kc (approved):

chipzoller:
I think that's fine. Going the Helm route is for SREs anyway and having them input a couple of additional options is reasonable. Thanks for calling this out.

thomasvn (approved):

PR #4729, opened 2026-07-13
Fix: Add Helm labels to OpenShift SCC templates so they are properly tracked and deleted on uninstall

ErikTorres1998:
## Links to Issues or Tickets

https://github.com/kubecost/operators/pull/68


thomasvn (approved):
Thank you!

PR #4730, opened 2026-07-14
Change Aggregator.retention10m to 0

TrevorKWalker:
## Release Notes Headline

Changing Aggregator.retention10m = 0 to disable 10m files from being ingested into the aggregator by default

## Commentary for Reviewers

This is done to help improve efficiency of the aggregator and decrease resource usage especially for customers of scale. Large scale customers can expect to see large improvements in the aggregator's performance. This will also help to alleviate the number of clickhouse memory overcommit errors by proxy. 

## Links to Issues or Tickets

[https://apptio.atlassian.net/browse/KCM-5520](https://apptio.atlassian.net/browse/KCM-5520
)
[Slack Thread](https://kubecost.slack.com/archives/C0BDABS696X/p1782735733658409)
## User Impact and Risks

The Risks for this PR is that by default the request right sizing will not work with a view of last 48 hours. This is the only screen that currently uses the 10m data.

## Testing

This change has been a recommended configuration change for many large scale environments who have all seen a noticeable improvement in performance and data integrity.

## Documentation Updates

Section will need to be added to the Resource rightsizing page highlighting the need for changing retention 10m if they intend to use rightsizing with a last 48 hours view.

### DIFF
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -991,8 +991,9 @@ aggregator:
 
   ## Retention values determine for what duration into the past data will be
   ## ingested into Aggregator's database. 10m resolution data is only used for the
-  ## Node rightsizing and Resource Quota rightsizing features for further granularity.
-  ## Enabling 10m data will increase resource usage of the aggregator. A good starting value is 36.
+  ## Node Group Right-Sizing and Resource Quota Right-Sizing features for further
+  ## granularity. Enabling 10m data will increase resource usage of the Aggregator.
+  ## A good starting value would be `retention10m: 36`
   retention1d: 91
   retention1h: 49
   retention10m: 0
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -994,7 +994,7 @@ aggregator:
   retention1d: 91
   retention1h: 49
   # 10m files are currently only use for specific right sizing operations. Enabling 10m data will increase resource usage of the aggregator.
-  retention10m: 0 
+  retention10m: 0
 
   # Replicas sets the number of Aggregator replicas.
   replicas: 1
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -990,8 +990,8 @@ aggregator:
   imagePullPolicy: IfNotPresent
 
   ## Retention values determine for what duration into the past data will be
-  ## ingested into Aggregator's database. 10m resolution data is only used for the 
-  ## Node rightsizing and Resource Quota rightsizing features for further granularity. 
+  ## ingested into Aggregator's database. 10m resolution data is only used for the
+  ## Node rightsizing and Resource Quota rightsizing features for further granularity.
   ## Enabling 10m data will increase resource usage of the aggregator. A good starting value is 36.
   retention1d: 91
   retention1h: 49
--- a/kubecost/values.yaml
+++ b/kubecost/values.yaml
@@ -993,7 +993,8 @@ aggregator:
   ## ingested into Aggregator's database
   retention1d: 91
   retention1h: 49
-  retention10m: 36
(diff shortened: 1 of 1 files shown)


kubecost/values.yaml
```diff
@@ -993,7 +993,8 @@ aggregator:
   ## ingested into Aggregator's database
   retention1d: 91
   retention1h: 49
-  retention10m: 36
+  # 10m files are currently only use for specific right sizing operations. Enabling 10m data will increase resource usage of the aggregator.
+  retention10m: 0
```
thomasvn: Can we be super specific about what features leverage `retention10m`? I believe it is primarily node group sizing that uses this resolution data. We should double check in code.

Also a minor stylistic request but instead of a comment directly on top of retention10m, can it go alongside the top-level comment about retention?

```
  ## Retention values determine for what duration into the past data will be
  ## ingested into Aggregator's database. 10m resolution data is used for the 
  ## __ features. Enabling 10m data will increase resource usage of the
  ## aggregator. A good starting value is 36.
```

TrevorKWalker: The querier file for node right sizing and resource quotas has a section that can set the table to 10m data:
[node sizing Code Ref](https://github.com/kubecost/kubecost-cost-model/blob/1b2be1833093106496bd3199f65b7fcf375b3996/pkg/db/nodegroup/querier.go#L574)
[Resource quota code ref](https://github.com/kubecost/kubecost-cost-model/blob/1b2be1833093106496bd3199f65b7fcf375b3996/pkg/db/resourcequota/querier.go#L83)
However this already seems to only query based on the retention 10m value and will safely fall back to 1h if retention10m is set to zero. 

 There is also nodegroupsizingquerier.go which looks to always build from a 1d table but I believe that this is only used for the recommendations screen:
[code ref](https://github.com/kubecost/kubecost-cost-model/blob/1b2be1833093106496bd3199f65b7fcf375b3996/pkg/db/savings/nodegroupsizingquerier.go#L31)

The other references to 10m in costmodel seem to just be fore build the db tables and testing.

TrevorKWalker: I will update the comment and format to mention this.

thomasvn: Great, thanks for validating!

kubecost/values.yaml
```diff
@@ -990,10 +990,12 @@ aggregator:
   imagePullPolicy: IfNotPresent
 
   ## Retention values determine for what duration into the past data will be
-  ## ingested into Aggregator's database
+  ## ingested into Aggregator's database. 10m resolution data is only used for the
+  ## Node rightsizing and Resource Quota rightsizing features for further granularity.
+  ## Enabling 10m data will increase resource usage of the aggregator. A good starting value is 36.
```
thomasvn: Minor change to the comment. Overall LGTM!

```suggestion
  ## ingested into Aggregator's database. 10m resolution data is only used for the
  ## Node Group Right-Sizing and Resource Quota Right-Sizing features for further
  ## granularity. Enabling 10m data will increase resource usage of the Aggregator.
  ## A good starting value would be `retention10m: 36`
```

thomasvn (approved):

TrevorKWalker:
@thomasvn 
Do I need to fix the lint checker before squashing and merging? 
Seems like it is caught on a url that Jesse added three years ago so I assume its fine.

kubecost/values.yaml
thomasvn: @TrevorKWalker Thoughts this? I noticed you resolved this without leaving a comment here, or updating with the suggested change

thomasvn:
@TrevorKWalker We can fix the Link Checker CI tests in a separate PR. I just have one minor comment for you that I'd like your thoughts on. Then I think we're good to merge

kubecost/values.yaml
TrevorKWalker: My bad I thought resolving it would apply it

PR #4735, opened 2026-07-28
Fix "check_pr_labels" workflow

thomasvn:
## Release Notes Headline

N/A

## Commentary for Reviewers

Modeled after https://github.com/kubecost/kubecost-cost-model/blob/develop/.github/workflows/check_pr_labels.yaml.

## Links to Issues or Tickets

N/A

## User Impact and Risks

N/A

## Testing

N/A

## Documentation Updates

N/A

### DIFF
--- a/.github/workflows/check_pr_labels.yaml
+++ b/.github/workflows/check_pr_labels.yaml
@@ -1,29 +1,40 @@
 name: Label Checker
 on:
-    pull_request:
-        branches:
-            - develop
-        types:
-            - opened
-            - synchronize
-            - reopened
-            - labeled
-            - unlabeled
-    merge_group:
-      types: [checks_requested]
+  pull_request:
+    branches:
+      - develop
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
+  merge_group:
+    types: [checks_requested]
 
 jobs:
   check_labels:
     name: Check labels
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
+      - name: Generate an app token with permissions to clone the github-actions repo
+        id: generate-app-token
+        uses: actions/create-github-app-token@v3.2.0 # Immutable tag
+        with:
+          client-id: ${{ vars.KC_ACTIONS_BOT_APP_CLIENT_ID }}
+          private-key: ${{ secrets.KC_ACTIONS_BOT_APP_PRIVATE_KEY }}
+          repositories: ${{ github.event.repository.name }},github-actions
+
       - name: Check out actions code
         uses: actions/checkout@v6
         with:
           repository: kubecost/github-actions
           path: ./github-actions
           ref: main
-          token: ${{ secrets.GH_PAT }}
+          token: ${{ steps.generate-app-token.outputs.token }}
 
       - id: check-labels
         uses: ./github-actions/github-action-required-labels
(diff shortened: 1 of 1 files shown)


thomasvn:
Potentially no longer needed, as the fix is now included in https://github.com/kubecost/kubecost/pull/4734

.github/workflows/check_pr_labels.yaml
```diff
@@ -1,38 +1,51 @@
...
   check_labels:
     name: Check labels
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
+      - name: Generate an app token with permissions to clone the github-actions repo
+        id: generate-app-token
+        uses: actions/create-github-app-token@v3.2.0 # Immutable tag
```
nealormsbee, line 25: I actually don't think this is immutable. We'd want to get the SHA associated with that tag and use that for immutability.

.github/workflows/check_pr_labels.yaml
```diff
@@ -1,38 +1,51 @@
...
+    branches:
+      - develop
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
+  merge_group:
+    types: [checks_requested]
```
nealormsbee, line 13: Nice

.github/workflows/check_pr_labels.yaml
```diff
@@ -1,38 +1,51 @@
...
 
 jobs:
   check_labels:
     name: Check labels
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
```
nealormsbee, line 21: I don't think we need `pull-requests: write` since we're just checking for the presence of labels.

.github/workflows/check_pr_labels.yaml
```diff
@@ -1,38 +1,51 @@
...
     steps:
+      - name: Generate an app token with permissions to clone the github-actions repo
+        id: generate-app-token
+        uses: actions/create-github-app-token@v3.2.0 # Immutable tag
+        with:
+          client-id: ${{ vars.KC_ACTIONS_BOT_APP_CLIENT_ID }}
+          private-key: ${{ secrets.KC_ACTIONS_BOT_APP_PRIVATE_KEY }}
+          repositories: ${{ github.event.repository.name }},github-actions
```
nealormsbee, line 29: I think this can be scoped down to just the `github-actions` repository, since we don't even clone the repo being evaluated.

.github/workflows/check_pr_labels.yaml
```diff
@@ -1,38 +1,51 @@
...
+          token: ${{ steps.generate-app-token.outputs.token }}
 
       - id: check-labels
         uses: ./github-actions/github-action-required-labels
         with:
           mode: minimum
           count: 1
           add_comment: true
-          message: "This PR requires a version label that you want the PRd feature to be released in, in the form of vMAJOR.MINOR"
+          message:
```
nealormsbee, line 45: Does this need to be `message: |` since we're making it multiline? I actually don't know, since the _value_ itself isn't multiline, but we're writing it on the next line of yaml. Would need to run through `yq`

nealormsbee (changes requested):
I like the formatting fixes and the idea of using an immutable ref. I think a few other things could be more locked-down.

PR #4762, opened 2026-08-20
Add mcp-kubecost subchart

jessegoodier:
## Summary
- Adds the [mcp-kubecost](https://github.com/kubecost/mcp-kubecost) Helm chart as a subchart (`alias: mcp`, pinned to `0.6.2`) so the FinOps MCP server deploys with Kubecost.
- The MCP server is **enabled by default**. Existing installs will start running it on upgrade unless they set `mcp.enabled=false`. It talks to the in-cluster frontend Service via a tpl `kubecostBaseUrl` (`http://{{ .Release.Name }}-frontend:9090`).
- Adds a CI workflow that builds dependencies, lints/templates the chart, asserts subchart wiring (frontend URL, `global:` inheritance, OpenShift/CI-CD cases), and kubeconform-validates the rendered manifests.

## Test plan
- [x] `helm dependency build ./kubecost` succeeds and `Chart.lock` stays in sync with `Chart.yaml` (`mcp-kubecost` 0.6.2)
- [x] Default render includes MCP resources (`Deployment`/`Service`/`ConfigMap` named `<release>-mcp`):
  ```sh
  rm -rf ./kubecost/charts
  helm repo add mcp-kubecost https://kubecost.github.io/mcp-kubecost
  helm repo add finops-agent https://kubecost.github.io/finops-agent-chart/
  helm repo update
  helm dependency build ./kubecost
  helm template kubecost ./kubecost --skip-schema-validation | grep -E 'kind: (Deployment|Service|ConfigMap)|name: kubecost-mcp'
  ```
- [x] Opt-out works: `helm template kubecost ./kubecost --set mcp.enabled=false` does **not** render `charts/mcp/` resources
- [ ] MCP `KUBECOST_BASE_URL` matches the rendered frontend Service (`http://<release>-frontend:9090`) for both `kubecost` and an alternate release name
- [ ] `.github/scripts/validate_mcp_subchart.sh kubecost` passes locally
- [ ] GitHub Actions `Validate mcp-kubecost subchart` is green on this PR

## Notes for reviewers
- `Chart.yaml` still comments that the subchart is “Disabled by default”; `values.yaml` and the validator now default `mcp.enabled` to `true`. Worth aligning that comment.
- The MCP HTTP endpoint defaults to unauthenticated in-cluster (`authMode: none`). Auth should be configured before exposing via ingress / HTTPRoute.
- `helm template` uses `--skip-schema-validation` because the subchart schema uses `additionalProperties: false` and does not declare Helm’s injected `enabled` field.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,10 +1,5 @@
 # Agents.md
 
-## GitHub Actions
-
-- Never downgrade (or change) a GitHub Actions version without first verifying the version exists. Use `execute_command` to look it up: `curl -s https://api.github.com/repos/<owner>/<repo>/git/refs/tags | grep '"ref"' | grep '@v<N>'` or check the GitHub Marketplace page. Do not assume a version does not exist based on training knowledge alone.
-- when a GitHub Action is modified, verify that all actions in the workflow are current and use online lookups, do not depend on stale training data. If a new version is available, always suggest updating it.
-
 ## PR Reviews
 
 - When asked to review a PR, compare this current branch against the `develop` branch. `git diff develop`.
@@ -29,3 +24,7 @@ helm dependency build ./kubecost
 # Templating
 helm template kubecost ./kubecost
 ```
+
+## Additional rules
+
+Always check for rules that agents must follow in the [.bob/rules](.bob/rules) folder.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,9 @@
 # Agents.md
 
+## GitHub Actions
+
+- Never downgrade (or change) a GitHub Actions version without first verifying the version exists. Use `execute_command` to look it up: `curl -s https://api.github.com/repos/<owner>/<repo>/git/refs/tags | grep '"ref"' | grep '@v<N>'` or check the GitHub Marketplace page. Do not assume a version does not exist based on training knowledge alone.
+
 ## PR Reviews
 
 - When asked to review a PR, compare this current branch against the `develop` branch. `git diff develop`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,6 +3,7 @@
 ## GitHub Actions
 
 - Never downgrade (or change) a GitHub Actions version without first verifying the version exists. Use `execute_command` to look it up: `curl -s https://api.github.com/repos/<owner>/<repo>/git/refs/tags | grep '"ref"' | grep '@v<N>'` or check the GitHub Marketplace page. Do not assume a version does not exist based on training knowledge alone.
+- when a GitHub Action is modified, verify that all actions in the workflow are current and use online lookups, do not depend on stale training data. If a new version is available, always suggest updating it.
 
 ## PR Reviews
 
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -14,11 +14,11 @@ dependencies:
     version: "v1.0.25"
     repository: https://kubecost.github.io/finops-agent-chart
     condition: finopsagent.enabled
-  # The FinOps MCP server for Kubecost analytics. Disabled by default; the pinned
+  # The FinOps MCP server for Kubecost analytics. Enabled by default; the pinned
   # version is bumped automatically by the mcp-kubecost release pipeline.
   # Ref: https://github.com/kubecost/mcp-kubecost
   - name: mcp-kubecost
     alias: mcp
-    version: "0.6.2"
+    version: "0.6.3"
     repository: https://kubecost.github.io/mcp-kubecost
     condition: mcp.enabled
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -19,6 +19,6 @@ dependencies:
   # Ref: https://github.com/kubecost/mcp-kubecost
   - name: mcp-kubecost
     alias: mcp
-    version: "0.6.3"
+    version: "0.7.0"
     repository: https://kubecost.github.io/mcp-kubecost
     condition: mcp.enabled
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -19,6 +19,6 @@ dependencies:
   # Ref: https://github.com/kubecost/mcp-kubecost
   - name: mcp-kubecost
     alias: mcp
-    version: "0.6.1"
+    version: "0.6.2"
     repository: https://kubecost.github.io/mcp-kubecost
     condition: mcp.enabled
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -14,3 +14,11 @@ dependencies:
     version: "v1.0.25"
     repository: https://kubecost.github.io/finops-agent-chart
     condition: finopsagent.enabled
+  # The FinOps MCP server for Kubecost analytics. Disabled by default; the pinned
+  # version is bumped automatically by the mcp-kubecost release pipeline.
+  # Ref: https://github.com/kubecost/mcp-kubecost
+  - name: mcp-kubecost
+    alias: mcp
+    version: "0.6.1"
+    repository: https://kubecost.github.io/mcp-kubecost
+    condition: mcp.enabled
--- a/kubecost/Chart.yaml
+++ b/kubecost/Chart.yaml
@@ -19,6 +19,6 @@ dependencies:
   # Ref: https://github.com/kubecost/mcp-kubecost
   - name: mcp-kubecost
     alias: mcp
-    version: "0.7.0"
+    version: "0.9.0"
     repository: https://kubecost.github.io/mcp-kubecost
     condition: mcp.enabled
--- a/kubecost/templates/_helpers.tpl
+++ b/kubecost/templates/_helpers.tpl
@@ -600,6 +600,46 @@ To resolve this, either:
 {{- end -}}
 
 {{/*
+Path prefix the MCP server is served under on the Kubecost frontend host,
+derived from the path component of mcp.config.oidc.baseUrl. Empty when the MCP
+server has its own hostname (baseUrl at a host root) or when OIDC is off.
+
+When non-empty, FastMCP advertises {baseUrl}/authorize, {baseUrl}/token and
+{baseUrl}/register in its OAuth metadata, so every OAuth path the MCP clients
+touch lives under this prefix instead of the shared host root. The frontend
+nginx strips the prefix before proxying, and the subchart serves the MCP
+endpoint at "/" to match (mcp-kubecost.httpPath).
+
+Rejects characters that are not safe to interpolate into the nginx rewrite
+regexes that consume this value.
+*/}}
+{{- define "kubecost.mcp.pathPrefix" -}}
+{{- $raw := ((((.Values.mcp).config).oidc).baseUrl) | default "" | trim -}}
+{{- if $raw -}}
+{{- $path := (urlParse $raw).path | default "" -}}
+{{- $path = printf "/%s" (trimAll "/" $path) -}}
+{{- if ne $path "/" -}}
+{{- if not (regexMatch "^(/[A-Za-z0-9._~-]+)+$" $path) -}}
+{{- fail (printf "\n\nFAILURE [kubecost / mcp]: the path in mcp.config.oidc.baseUrl (%q) is not a usable nginx location prefix. Use only unreserved URL characters, e.g. https://kubecost.example.com/mcp\n" $path) -}}
+{{- end -}}
+{{- $path -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.kubecostApiBaseUrl" -}}
+{{- tpl (default (printf "http://%s-aggregator" .Release.Name) ((.Values.mcp).config).kubecostApiBaseUrl) . -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.kubecostApiPort" -}}
+{{- default 9004 ((.Values.mcp).config).kubecostApiPort -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.kubecostApiBasePath" -}}
+{{- default "/" ((.Values.mcp).config).kubecostApiBasePath -}}
+{{- end -}}
+
+{{/*
 Service name of the mcp-kubecost subchart. Mirrors mcp-kubecost.fullname so the
 frontend nginx upstream targets the same Service the subchart renders.
 */}}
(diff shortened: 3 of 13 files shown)


kubecost/values.yaml
```diff
@@ -654,6 +656,36 @@ finopsagent:
...
+## The FinOps MCP server for Kubecost analytics. Below are a subset of values from
+## the mcp-kubecost subchart that are most commonly changed.
+## Ref: https://github.com/kubecost/mcp-kubecost/blob/main/charts/mcp-kubecost/values.yaml
+##
+mcp:
+  enabled: true
+  ## The MCP server reads cost data through the Kubecost frontend Service
+  ## (`<release>-frontend:9090`). This value is tpl-evaluated by the subchart.
+  config:
+    kubecostBaseUrl: "http://{{ .Release.Name }}-frontend:9090"
```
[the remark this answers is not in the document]
jessegoodier: default is no ingress and kubecost defaults are open in general.

kubecost/templates/_helpers.tpl
```diff
@@ -575,6 +575,74 @@ To resolve this, either:
...
+{{- $authMode := include "kubecost.mcp.authMode" . -}}
+{{- if and (ne $authMode "oidc") (ne $authMode "both") (ne $authMode "api_key") }}
+{{- if and .Values.global.platforms.cicd.enabled .Values.global.platforms.cicd.skipSanityChecks }}
+
+WARNING: MCP.CONFIG.KUBECOSTAPIKEY IS SET BUT MCP.CONFIG.AUTHMODE IS NOT "OIDC", "BOTH", OR "API_KEY". THIS IS
+NOT SECURE: THE MCP SERVER HOLDS A KUBECOST API KEY WHILE THE MCP HTTP ENDPOINT IS NOT PROTECTED.
+SET MCP.CONFIG.AUTHMODE TO OIDC, BOTH, OR API_KEY, OR CLEAR THE API KEY.
+SKIPSANITYCHECKS IS TRUE SO THIS CHECK DID NOT FAIL.
+{{- else }}
+{{- fail "\n\nFAILURE: mcp.config.kubecostApiKey is set but mcp.config.authMode is not \"oidc\", \"both\", or \"api_key\". This is not secure: the MCP server holds a Kubecost API key while the MCP HTTP endpoint is not protected. Set mcp.config.authMode to oidc, both, or api_key, or clear the API key\n" }}
```
jessegoodier: This failure is the biggest risk for users that don't fully understand the auth options.

Goal with the options is to support configs where they don't want to use API Keys and prefer to hard code the readonly service account in the mcp env vars and have the users authenticate with the SSO option via oauth.

kubecost/values.yaml
```diff
@@ -654,6 +656,36 @@ finopsagent:
       spotLabel: ""
       spotLabelValue: ""
 
+## The FinOps MCP server for Kubecost analytics. Below are a subset of values from
+## the mcp-kubecost subchart that are most commonly changed.
+## Ref: https://github.com/kubecost/mcp-kubecost/blob/main/charts/mcp-kubecost/values.yaml
+##
+mcp:
```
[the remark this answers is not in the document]
jessegoodier: Bob was replying to some mismatch agents/readme concerns. Per Chip, we want to enable by default. Nothing breaks.

AGENTS.md
```diff
@@ -1,5 +1,10 @@
 # Agents.md
 
+## GitHub Actions
+
+- Never downgrade (or change) a GitHub Actions version without first verifying the version exists. Use `execute_command` to look it up: `curl -s https://api.github.com/repos/<owner>/<repo>/git/refs/tags | grep '"ref"' | grep '@v<N>'` or check the GitHub Marketplace page. Do not assume a version does not exist based on training knowledge alone.
```
jessegoodier: Bob's training data is from last year sometime. Trying to avoid  future frustration.

kubecost/values.yaml
```diff
@@ -654,6 +656,35 @@ finopsagent:
...
+  enabled: true
+  ## The MCP server reads cost data through the Kubecost frontend Service
+  ## (`<release>-frontend:9090`). This value is tpl-evaluated by the subchart.
+  config:
+    kubecostBaseUrl: "http://{{ .Release.Name }}-frontend:9090"
+    kubecostApiBasePath: /model
+    ## Authentication for the MCP HTTP endpoint. One of: none, oidc, api_key, both.
+    ## The default of "none" leaves the endpoint unauthenticated inside the cluster;
+    ## set mcp.config.authMode before exposing it through ingress.hosts or httpRoute.
+    authMode: none
```
thomasvn: This may warrant a bigger discussion beyond the scope of this PR, but would it be better for the MCP server to query the Aggregator directly, instead of going through the frontend? That would help resolve some of these authMode issues?

AGENTS.md
thomasvn: Nice, thank you!

kubecost/Chart.yaml
```diff
@@ -14,3 +14,11 @@ dependencies:
     version: "v1.0.25"
     repository: https://kubecost.github.io/finops-agent-chart
     condition: finopsagent.enabled
+  # The FinOps MCP server for Kubecost analytics. Enabled by default; the pinned
+  # version is bumped automatically by the mcp-kubecost release pipeline.
+  # Ref: https://github.com/kubecost/mcp-kubecost
+  - name: mcp-kubecost
+    alias: mcp
+    version: "0.6.3"
```
thomasvn: I believe this needs to be bumped to "0.6.4", to align with the updates to `.Values.config.authMode` https://github.com/kubecost/mcp-kubecost/releases/tag/v0.6.4

kubecost/values.yaml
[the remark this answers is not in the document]
thomasvn: Agree! Enabled by default should be ok.

kubecost/templates/_helpers.tpl
```diff
@@ -575,6 +575,74 @@ To resolve this, either:
...
+{{- if ((.Values.mcp).config).requireClientApiKey }}
+{{- printf "true" -}}
+{{- else -}}
+{{- printf "false" -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.redirectPath" -}}
+{{- default "/auth-mcp" (((.Values.mcp).config).oidc).redirectPath -}}
+{{- end -}}
```
thomasvn, line 600: Probably best not to set any defaults in the parent chart? To reduce confusion? I'm thinking it would be best to inherit whatever the default is of the subchart. Same goes for `kubecost.mcp.authMode`

thomasvn: Same for `kubecost.mcp.servicePort`?

kubecost/templates/frontend/frontend-configmap.yaml
```diff
@@ -484,6 +498,33 @@ data:
...
+            {{- include "kubecost.frontend.mcpProxyDirectives" . | nindent 12 }}
+        }
+
+        location ^~ /.well-known/oauth-protected-resource {
+            {{- include "kubecost.frontend.mcpProxyDirectives" . | nindent 12 }}
+        }
+
+        location ^~ /.well-known/oauth-authorization-server {
+            {{- include "kubecost.frontend.mcpProxyDirectives" . | nindent 12 }}
+        }
```
thomasvn, line 592: I might need more help understanding the auth-related pieces of this PR. Maybe we can talk this over next week!

kubecost/Chart.yaml
jessegoodier: yes. mcp will also automatically maintain the develop branch when new versions are released.

probably auto tag cherry-picking in the future.

kubecost/templates/_helpers.tpl
jessegoodier: agree on redirectPath
mcp.servicePort is used for the frontend configmap

kubecost/templates/frontend/frontend-configmap.yaml
jessegoodier: yes, please!

jessegoodier:
**Added in latest commit ([b31ea35](https://github.com/kubecost/kubecost/commit/b31ea35b)):** remove `authMode=both`

Removes all references to the `both` auth mode from the parent chart, in sync with [mcp-kubecost#60](https://github.com/kubecost/mcp-kubecost/pull/60).

**Changes:**
- `kubecost/templates/_helpers.tpl` — `apiKeyAuthCheck`: removed `(ne $authMode "both")` condition + all warning/fail message references; `openRouteCheck`: updated warning/fail messages to drop `"both"`/`"BOTH"`
- `kubecost/values.yaml` — updated `mcp.config.authMode` comment to list valid values (`none`, `open`, `oidc`, `api_key`) with migration hint

**Migration:** `authMode: both` → `authMode: oidc` + `requireClientApiKey: true`

kubecost/templates/_helpers.tpl
```diff
@@ -575,6 +575,179 @@ To resolve this, either:
...
+{{/*
+Fail when authMode is "none" and either an MCP route (httpRoute or ingress) is
+enabled, or kubecostApiPort is 9008 (aggregator SSO bypass). In both cases the
+MCP endpoint would be unauthenticated while able to reach Kubecost without
+SSO. When CI/CD skipSanityChecks is set, emit a warning instead.
+This does not prevent an mcp.httpRoute or mcp.ingress from being enabled. Those checks exist in the sub-chart itself.
+*/}}
+{{- define "kubecost.mcp.openRouteCheck" -}}
+{{- if (.Values.mcp).enabled }}
+{{- $routeEnabled := or ((.Values.mcp).httpRoute).enabled ((.Values.mcp).ingress).enabled (.Values.ingress).enabled (.Values.httpRoute).enabled -}}
```
thomasvn, line 675: According to this logic, any user who has `ingress.enabled: true` will fail upon `helm upgrade` to the next version. `mcp.enabled` defaults to `true` and `mcp.config.authMode` defaults to `none`

```gotemplate
{{- $routeEnabled := or ((.Values.mcp).httpRoute).enabled ((.Values.mcp).ingress).enabled
                        (.Values.ingress).enabled (.Values.httpRoute).enabled -}}
```

So an install that has nothing to do with MCP now hard-fails:

```console
$ helm template kubecost ./kubecost --set ingress.enabled=true
Error: execution error at (kubecost/templates/NOTES.txt:50:4):
FAILURE: mcp.config.authMode is "none" while mcp.httpRoute or mcp.ingress is enabled, ...
```

kubecost/values.yaml
```diff
@@ -654,6 +656,66 @@ finopsagent:
...
+      clientSecret: ""
+      # mcp.config.oidc.existingSecret references a pre-created Secret containing OIDC credentials.
+      # Required keys: OIDC_CLIENT_ID, OIDC_CLIENT_SECRET. Preferred over clientId/clientSecret.
+      existingSecret: ""
+  ## The MCP server is reachable in-cluster at <release>-mcp:3030. Enable
+  ## httpRoute (Gateway API) or ingress to expose it outside the cluster.
+  httpRoute:
+    enabled: false
+  ingress:
+    enabled: false
```
thomasvn, line 710: Thinking out loud here.

If an ingress or httpRoute is enabled, that means the Kubecost MCP server is (likely) accessible outside the cluster ([ref](https://github.com/kubecost/mcp-kubecost/blob/main/charts/mcp-kubecost/templates/ingress.yaml)). If that's the case, why do we need all the extra Kubecost Frontend Nginx logic when `.Values.mcp.ingress.enabled=true`? The user _shouldn't_ be hitting the Kubecost Frontend to get to the MCP server. They should instead just be hitting the MCP Server's Ingress endpiont?

Let me know what you think!

kubecost/templates/frontend/_helpers.tpl
```diff
@@ -87,4 +87,38 @@ Merge bufferConfig.directives with extraServerConfig lines, render nginx directi
...
+When mcp.config.authMode is "open" the operator has explicitly acknowledged
+unauthenticated exposure — requests are proxied normally without auth enforcement.
+When mcp.config.authMode is "oidc/api_key" the endpoint is proxied with auth enforcement being handled by the MCP backend.
+This does not prevent an mcp.httpRoute or mcp.ingress from being enabled. Those checks exist in the sub-chart itself.
+*/}}
+{{- define "kubecost.frontend.mcpProxyDirectives" -}}
+{{- if and (eq (include "kubecost.mcp.authMode" .) "none") (or .Values.httpRoute.enabled .Values.ingress.enabled) -}}
+  add_header Content-Type text/plain;
+  return 424 "MCP endpoint is not configured. Set mcp.config.authMode to enable access.";
+{{- else -}}
```
thomasvn: This would also break users who `helm upgrade` to v3.3, but have `.Values.ingress.enabled=true`

kubecost/templates/_helpers.tpl
jessegoodier: Thank you for this! 
I was making a huge logic change at the time.

Intent was to block if using 9008 AND ingress is enabled. because 9008 is not default, this is just preventing the user from inadvertently exposing the unprotected port.

kubecost/templates/frontend/_helpers.tpl
jessegoodier: Only disables the MCP, doesn't break kubecost itself though, right?

kubecost/values.yaml
jessegoodier: good call!
will disable /mcp when mpc.(ingress/httproute) is enabled.

kubecost/templates/frontend/_helpers.tpl
jessegoodier: Actually, updated the logic based on this.

MCP enabled unless the user sets it to port 9008. 

Which means the MCP enabled by default in for any configuration a user would be upgrading from.

kubecost/templates/_helpers.tpl
jessegoodier: done. see latest commit.

jessegoodier:
tests look good.
EKS test will fail until the image is copied. the image will copy after it is in the helm chart automatically.

https://github.com/kubecost/kubecost/actions/runs/32988452089
image

jessegoodier:
I'll fix the linkchecker in the 3.3 docs PR

thomasvn:
@jessegoodier Just confirming my understanding of the new implementation!

- MCP server will be on by default in Kubecost installs.
- If Kubecost SSO is enabled, the default deployed MCP server will not yet be functional. It will be querying Aggregator port 9004 which requires authentication.
- To get the MCP server working, set up auth for the MCP server. Or acknowledge that the MCP server would be exposed by setting authmode="open".

kubecost/templates/_helpers.tpl
```diff
@@ -575,6 +575,182 @@ To resolve this, either:
...
+{{- define "kubecost.mcp.requireClientApiKey" }}
+{{- if ((.Values.mcp).config).requireClientApiKey }}
+{{- printf "true" -}}
+{{- else -}}
+{{- printf "false" -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.redirectPath" -}}
+{{- default "/auth-mcp" (((.Values.mcp).config).oidc).redirectPath -}}
```
thomasvn, line 599: Below is a valid finding ([ref](https://github.com/kubecost/mcp-kubecost/blob/v0.9.0/charts/mcp-kubecost/templates/_helpers.tpl#L133))

More importantly though, do we really even need to define these functions? And if we do need the function, is there any way we can call the subchart's functions?

## `redirectPath` is not normalized, unlike the subchart's copy

```gotemplate
{{- define "kubecost.mcp.redirectPath" -}}
{{- default "/auth-mcp" (((.Values.mcp).config).oidc).redirectPath -}}
{{- end -}}
```

The subchart's `mcp-kubecost.oidcRedirectPath` normalizes with `printf "/%s" (trimAll "/" $raw)`;
the parent uses the raw value. Reproduced with `--set 'mcp.config.oidc.redirectPath=auth-mcp/'`:

* pod ConfigMap: `OIDC_REDIRECT_PATH: /auth-mcp`
* frontend nginx: `location auth-mcp/ {` — a prefix that no request URI can ever match
* `productConfigs`: `"mcpRedirectPath": "auth-mcp/"`

The IdP then redirects to `/auth-mcp`, which falls into `location /auth` (Kubecost's
`auth_request` proxy to the aggregator) and MCP OIDC login breaks with a confusing error.
Reuse the subchart's normalization (`printf "/%s" (trimAll "/" ...)`).

kubecost/templates/_helpers.tpl
```diff
@@ -575,6 +575,182 @@ To resolve this, either:
...
+{{- else -}}
+{{- printf "false" -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.authMode" -}}
+{{- default "none" ((.Values.mcp).config).authMode -}}
+{{- end -}}
+
+{{- define "kubecost.mcp.requireClientApiKey" }}
```
thomasvn, line 590: Again, looks like there is a disconnect between here and the subchart https://github.com/kubecost/mcp-kubecost/blob/v0.9.0/charts/mcp-kubecost/templates/_helpers.tpl#L295

kubecost/templates/frontend/frontend-configmap.yaml
```diff
@@ -484,6 +498,102 @@ data:
...
+        {{- else }}
+        # mcpHostRootMode: no path prefix, so the subchart serves the MCP
+        # endpoint at /mcp (config.http.path) and every OAuth route at the
+        # container root. Paths pass through unrewritten.
+        # MCP streamable HTTP — FastMCP serves at /mcp on port {{ include "kubecost.mcp.servicePort" . }}
+        location /mcp {
+            {{- include "kubecost.frontend.mcpProxyDirectives" . | nindent 12 }}
+        }
```
thomasvn, line 568: This is a valid point, but lower priority for now

## `mcpEnabled: "true"` reported when the frontend proxies nothing

`mcpEnabled` is derived from `mcp.enabled` alone, but the `/mcp` location block is gated on `and mcp.enabled (not mcp.httpRoute.enabled) (not mcp.ingress.enabled)`. Reproduced with `--set mcp.httpRoute.enabled=true --set mcp.config.authMode=open`: nginx has **no** `location /mcp` yet `productConfigs` still says `"mcpEnabled": "true"`. A same-origin `/mcp` request then falls through to `location /` and returns the SPA's `index.html` with HTTP 200 — the client sees HTML where it expects an MCP/JSON response, which is harder to diagnose than a 404.

Either report `mcpEnabled` using the same three-way condition, or surface the external MCP URL so the UI knows to go elsewhere.

kubecost/templates/frontend/frontend-configmap.yaml
```diff
@@ -484,6 +498,102 @@ data:
...
+        location ^~ /.well-known/oauth-authorization-server{{ $mcpPrefix }} {
+            rewrite ^ /.well-known/oauth-authorization-server break;
+            {{- include "kubecost.frontend.mcpProxyDirectives" . | nindent 12 }}
+        }
+        {{- else }}
+        # mcpHostRootMode: no path prefix, so the subchart serves the MCP
+        # endpoint at /mcp (config.http.path) and every OAuth route at the
+        # container root. Paths pass through unrewritten.
+        # MCP streamable HTTP — FastMCP serves at /mcp on port {{ include "kubecost.mcp.servicePort" . }}
+        location /mcp {
```
thomasvn, line 566: I think this is a valid finding, but low priority. We should put a TODO to fix though.

## host-root mode hardcodes `/mcp`, ignoring `mcp.config.http.path`

The subchart honours `config.http.path` (`mcp-kubecost.httpPath`), but the parent's host-root nginx block hardcodes `location /mcp` and `location ^~ /.well-known/oauth-protected-resource/mcp` and does not rewrite. Setting `mcp.config.http.path: /foo` leaves the container serving `/foo` while nginx forwards `/mcp` unchanged → 404 with no chart-side error. Either read `mcp-kubecost.httpPath` or fail when `config.http.path` is set to something other than `/mcp`.

kubecost/templates/_helpers.tpl
```diff
@@ -575,6 +575,182 @@ To resolve this, either:
...
+*/}}
+{{- define "kubecost.mcp.pathPrefix" -}}
+{{- if eq (include "kubecost.mcp.authMode" .) "oidc" -}}
+{{- $raw := ((((.Values.mcp).config).oidc).baseUrl) | default "" | trim -}}
+{{- if $raw -}}
+{{- $path := (urlParse $raw).path | default "" -}}
+{{- $path = printf "/%s" (trimAll "/" $path) -}}
+{{- if ne $path "/" -}}
+{{- if not (regexMatch "^(/[A-Za-z0-9._~-]+)+$" $path) -}}
+{{- fail (printf "\n\nFAILURE [kubecost / mcp]: the path in mcp.config.oidc.baseUrl (%q) is not a usable nginx location prefix. Use only unreserved URL characters, e.g. https://kubecost.example.com/mcp\n" $path) -}}
```
thomasvn, line 624: Valid low priority finding

## `pathPrefix` can abort the whole render even with MCP disabled

`kubecost.mcp.pathPrefix` is invoked unconditionally from `productConfigs`, and its own guard is `authMode == "oidc"` — not `mcp.enabled`. Reproduced:

```
--set mcp.enabled=false --set mcp.config.authMode=oidc \
  --set 'mcp.config.oidc.baseUrl=https://k.example.com/my%20mcp'
Error: execution error at (kubecost/templates/local-store/cluster-storage-deployment.yaml:37:29):
FAILURE [kubecost / mcp]: the path in mcp.config.oidc.baseUrl ("/my mcp") is not a usable nginx location prefix
```

A stale MCP value blocks an install that has MCP turned off, and Helm attributes the error to an unrelated template (the config-checksum helper pulls in the nginx ConfigMap), making it hard to diagnose. Gate the helper on `(.Values.mcp).enabled`.

thomasvn:
Lots of small findings that we definitely should come back and fix. Overall I'm in agreement with the direction here though! Let's get this first big PR in, then make incremental improvements.

thomasvn (approved):

jessegoodier:
> @jessegoodier Just confirming my understanding of the new implementation!
> 
> * MCP server will be on by default in Kubecost installs.
> * If Kubecost SSO is enabled, the default deployed MCP server will not yet be functional. It will be querying Aggregator port 9004 which requires authentication.
> * To get the MCP server working, set up auth for the MCP server. Or acknowledge that the MCP server would be exposed by setting authmode="open".

yes. when SSO is enabled, without any mcp setting, the user would get:
```
{
  "status": "error",
  "message": "[data_unavailable] Could not connect to the Kubecost API. The service may be down or the base URL may be incorrect. (retryable=true) Action: Verify KUBECOST_BASE_URL is reachable and the Kubecost service is running.",
  "recommended_action": "Check Kubecost connectivity and credentials, then retry.",
  "categories": [],
  "total_savings_per_month": 0.0,
  "category_count": 0
}
```

jessegoodier:
> Lots of small findings that we definitely should come back and fix. Overall I'm in agreement with the direction here though! Let's get this first big PR in, then make incremental improvements.

nice finds! I'll have a PR up tomorrow AM to fix.

PR #4797, opened 2026-09-09
feat: disable MCP subchart when aggregator.enabled=false

jessegoodier:
## Summary

Changes the MCP subchart's default behaviour so it follows `aggregator.enabled` instead of always being on. Agent-only installs (`aggregator.enabled=false`) no longer pick up the MCP server on upgrade without any user action.

### How it works

`Chart.yaml` sets a two-path condition:

```yaml
condition: mcp.enabled,aggregator.enabled
```

Helm's first-boolean-wins rule means:
- `mcp.enabled` is **omitted** from `values.yaml` → condition falls through to `aggregator.enabled`
- `aggregator.enabled: true` (default) → MCP renders as before
- `aggregator.enabled: false` → MCP is skipped
- `mcp.enabled: true` → MCP always renders (explicit override, e.g. external aggregator)
- `mcp.enabled: false` → MCP always skipped (explicit override)

The `kubecost.mcp.enabled` helper in `_helpers.tpl` mirrors this logic so nginx and NOTES.txt stay in sync with the subchart condition.

### Scenarios validated

| Scenario | MCP renders? | `mcpEnabled` in nginx |
|---|---|---|
| defaults (`aggregator.enabled=true`) | ✅ yes | `true` |
| `aggregator.enabled=false`, `mcp.enabled` unset | ✅ no | `false` |
| `aggregator.enabled=false`, `mcp.enabled=true` | ✅ yes | `true` |
| `aggregator.enabled=true`, `mcp.enabled=false` | ✅ no | `false` |

`helm lint` passes. `aggregatorCheck` warning fires correctly when `mcp.enabled=true` + `aggregator.enabled=false` with default URL.

### Files changed

| File | Change |
|---|---|
| `kubecost/Chart.yaml` | Two-path condition `mcp.enabled,aggregator.enabled`; updated comment |
| `kubecost/values.yaml` | Remove `mcp.enabled: true`; replace with explanatory comment |
| `kubecost/templates/_helpers.tpl` | `kubecost.mcp.enabled` mirrors Chart.yaml condition; all callers updated |
| `kubecost/templates/frontend/frontend-configmap.yaml` | Use helper instead of raw `.Values.mcp.enabled` |
| `.github/scripts/validate_mcp_subchart.sh` | Update assertions + add `aggregator-off` and `mcp-forced-on` test scenarios |

### DIFF
--- a/.github/scripts/validate_mcp_subchart.sh
+++ b/.github/scripts/validate_mcp_subchart.sh
@@ -177,7 +177,7 @@ helm lint "$CHART_DIR" "${skip_schema[@]}" --set "${values_key}.enabled=true"
 pass "helm lint (--set ${values_key}.enabled=true)"
 
 # ---------------------------------------------------------------------------
-group "Subchart is enabled by default"
+group "Default-render behaviour (follows aggregator.enabled)"
 
 helm template "$RELEASE_NAME" "$CHART_DIR" "${skip_schema[@]}" > "${RENDER_DIR}/default.yaml"
 pass "helm template (defaults) rendered without errors"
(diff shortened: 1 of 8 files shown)


thomasvn (approved):
Clever!

PR #4798, opened 2026-09-09
CP: disable MCP subchart when aggregator.enabled=false (#4797)

jessegoodier:
manual cp for failed cherry-pick bot

* Follow aggregator.enabled when mcp.enabled is unset

Helm cannot use templates as a subchart condition. Omitting mcp.enabled lets Chart.yaml fall through to aggregator.enabled so agent-only upgrades skip MCP without a new flag.

* fix: review follow-ups for mcp/follow-aggregator-enabled

- Rename validate script section header from 'Subchart is enabled by default' to 'Default-render behaviour (follows aggregator.enabled)' to reflect the new conditional-enable semantics.

- Add mcpEnabled=true and upstream mcpKubecost assertions to the mcp-forced-on test scenario (mirrors the existing aggregator-off assertions for symmetry).

- Normalize kubecost.mcp.enabled helper: collapse the fallthrough branches to inline style consistent with the explicit-bool branches. Functionally identical; trim markers already stripped the newlines.

* fix chart lock

* ci: add kubeconform renders for aggregator-follows-mcp scenarios

Add four kubeconform renders to validate-mcp-subchart.yml to cover all toggle combinations introduced by the mcp.enabled,aggregator.enabled condition:

- mcp-default: no mcp.enabled set, aggregator.enabled=true (default) — the MCP-follows-aggregator path that was never schema-validated before
- mcp-enabled: explicit mcp.enabled=true (existing backwards-compat path, now renamed for clarity)
- aggregator-off: aggregator.enabled=false, mcp.enabled unset — ensures the chart without MCP resources still produces valid K8s YAML
- mcp-forced-on: aggregator.enabled=false, mcp.enabled=true — the external-aggregator override path

Each render gets its own kubeconform step so failures are scoped to the exact scenario that broke.

* chore: update .gitignore to include .helm-render directory

### DIFF
--- a/.github/scripts/validate_mcp_subchart.sh
+++ b/.github/scripts/validate_mcp_subchart.sh
@@ -123,8 +123,8 @@ values_key="${dep_alias:-$SUBCHART_NAME}"
 
 assert_eq "dependency repository is the mcp-kubecost Helm repo" \
   "https://kubecost.github.io/mcp-kubecost" "$dep_repository"
-assert_eq "dependency condition gates the subchart" \
-  "${values_key}.enabled" "$dep_condition"
+assert_eq "dependency condition falls through to aggregator.enabled" \
+  "${values_key}.enabled,aggregator.enabled" "$dep_condition"
 
 if [ "$dep_version" = "*" ] || [ "${dep_version#\^}" != "$dep_version" ] \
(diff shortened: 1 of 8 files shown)


thomasvn:
Cherry-pick of https://github.com/kubecost/kubecost/pull/4797

thomasvn (approved):

PR #4812, opened 2026-09-11
fix pr backport checker script

jessegoodier:
use pinned uv version
fix slack message formatting


jessegoodier:
merging as this is fixing a bug and only impact is the checker job.

PR #4825, opened 2026-09-15
fix(network-costs): let additionalSecurityContext override securityContext defaults

zoispag:
## Release Notes Headline

`networkCosts.additionalSecurityContext` now overrides the container's default `privileged` and `runAsUser` values instead of emitting a duplicate YAML key.

## Commentary for Reviewers

`network-costs-daemonset.yaml` hardcodes the container security context and then appends `additionalSecurityContext` underneath it:

```yaml
        securityContext:
          privileged: true  # Required
          runAsUser: 0  # Required
        {{- if .Values.networkCosts.additionalSecurityContext }}
        {{- toYaml .Values.networkCosts.additionalSecurityContext | nindent 10 }}
        {{- end }}
```

Any key the user sets that collides with a default is therefore emitted twice. Setting `privileged: false` renders:

```yaml
        securityContext:
          privileged: true  # Required
          runAsUser: 0  # Required
          privileged: false     # <-- duplicate key
```

Duplicate mapping keys are invalid under the YAML 1.2 spec. In practice the last one currently wins in both `kubectl` and the API server, so the value does take effect, but it relies on lenient parsing and `yamllint` rejects the rendered manifest outright. It also makes the value silently order-dependent on the template.

This PR merges the user map over the defaults rather than appending it, so a colliding key overrides cleanly and non-colliding keys behave exactly as before.

### Why a user would want to override `privileged`

`network-costs` reads conntrack flows over the netlink (ctnetlink) API, which the kernel gates behind `CAP_NET_ADMIN` — not behind full `privileged`. I verified this on a live 6.18 node by running `network-costs:v0.19.1` unchanged apart from the security context:

| Security context | Result |
|---|---|
| `privileged: true` (chart default) | 15,330 parsed conntrack entries |
| `privileged: false`, `drop: [ALL]` | **0 entries**, `Failed to parse conntrack flows: netlink error ... error: -1` every 5s |
| `privileged: false`, `drop: [ALL]`, `add: [NET_ADMIN]` | 15,340 parsed entries, no errors |

So `privileged: true` is stricter than necessary, and users on clusters with Pod Security Standards or an OPA/Kyverno policy against privileged containers can run the DaemonSet with just `CAP_NET_ADMIN`.

Worth noting for anyone who finds this via the security scanners that flag this DaemonSet: **dropping `privileged` without adding `NET_ADMIN` fails silently.** The pod stays `Ready`, `/metrics` keeps serving, and the only symptom is `kubecost_network_costs_parsed_entries` sitting at `0`. I've documented the working combination in `values.yaml` so people don't land on the broken half of that.

I deliberately did **not** change the default — that is a behavioural change for every user and belongs in its own PR if you want it. Happy to follow up with one if you'd like.

## Links to Issues or Tickets

N/A

## User Impact and Risks

Low. With default values the rendered chart is byte-identical apart from two dropped inline comments (`# Required`), whose content is now documented in `values.yaml`:

```console
$ diff <(helm template kubecost ./kubecost)  # develop
        <(helm template kubecost ./kubecost)  # this branch
1222,1223c1222,1223
<           privileged: true  # Required
<           runAsUser: 0  # Required
---
>           privileged: true
>           runAsUser: 0
```

Users who already set `additionalSecurityContext` with a *non*-colliding key (the only usage that was well-defined before) see no change. Users who set a *colliding* key get the same effective value they get today, just without the duplicate key.

## Testing

Per `AGENTS.md`, `helm template` before and after, full chart with subcharts built.

**BEFORE — default values**
```yaml
        securityContext:
          privileged: true  # Required
          runAsUser: 0  # Required
```

**AFTER — default values**
```yaml
        securityContext:
          privileged: true
          runAsUser: 0
```

**BEFORE — `additionalSecurityContext` overriding `privileged`**
```yaml
        securityContext:
          privileged: true  # Required
          runAsUser: 0  # Required
          capabilities:
            add:
            - NET_ADMIN
            drop:
            - ALL
          privileged: false
```

**AFTER — same values**
```yaml
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
            drop:
            - ALL
          privileged: false
          runAsUser: 0
```

Also checked:

- `helm lint ./kubecost` passes with both default and overriding values.
- `kubeconform -strict` validates the rendered DaemonSet.
- `yamllint` with `key-duplicates` enabled flags the BEFORE render (`error duplication of key "privileged" in mapping`) and passes clean on AFTER.
- Confirmed `mergeOverwrite` propagates a boolean `false` from the user map rather than skipping it as an empty value.
- Functionally verified on a live EKS cluster (Kubernetes 1.34, kernel 6.18, Amazon Linux 2023) as described above.

## Documentation Updates

Documented inline in `kubecost/values.yaml`, including the note that `runAsUser: 0` is still required for the conntrack accounting sysctl and for `hostProc` mode.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jessegoodier:
This is a great call out @zoispag! thanks.

See my changes. Are you good with this?

notes:
3.2.x defaults were:

```yaml
securityContext:
  privileged: true
  runAsUser: 0
```

and was not possible to change it.

new defaults are least privileged and now can be modified.

```yaml
  securityContext:
    privileged: false
    runAsUser: 0
    capabilities:
      drop: ["ALL"]
      add: ["NET_ADMIN"]
```

This is two breaking changes:
1. removes additionalSecurityContext
2. changes privileged: from true to false

I'm going to defer to @thomasvn on approval and version. 

I tested this in a live eks cluster and network costs logs are clean.

Tests removing `runAsUser: 0` does fail with the logs:

```
ERROR network_costs_rs::conntrack::tracer: Failed to parse conntrack flows: netlink error: Nlmsgerr(Nlmsgerr { error: -1, nlmsg: NlmsghdrErr { nl_len: 20, nl_type: Conntrack, nl_flags: NlmF(769), nl_seq: 3, nl_pid: 4016060590, nl_payload: Genlmsghdr { cmd: 0, version: 0, reserved: 0, header: NoUserHeader, attrs: GenlBuffer([]) } }, ext_ack: GenlBuffer([]) })
```

zoispag:
@jessegoodier absolutely! Looks clean. My instinct initially was to do the same as your approach, but I was trying to avoid breaking changes.
Thanks! 🙏

jessegoodier:
> @jessegoodier absolutely! Looks clean. My instinct initially was to do the same as your approach, but I was trying to avoid breaking changes. Thanks! 🙏

For sure. We are looking at 3.4 GA tomorrow. Really close on this getting into that, but limited time for testing.

I'll talk to my team and see what we can do.

jessegoodier (approved):

jessegoodier:
This is going into 3.3.1, should be early next week.

PR #4490, opened 2025-12-11
KCM-4903: Add annotations and creation logic to cluster controller SA

HMetcalfeW:
## Release Notes Headline

Adjust the Cluster Controller's Service account to take into account `.Values.serviceAccount.create` and adds the appropriate annotations `.Values.serviceAccount.annotations` so that IRSA/Pod Identity annotations are appropriately added.

## Commentary for Reviewers

Mimic of Cloud Cost's Service Account ([ref](https://github.com/kubecost/kubecost/blob/b8c044792a64c5a83a2a6406e10c55200ccfa53a/kubecost/templates/cloud-cost/cloud-cost-service-account.yaml#L9))

## Links to Issues or Tickets

Addresses a known bug.

## User Impact and Risks

No risk. SA is not created if `.Values.serviceAccount.create` is false respecting the original intent of the flag.

## Testing

Using the workflow
```
cd kubecost
helm dependency update
helm template . --values.yaml > manifest.yaml
```

Adjusted the values.yaml for the following and verified the SA is not created in the `manifest.yaml`
```
clusterController:
   serviceAccount:
      create: false
```

Adjusted the `values.yaml` to add
```
clusterController:
   serviceAccount:
      create: true
      annotations:
         eks.amazonaws.com/role-arn: arn:aws:iam::123456789:role/kubecost-role
```

Rendered `manifest.yaml`
```
# Source: kubecost/templates/cluster-controller/cluster-controller-service-account.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: release-name-cluster-controller
  namespace: default
  labels:
    app.kubernetes.io/name: kubecost
    helm.sh/chart: kubecost-0.0.0-dev
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
    app: kubecost
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::123456789:role/kubecost-role
```

## Documentation Updates

N/A

PR #4842, opened 2026-09-16
fix Cluster Controller Service Accounts to work with a different name

TrevorKWalker:
## Release Notes Headline

Provides a fix to the cluster controller section of the helm chart to allow users to point to a service account of a different name.

## Commentary for Reviewers

Currently the Cluster controller is hardcoded to set the service account name to <deployment>-cluster-controller. This works if `clusterController.serviceAccount.create: true` but if a customer wants to use their own SA then they are currently forced to name it in the same way. The fix allows customers to use `clusterController.serviceAccount.name` field to change the name that cluster controller will bind to

## Links to Issues or Tickets

https://apptio.atlassian.net/browse/KCM-6451

## User Impact and Risks

Shouldn't affect any working configurations as defaults remain the same. The change only adds functionality of being more flexible in adding custom service accounts

## Documentation Updates

Will need to add a section to explaining the ability to change the name [https://www.ibm.com/docs/en/kubecost/self-hosted/3.x?topic=SSW0JQG_2.x/install-and-configure/advanced-configuration/controller/cluster-controller.html?cp=SSW0JQG_3.0.x](https://www.ibm.com/docs/en/kubecost/self-hosted/3.x?topic=SSW0JQG_2.x/install-and-configure/advanced-configuration/controller/cluster-controller.html?cp=SSW0JQG_3.0.x)


HMetcalfeW (changes requested):
## Verdict

Overall looks good. A couple things to fix.

## Findings

### 1. Must fix: OpenShift SCC still grants to the fixed name

`kubecost/templates/cluster-controller/cluster-controller-ocp-scc.yaml` line 32 keeps `system:serviceaccount:<ns>:{{ include "kubecost.clusterController.name" . }}`. It is the one consumer of the SA name the PR missed. Rendered with `values-openshift.yaml` (which sets `global.platforms.openshift.scc.clusterController: true`) plus the customer values, the pod runs as `kubecost-serviceaccount` while the SCC lists `kubecost-cluster-controller`; same mismatch with `create: true, name: cc-custom-sa`. Fix: use `kubecost.clusterController.serviceAccountName` in the `users` entry. The SCC's own `metadata.name` should stay on `kubecost.clusterController.name`.

### 2. Should fix: `create: false` with no name silently rebinds to the global service account

Base behavior for `create: false` with `name` unset is `<release>-cluster-controller`; since the `create` flag landed (#4490, Dec 2025) the only way that path ever worked was to pre-create an account of that name, typically carrying the cloud identity annotation. Head changes the fallback to the global account (`kubecost`, or `serviceAccount.name`): on upgrade those installs move the pod off their pre-created account, drop its annotation, and the CRB now binds the cluster controller ClusterRole (cluster-wide patch/delete on nodes, pods, deployments, eviction, namespaces) to the account shared by aggregator, cloud-cost, and forecasting. That is the exact coupling Trevor argued against in the thread, arriving through the fallback path with no release note. The ClusterRole template's own header says these permissions "will not be added to any clusters by default".

Fix: keep the historical default in both branches. That collapses the helper to one line, matches the aggregator and cloud-cost shape (explicit name wins, otherwise the component default), and never touches the global account:

```
{{- define "kubecost.clusterController.serviceAccountName" -}}
{{- default (include "kubecost.clusterController.name" .) .Values.clusterController.serviceAccount.name -}}
{{- end -}}
```

With this, `create: false` means "I will supply the account named here (or the default name)", which is what the customer expects and what the flag has meant so far. If a global-account fallback is genuinely wanted later, it should be an explicit value, not the empty-name default.

### 3. Nits

- `values.yaml`: the new `name` lacks the `## @param clusterController.serviceAccount.name ...` annotation its siblings carry (`create`, `annotations`). Suggested text: "Name of the service account to use. Defaults to `<release>-cluster-controller`. When `create` is false the account must already exist."
- `create: true` with `name` equal to the global account's name renders two ServiceAccount objects of the same name (component template plus `kubecost-service-account.yaml`); `helm template` and `helm lint` do not flag it. The docs update should steer users to `create: false` when pointing at an existing account.
- `clusterController.serviceAccount.annotations` is ignored when `create` is false (nothing is rendered); the customer's values carry it redundantly. Worth one sentence in the docs.
- PR title: the repo's recent history prefixes the ticket (`KCM-4903: ...` on #4490); `KCM-6451:` would match. The Testing section of the PR body is empty; `AGENTS.md` asks for `helm template` before/after output, and the table above can be pasted in. Might be worth a Jira ticket to have traceback on the customer behalf

HMetcalfeW (approved):

